PR #19, opened 2021-08-31
Add Typography to Watt and Storybook

### LINKED ISSUE #9: Typography
### Is your feature request related to a problem? Please describe.
Add correct typography to storybook project

### Describe the solution you'd like.
The correct typography should be implemented and documented 

### Describe alternatives you've considered.
NONE

dzhavat:
## Description

* Set default font to Calibri (does not apply to Angular Material components)
* Add typography styles to Watt design guide
* Document typography in Storybook

## References

Closes #9


libs/ui-watt/src/lib/foundations/typography/storybook/shared/styles.ts
```diff
@@ -0,0 +1,13 @@
+export const typographyHtmlSnippets = {
```
frontend-specialisten, line 1: Consider renaming this file to snippets, or something else not to confuse with css styles

libs/ui-watt/src/lib/foundations/typography/index.scss
```diff
@@ -0,0 +1,92 @@
+body h1,
```
frontend-specialisten, line 1: These styles should also be available in either css-vars or some scss function so they can be received inside of a stylesheet of a component

frontend-specialisten (changes requested):

frontend-specialisten:
Great trick about the template however!

libs/ui-watt/src/lib/foundations/typography/index.scss
dzhavat: Question: Do you think that would be necessary? Since these are global styles, they will just "apply" everywhere. Unless of course there's another style that overwrites them. If the Angular theme is configured to use the same font, things will "just work". And besides the tag (`h1`) selector , there's also a `.watt-heading-1` class in case someone wants a `p` that looks like a heading (which we must be careful of :D)

frontend-specialisten: That's true, lets skip it until it's needed

frontend-specialisten (approved):

PR #24, opened 2021-09-03
Adjust typography examples in Storybook

### LINKED ISSUE #9: Typography

dzhavat:
## Description

Adjust typography examples in Storybook

## References

#9

### DIFF
--- a/apps/dh3-admin/src/index.html
+++ b/apps/dh3-admin/src/index.html
@@ -6,7 +6,13 @@
     <base href="/" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
-    <link rel="preconnect" href="https://fonts.gstatic.com" />
+
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;600;700&display=swap"
+      rel="stylesheet"
+    />
     <link
       href="https://fonts.googleapis.com/icon?family=Material+Icons"
       rel="stylesheet"
--- a/libs/ui-watt/src/lib/foundations/typography/storybook/typography.component.html
+++ b/libs/ui-watt/src/lib/foundations/typography/storybook/typography.component.html
@@ -5,7 +5,7 @@
 <mat-card>
   <h3>Font</h3>
   <p>
-    <strong>Open Sans</strong> is the typeface used on Energinet platforms. The
+    <strong>Open Sans</strong> is the typeface used on Energinet platforms.
     fallback font is <strong>Calibri.</strong>
   </p>
 </mat-card>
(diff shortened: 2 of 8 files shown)


libs/ui-watt/src/lib/foundations/typography/storybook/typography.component.html
```diff
@@ -5,8 +5,8 @@ <h1>Typography</h1>
 <mat-card>
   <h3>Font</h3>
   <p>
-    <strong>DIN Next LT Pro</strong> is the typeface used on Energinet
-    platforms. The fallback font is <strong>Calibri.</strong>
+    <strong>Open Sans</strong> is the typeface used on Energinet platforms. The
```
frontend-specialisten, line 8:
```suggestion
    <strong>Open Sans</strong> is the typeface used on Energinet platforms.
```

libs/ui-watt/src/lib/foundations/typography/storybook/typography.component.html
```diff
@@ -5,8 +5,8 @@ <h1>Typography</h1>
...
+    fallback font is <strong>Calibri.</strong>
```
frontend-specialisten, line 9:
```suggestion
```

libs/ui-watt/src/lib/watt.scss
```diff
@@ -1,13 +1,15 @@
...
+@import './foundations/box-sizing';
+
 html,
 body {
   height: 100%;
 }
 body {
   margin: 0;
-  font-family: Calibri, sans-serif;
+  font-family: 'Open Sans', Calibri, sans-serif;
```
frontend-specialisten, line 12:
```suggestion
  font-family: 'Open Sans', sans-serif;
```

apps/dh3-admin/src/index.html
```diff
@@ -6,7 +6,13 @@
     <base href="/" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
-    <link rel="preconnect" href="https://fonts.gstatic.com" />
+
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
```
frontend-specialisten, line 12: Please move this to the scss styles instead, so we don't have to do the import in every application:
@import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;600;700&display=swap');

frontend-specialisten (changes requested):
Looks good, only have two comments. I don't think we should implement calibri as fallback anymore, so I have suggested it to be removed. Also just a minor request of how we import the fonts.

apps/dh3-admin/src/index.html
dzhavat: Question: If we import the font this way, wouldn't we lose the connection between the import and the `link rel="preconnect"` tags? It will look like they are there some sort of unnecessary legacy code. 

I don't think we can avoid not having some manual setup steps when generating a new application. For example, the `watt.scss` file still need to be added to the `styles` array in `angular.json` file for each new project in order to use the Design System. The `link rel="preconnect"` tags will need to be added manually as well.

I think we probably need a guide with all the steps necessary to properly setup a newly generated app.

frontend-specialisten: Due to performance we keep it as is, and document these in our usage / getting started

frontend-specialisten (approved):

PR #26, opened 2021-09-09
added CD for design system

Renetnielsen:
## References

* #0000


LayZeeDK:
suggestion: use `affected` status of `ui-watt` to determine whether to run this workflow, similar to https://github.com/Energinet-DataHub/greenforce-frontend/blob/main/.github/workflows/ett.yml#L24-L71

PR #31, opened 2021-09-20
Change prefix for DataHub project

dzhavat:
## Description

* Change prefix for DataHub project 
* Change NX default base branch

## References

-#

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -5,7 +5,7 @@
       "projectType": "application",
       "root": "apps/dh3-admin",
       "sourceRoot": "apps/dh3-admin/src",
-      "prefix": "energinet",
+      "prefix": "dh",
       "architect": {
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
--- a/apps/dh3-admin/.eslintrc.json
+++ b/apps/dh3-admin/.eslintrc.json
@@ -13,7 +13,7 @@
           "error",
           {
             "type": "attribute",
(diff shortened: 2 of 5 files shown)


LayZeeDK (changes requested):
I thought we were using `gf` (Green Force), not `dh` (DataHub) 🤔

We're going to use `ett` (Energy Track and Trace) in Energioprindelse.

dzhavat:
> I though we were using gf (Green Force), not dh (DataHub) 🤔

"Green Force" is the name of the ART. The DataHub app is still called DataHub.

LayZeeDK:
Maybe we shouldn't have a global prefix setting then. Each project will need to have `dh` or `ett` specified.

dzhavat:
Do you mean the `defaultProject` setting?

LayZeeDK:
> Do you mean the defaultProject setting?

Sorry, I thought this was a global `prefix` setting. Nevermind me!

LayZeeDK (approved):

dzhavat:
> Which npmScope will we use?

I'm not sure what is the difference if we change it. Right now it's `energinet`. The Watt Design System is also imported by `@energinet/watt` which I think is fine. Even if it gets moved to a different repo, the same import still makes sense.

frontend-specialisten (approved):
Agreed, I think energinet makes sense of npmScope and the different applications should have their own prefix

PR #33, opened 2021-09-21
ci: add workspace workflow

LayZeeDK:
### Pull request checks
- Production build of affected products
- End-to-end test affected products
- Format and check formatting of affected projects
- Lint affected projects with fixers
- Unit and integration test affected projects

### Merge jobs
- Production build of all products
- End-to-end test of all products
- Lint all projects
- Unit and integration test all projects

## References

- https://github.com/Energinet-DataHub/energy-track-and-trace/issues/45


dzhavat:
@LayZeeDK Just pushed some changes related to unit tests and building of the Watt library. 

The Watt e2e tests however still fail because there are none in the `ui-watt-e2e` folder. Maybe I should just add one to unblock you but this needs to be looked at.

dzhavat:
@LayZeeDK I'm going to delete the `ui-watt-e2e` project. We'll use Chromatic for visually testing the Design System instead.

libs/ui-watt/src/index.ts
```diff
@@ -1 +1,3 @@
-export * from './lib/ui-watt.module';
+export { WattModule } from './lib/ui-watt.module';
+export { AutocompleteComponent } from './lib/components/autocomplete/autocomplete.component';
+export { AutocompleteModule } from './lib/components/autocomplete/autocomplete.module';
```
dzhavat, line 3: Question: Is it correctly understood that these artifacts must be exported here even though they are also part of `WattModule`? The build was failing if they weren't. That's how I understood it from here at least https://angular.io/guide/creating-libraries

nx.json
```diff
@@ -15,7 +15,9 @@
       "runner": "@nrwl/nx-cloud",
       "options": {
         "cacheableOperations": ["build", "lint", "test", "e2e"],
-        "accessToken": "M2QwZmQ2YTQtMjFjZi00ZmI4LTk0MDItZjEyOTQ4ZmMwMTUyfHJlYWQtd3JpdGU="
+        "accessToken": "M2QwZmQ2YTQtMjFjZi00ZmI4LTk0MDItZjEyOTQ4ZmMwMTUyfHJlYWQtd3JpdGU=",
+        "canTrackAnalytics": true,
+        "maxParallel": 6
```
dzhavat, line 20: Question: Should this number and `NX_MAX_PARALLEL` in the build script match?

package.json
```diff
@@ -9,32 +9,35 @@
     "yarn": "1.22.11"
   },
   "scripts": {
-    "ng": "nx",
-    "postinstall": "node ./decorate-angular-cli.js && ngcc --properties es2015 browser module main",
```
dzhavat, line 13: Question: What is the reason for not running `ngcc` after install locally but including it in the pipeline?

.github/workflows/workspace.yml
```diff
@@ -0,0 +1,275 @@
...
+
+env:
+  BASE_BRANCH: ${{ github.event.pull_request.base.ref }}
+  is-main-branch: ${{ github.ref == 'refs/heads/main' }}
+  is-pull-request: ${{ github.event_name  == 'pull_request' }}
+  NODE_OPTIONS: --max_old_space_size=6144
+  node-version: 14.x
+  NX_BRANCH: ${{ github.event.number }}
+  NX_MAX_PARALLEL: 2
+  NX_RUN_GROUP: ${{ github.run_id }}
```
dzhavat, line 11: Nitpick: `NX_RUN_GROUP` seems unused

.github/workflows/workspace.yml
```diff
@@ -0,0 +1,275 @@
+name: Green Force Frontend
+
+env:
+  BASE_BRANCH: ${{ github.event.pull_request.base.ref }}
+  is-main-branch: ${{ github.ref == 'refs/heads/main' }}
+  is-pull-request: ${{ github.event_name  == 'pull_request' }}
+  NODE_OPTIONS: --max_old_space_size=6144
+  node-version: 14.x
+  NX_BRANCH: ${{ github.event.number }}
```
dzhavat, line 9: Nitpick: `NX_BRANCH` seems unused

libs/ui-watt/src/index.ts
frontend-specialisten: Yes we need to provide these, as I remember it was since version 9 this is required otherwise the compiler will throw an error. So everything we export in our module, we need export in our index

nx.json
frontend-specialisten: I guess NX_MAX_PARALLEL is for the CI, and maxParallel in nx.json is for our local environment? @LayZeeDK

frontend-specialisten (commented):
Besides @dzhavat comments, it looks good to me

nx.json
LayZeeDK: Exactly. To the best of my knowledge, 2 parallel jobs work best in GitHub workflows and 6 is the max beneficial with a powerful local machine.

package.json
LayZeeDK: It's not required for certain tooling so I don't want to waste precious time or resources in jobs that don't require it. Additionally, it can be optimized further for single applications. Especially Angular CDK and Material are terrible with NGCC because of their dozens of entry point. Angular CLI will run NGCC on demand if we don't do it in postinstall. However, the ngcc script will be faster than default NGCC settings.

.github/workflows/workspace.yml
LayZeeDK: It's used by Nx Cloud.

.github/workflows/workspace.yml
LayZeeDK: It's used d by Nx Cloud.

LayZeeDK:
Let me know if you're satisfied with the answers, @dzhavat

dzhavat:
@LayZeeDK I've resolved some of the comments. As for the rest, please resolve them as you decide whether you'll take action on them or not. I'm fine either way.

Note: Please rebase/merge `main` due to file conflicts.

.github/workflows/workspace.yml
LayZeeDK: Added a variable group for Nx Cloud.

.github/workflows/workspace.yml
LayZeeDK: Added a variable group for Nx Cloud.

dzhavat (approved):

PR #48, opened 2021-09-27
feat(color): add colors to designsystem

frontend-specialisten:


libs/ui-watt/src/lib/foundations/color/colors.ts
```diff
@@ -0,0 +1,35 @@
+/*
+ * @Please notice!!! 
+ * These values should be kept in-sync with _colors.import.scss
+ * Naming convention is as follow: --watt-{COLOR GROUP}-colors-{COLOR NAME}
+ */
+export enum Colors {
+    'primary' = '--watt-primary-colors-primary',
+    'primary-light' = '--watt-primary-colors-primary-light',
```
dzhavat, line 8: Suggestion: Use Pascal case for enum keys, e.g. `PrimaryLight`. This will also make it unnecessary to quote the keys.

Suggestion: What do you think about the naming convention that looks like `--watt-color-primary`, `--watt-color-primary-dark`, `--watt-color-neutral-grey-100`, etc. instead? The `color` part after `--watt` can be used for grouping as well.

libs/ui-watt/src/lib/foundations/color/storybook/color.component.html
```diff
@@ -0,0 +1,15 @@
...
+    <h1>Color</h1>
+</mat-card>
+
+<mat-card *ngFor="let type of colorTypes">
+    <h2>{{type.title}}</h2>
+    <p>{{type.description}}</p>
+
+    <div class="color-grid">
+        <figure *ngFor="let color of type.colors">
+            <div class="color-sample" [style.backgroundColor]="color.color" [style.color]="color.contrast">Text on {{color.name}}</div>
```
dzhavat, line 11: Question: The "Text on ..." part is not on Figma. Can you align with @stephaniegitha on that? It'd be nice if we can keep these things consistent.

libs/ui-watt/src/lib/foundations/color/storybook/color.component.html
```diff
@@ -0,0 +1,15 @@
...
+            <figcaption>{{color.name}}</figcaption>
```
dzhavat, line 12: Suggestion: Use `watt-label` class to style the text similar to what is used in Figma.

libs/ui-watt/src/lib/foundations/color/storybook/color.component.scss
```diff
@@ -0,0 +1,37 @@
...
+    text-transform: uppercase;
+}
+
+.color-sample {
+    height: 150px;
+    width: 150px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    border: 1px solid #{colors.getColor('grey-300')};
```
dzhavat, line 36: Suggestion: There's border only on the "white" color swatch in Figma. Consider aligning for consistency.

libs/ui-watt/src/lib/foundations/color/storybook/color.component.scss
```diff
@@ -0,0 +1,37 @@
...
+    max-width: 688px;
+}
+
+figure {
+    margin: 10px;
+}
+
+figcaption {
+    text-transform: uppercase;
+}
```
dzhavat, line 28: Suggestion: Use the `.watt-label` on the element. It will uppercase the text automatically.

libs/ui-watt/src/lib/foundations/color/storybook/color.component.ts
```diff
@@ -0,0 +1,71 @@
+import { Component } from '@angular/core';
+import { ColorHelperService } from '../color-helper.service';
```
dzhavat, line 2: Suggestion: Remove unused import.

libs/ui-watt/src/lib/foundations/color/storybook/color.stories.ts
```diff
@@ -0,0 +1,45 @@
...
+export default {
+  title: 'Foundations/Color',
+  component: ColorComponent,
+  decorators: [
+    moduleMetadata({
+      imports: [MatTableModule, MatCardModule],
+      providers: [
+        { provide: Window, useValue: window }
+      ]
+    })
```
dzhavat, line 16: Suggestion: Consider creating a `ColorModule` module similar to `TypographyModule` and move these things there. It makes it easier to see how the Material modules are related to the component where they are used in and it's easier to reuse (if necessary).

libs/ui-watt/src/lib/foundations/color/storybook/color.component.ts
```diff
@@ -0,0 +1,71 @@
...
+      title: 'State Colors',
+      description:
+        'State color helps users find people, identify status, see actions, locate help, and understand next steps. The consistent use of color keeps cognitive load low and makes for a unified and engaging user experience.',
+      colors: [],
+    },
+  ];
+
+  constructor() {
+    this.groupColors();
+    console.log(this.colorTypes);
```
dzhavat, line 46: Nitpick: Remove extra console log.

dzhavat (changes requested):

libs/ui-watt/src/lib/foundations/_variables.scss
```diff
@@ -0,0 +1,35 @@
...
+    // States
+    @include colors.setColor('danger', #CC493E, colors.getColor('white'));
+    @include colors.setColor('danger-light', #F7E2E0, colors.getColor('black'));
+    @include colors.setColor('warning', #F8AE3C, colors.getColor('black'));
+    @include colors.setColor('warning-light', #FEF2E0, colors.getColor('black'));
+    @include colors.setColor('success', #36b37e, colors.getColor('black'));
+    @include colors.setColor('success-light', #DFF3EA, colors.getColor('black'));
+    @include colors.setColor('info', #00A6BC, colors.getColor('black'));
+    @include colors.setColor('info-light', #D6F1F4, colors.getColor('black'));
+  }
```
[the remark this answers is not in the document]
frontend-specialisten: The reason is to avoid "magic strings" of variables floating across the applications. With the setColor function we have the control as the consumers don't set the variables directly.

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
```diff
@@ -0,0 +1,103 @@
...
+ * @description
+ *
+ * Get color by name.
+ *
+ * @publicApi
+ */
+@function getColor($name) {
+    $color-var: map-get($_colors, $name);
+    @return var(#{$color-var});
+}
```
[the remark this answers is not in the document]
frontend-specialisten: The reason is to avoid "magic strings" of variables floating across the applications. With the getColor function we have the control as the consumers don't get the variables directly.

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
```diff
@@ -0,0 +1,103 @@
...
+ * @description
+ *
+ * Get text-color / contrast color by name.
+ *
+ * @publicApi
+ */
+@function getColorContrast($name) {
+    $color-var: map-get($_colors, 'contrast', $name);
+    @return var(#{$color-var});
+}
```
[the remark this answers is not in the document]
frontend-specialisten: If you as a consumer needs to style something with the colors, you maybe also need to set the contrast / text color.

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
```diff
@@ -0,0 +1,103 @@
...
+ * @description
+ *
+ * Set color and contrast by name.
+ *
+ * @publicApi
+ */
+@mixin setColor($name, $color, $contrast) {
+    #{map-get($_colors, $name)}: $color;
+    #{map-get($_colors, 'contrast', $name)}: $contrast;
+}
```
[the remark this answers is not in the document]
frontend-specialisten: It's needed to avoid magic strings, the css variables has to be set at some point.

libs/ui-watt/src/lib/foundations/color/color-helper.service.ts
```diff
@@ -0,0 +1,62 @@
...
+  private renderer: Renderer2;
+  private colorContrastSuffix = 'contrast';
+
+  constructor(
+    private rendererFactory: RendererFactory2,
+    @Inject(DOCUMENT) private document: Document,
+    private window: Window
+  ) {
+    this.renderer = this.rendererFactory.createRenderer(null, null);
+  }
```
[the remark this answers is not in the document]
frontend-specialisten: That's a fair question, I've just made the implementation to be identical with the capabilities in scss

libs/ui-watt/src/lib/foundations/_variables.scss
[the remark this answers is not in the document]
dzhavat: With Custom Properties we still keep the control and defined the variables in the Design System. The applications just consume them directly instead of using Sass functions.

libs/ui-watt/src/lib/foundations/color/storybook/color.component.scss
frontend-specialisten: This is done to avoid having to look for if the color is white as the colors are printed dynamically, which adds unnecessary logic and really is fragile as this properly only applies when the background is white so the real logic should be, that it should have a border if the color and background is the same or even the contrast aren't good enough. Alternative manually add the color-samples manually but really not better either, at least in my opinion. 
I don't think @stephaniegitha minds, it's barely visible and takes some time to fix which really doesn't give us any value besides matching figma perfectly.

libs/ui-watt/src/lib/foundations/color/storybook/color.component.scss
frontend-specialisten: Totally agree, will fix asap

libs/ui-watt/src/lib/foundations/color/color-helper.service.ts
```diff
@@ -0,0 +1,62 @@
...
+  public getColor(colorName: Colors): string {
+    return this.getComputedStyle().getPropertyValue(colorName);
+  }
+
+  public getColorContrast(colorName: Colors): string {
+    return this.getComputedStyle().getPropertyValue(
+      `${colorName}-${this.colorContrastSuffix}`
+    );
+  }
+
```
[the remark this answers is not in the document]
frontend-specialisten: I actually don't think it's an uncommon usecase remember the constrast color equals the text color

libs/ui-watt/src/lib/foundations/color/colors.ts
frontend-specialisten: Agreed, I will change that asap

libs/ui-watt/src/lib/foundations/color/storybook/color.component.html
frontend-specialisten: Yes we need to align this as Figma doesn't include any text / contrast colors

libs/ui-watt/src/lib/foundations/color/storybook/color.component.scss
dzhavat: It's not only the value that we get but also the time that we use to create something. What is the point of creating something in one place and do it differently somewhere else? If Figma is the source of truth for the Design System, then we should align with it. And sometimes we might do something in Storybook which initiates changes in Figma. So it goes both ways. If we can align, let's do it.

frontend-specialisten: Totally agree, we should align

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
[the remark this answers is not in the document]
dzhavat: They are still available/visible in the apps/DevTools. So nothing prevents a feature developer to access these values directly using CSS Custom Properties API (i.e. `var()`).

image

libs/ui-watt/src/lib/foundations/color/color-helper.service.ts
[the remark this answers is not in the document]
dzhavat: I can see when this is needed when building a theme, but that is internal. I'm not sure this is needed outside the Design System.

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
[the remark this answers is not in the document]
frontend-specialisten: That's true, if the consumer wants to break the system they're able to do it because that we're using custom vars, however then they haven't followed the documentation and been creative by themselves. We can't really do anything about that. However this enables us not having "public" variables floating around, so we shouldn't change it more than 2 places if we decide to rename, like you suggest in another comment. Also the setColor function gives us the control of ensuring a color always has a contrast / text color configured as well. Also if I could choose as a developer it's easier for me to remember getColor('primary') than --watt-color-some group-something-something. But I see your point...
 
I could add some extra documentation which states that you shouldn't set the custom vars directly, how about that?
If you're totally against it, I will ofc. change it :)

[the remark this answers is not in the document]
dzhavat: I don't currently see a case where a developer might need contrast colors and the option to set a color. 

That's because the contrast values will be used only when building the Material theme. If we still need to change the style of a certain Material component, we can do that in the specific app using specific selectors.

On the other hand, the `setColor` function will change the value of a CSS Custom Property for a certain selector but then the `getValue` function still has to be used in order to apply the color. e.g.:

image

Another point is that the developers don't know what color keywords are available for them to use. That can be fixed with documenting them.

libs/ui-watt/src/lib/foundations/_variables.scss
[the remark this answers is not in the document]
frontend-specialisten: setColor has now been removed

libs/ui-watt/src/lib/foundations/_variables.scss
```diff
@@ -0,0 +1,60 @@
+@use './color/colors.import' as colors;
```
dzhavat, line 1: Suggestion: Remove this import since it's no longer necessary.

Suggestion: Move this file under `color` directory and rename it to `index.scss` (like in Typography). So the import in `watt.scss` will become:

```
@use './../../foundations/color';
```

Question: Do you think it's better if we rename the "color" folder to "colors" (plural). It sounds more correct when looking from outside since there are multiple colors defined in it.

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
```diff
@@ -0,0 +1,89 @@
+@use 'sass:map';
+
+/*
+ * @Please notice!!! 
+ * These values should be kept in-sync with colors.ts
+ * Naming convention is as follow: --watt-color-{COLOR GROUP (ONLY IF NOT THE SAME AS COLOR NAME)}-{COLOR NAME}
```
dzhavat, line 6: Suggestion: Consider removing this line because it sounds confusing having a conditional in it.

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
```diff
@@ -0,0 +1,89 @@
...
+        // States
+        'danger': var(--watt-color-state-danger-contrast),
+        'danger-light': var(--watt-color-state-danger-light-contrast),
+        'warning': var(--watt-color-state-warning-contrast),
+        'warning-light': var(--watt-color-state-warning-light-contrast),
+        'success': var(--watt-color-state-success-contrast),
+        'success-light': var(--watt-color-state-success-light-contrast),
+        'info': var(--watt-color-state-info-contrast),
+        'info-light': var(--watt-color-state-info-light-contras)t
+    )
```
dzhavat, line 66: Suggestion: Move these values to a separate map (e.g. `$_contrastColors` or similar). It will make it easier to distinguish what is what.

libs/ui-watt/src/lib/foundations/_variables.scss
```diff
@@ -0,0 +1,60 @@
...
+
+:root {
+  /**
+   * Colors
```
dzhavat, line 5: Nitpick: Change this to "Primary"

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
```diff
@@ -0,0 +1,89 @@
+@use 'sass:map';
```
dzhavat, line 1: Question: Do you think renaming this file to `_colors.util.scss` or `_colors.api.scss` will make it more clear as to what it does? Or maybe split it into two files - one with API functions and one with color maps.

libs/ui-watt/src/lib/foundations/color/colors.ts
```diff
@@ -0,0 +1,35 @@
+/*
+ * @Please notice!!! 
+ * These values should be kept in-sync with _colors.import.scss
+ * Naming convention is as follow: --watt-color-{COLOR GROUP (ONLY IF NOT THE SAME AS COLOR NAME)}-{COLOR NAME}
```
dzhavat, line 4: Suggestion: Consider removing this line because it sounds confusing having a conditional in it.

libs/ui-watt/src/lib/foundations/color/colors.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export enum Colors {
+    'primary' = '--watt-color-primary',
```
dzhavat, line 7: Nitpick: The enum keys can be unquoted since they don't contain dashes anymore.

libs/ui-watt/src/lib/foundations/color/storybook/color.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+//👇 Each story then reuses that template
+export const Color = Template.bind({});
+Color.parameters = {
+  docs: {
+    source: {
+      code: ` // Usage from SCSS:
+@use '@energinet/watt' as watt;
+.my-element {
+  background: watt.getColor('COLOR NAME');
+  color: watt.getColorContrast('COLOR NAME-contrast');
```
dzhavat: Question: Shouldn't the "COLOR NAME-contrast" part be just "COLOR NAME"?

Nitpick: Consider changing "COLOR NAME" string to "<color-name>" to make it more explicit.

libs/ui-watt/src/lib/foundations/color/storybook/color.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Usage from TypeScript:
+1. import { Colors, ColorHelperService } from '@energinet/watt';      
+1. Inject the ColorHelperService
+2. Use ColorHelperService.getColor(Colors.COLOR NAME); or ColorHelperService.getColorContrast(Colors.COLOR NAME);
```
dzhavat: Question: Do we need a list of the available color names?

libs/ui-watt/src/lib/foundations/color/storybook/color.stories.ts
```diff
@@ -0,0 +1,41 @@
...
+      code: ` // Usage from SCSS:
+@use '@energinet/watt' as watt;
+.my-element {
+  background: watt.getColor('COLOR NAME');
+  color: watt.getColorContrast('COLOR NAME-contrast');
+}
+
+// Usage from TypeScript:
+1. import { Colors, ColorHelperService } from '@energinet/watt';      
+1. Inject the ColorHelperService
```
dzhavat, line 33: Nitpick: Two times 1.

dzhavat (changes requested):

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
frontend-specialisten: The structure of the map is done to be as close as how Angular Material is doing the color palettes. If it doesn't give us any benefits when the theming is applied then I agree we should change it. I will leave it as is for now.

libs/ui-watt/src/lib/foundations/color/_colors.import.scss
frontend-specialisten: I've taken the naming from angular material, the file does only contain the color map now

libs/ui-watt/src/lib/foundations/color/storybook/color.stories.ts
frontend-specialisten: I've done a copy to clipboard functionality now, when you hover the color samples. I will try take a look at how we can document the full-list, maybe we can make use of combodoc for that purpose

frontend-specialisten: I will give storybook-addon-cssprops a try

dzhavat: Just saw that the variable names are on top of the colors. It looks fine as a first version. The only thing is aligning it with Stephanie. Maybe she'll need to update Figma accordingly.

frontend-specialisten: storybook-addon-cssprops printed out errors in the console due to wrong implementation in React, so that's not an option. If we wanna document it further down the road, we're properly forced to do something ourselves.

libs/ui-watt/src/lib/foundations/color/storybook/color.component.scss
```diff
@@ -0,0 +1,47 @@
+@use 'sass:map';
```
dzhavat, line 1: Nitpick: Unused import.

libs/ui-watt/src/lib/foundations/color/storybook/color.stories.ts
```diff
@@ -0,0 +1,38 @@
...
+    source: {
+      code: ` // Usage from SCSS / CSS (tip: hover over the color sample, and click to copy to clipboard):
+.my-element {
+  background: var(COLOR NAME);
+}
+
+// Usage from TypeScript:
+1. import { Colors, ColorHelperService } from '@energinet/watt';      
+2. Inject the ColorHelperService
+3. Use ColorHelperService.getColor(Colors.COLOR NAME);
```
dzhavat, line 34: I can see now that GitHub "ate" part of my previous comment related to this line. The last line should've been: 

Suggestion: Consider changing "COLOR NAME" string to `<color-name>` to make it more explicit.

E.g.

```
 3. Use ColorHelperService.getColor(Colors.<color-name>);
```

frontend-specialisten: Roger that

dzhavat (approved):

PR #49, opened 2021-09-29
Ignore false positive secret flagged by GitGuardian

dzhavat:
## References

- #0000

### DIFF
--- a/.gitguardian.yml
+++ b/.gitguardian.yml
@@ -1,0 +1,3 @@
+matches-ignore:
+  - name: nxCloudAccessToken
+    match: 994b8e7c00028ee5981c35896a36f38356a7997f # read-only token
--- a/.gitguardian.yml
+++ b/.gitguardian.yml
@@ -1,3 +1,3 @@
 matches-ignore:
   - name: nxCloudAccessToken
-    match: 994b8e7c00028ee5981c35896a36f38356a7997f # read-only token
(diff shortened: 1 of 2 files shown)


.gitguardian.yaml
```diff
@@ -0,0 +1,3 @@
+matches-ignore:
```
LayZeeDK, line 1: suggestion: try renaming to `gitguardian.yml` as per [the documentation](https://docs.gitguardian.com/internal-repositories-monitoring/gg_shield/configuration#configuration-file).

dzhavat: Spot on! I'm pretty sure I copied the file name from the documentation somewhere. They have a typo as well :) 

But that doesn't seem to have an effect as well. In fact, the GitGuardian link in the checks section points to the previous commit.

LayZeeDK:
I tried squashing the commits.

.gitguardian.yml
```diff
@@ -0,0 +1,3 @@
+matches-ignore:
+  - name: nxCloudAccessToken
+    match: 994b8e7c00028ee5981c35896a36f38356a7997f # read-only token
```
dzhavat, line 3: Question: Where is this value coming from?

LayZeeDK: It's a commit hash. Trying to figure out what the correct value is.

PR #51, opened 2021-09-29
chore: allow Nx Cloud read-only token

LayZeeDK:
## Description
Attempt to allow the Nx Cloud read-only access token in GitGuardian.

## References

-

### DIFF
--- a/.github/workflows/gitguardian-ignore.yml
+++ b/.github/workflows/gitguardian-ignore.yml
@@ -15,6 +15,9 @@ jobs:
       - name: Install package dependencies
         run: pip install ggshield
 
+      - name: Scan for exposed secrets
+        run: ggshield scan ci
+
       - name: Ignore last found secret
         run: ggshield ignore --last-found
 
--- a/.github/workflows/gitguardian-ignore.yml
+++ b/.github/workflows/gitguardian-ignore.yml
@@ -5,16 +5,6 @@ on:
   workflow_dispatch: {}
 
 jobs:
-  # gitguardian-ignore:
-  #   name: GitGuardian ignore
-  #   runs-on: ubuntu-latest
-  #   container:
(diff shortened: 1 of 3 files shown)


LayZeeDK:
Need GitGuardian API key.

.github/workflows/gitguardian-ignore.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Install package dependencies
+        run: pip install ggshield
+
+      - name: Scan for exposed secrets
+        run: ggshield scan ci
+
+      - name: Ignore last found secret
+        run: ggshield ignore --last-found
```
dzhavat, line 22: Question: Wouldn't this make it possible to expose real secrets without them being caught on time by a developer?

Say a developer opens a PR with a secret that must not be exposed. The pipeline will find the secret, add it to the `.gitguardian.yaml` file and commit the file in the repo. Furthermore, the "GitGuardian Security Checks" check will probably raise a warning but that is not so obvious. So we'll have to rely on catching the secret manually during PR reviews, which not might start until after several hours. 

If you think this is unlikely scenario, please ignore the previous section.

Question: If the previous scenario is likely to happen, should the process of ignoring secrets be manual instead? So someone is required to take an intentional action in order to let a secret stay in the repo.

LayZeeDK: This is a manually triggered workflow. I'm trying to figure out the right configuration entry format but we can't do anything without an API key.

frontend-specialisten:
@LayZeeDK Is this still relevant?

LayZeeDK:
@frontend-specialisten
I don't see GitGuardian status checks failing anymore so we should be able to close it. I'll keep the branch for now.

PR #67, opened 2021-10-14
ci(ett): fix affected job

LayZeeDK:
### CI
- Fix error in affected job for ETT

## References

*N/A*


LayZeeDK:
Error is only in #66.

PR #90, opened 2021-10-28
Establish initial BFF CI

dstsyst:
## Description

- Updated development.md to mention CI build
- Established an initial BFF CI using a shared workflow.

## References

https://app.zenhub.com/workspaces/mighty-ducks-60ae1565be2dfe0017ff66b2/issues/energinet-datahub/mightyducks/158

### DIFF
--- a/libs/api-dh-app/source/DataHub.WebApi/WeatherForecast.cs
+++ b/libs/api-dh-app/source/DataHub.WebApi/WeatherForecast.cs
@@ -1,4 +1,4 @@
-// Copyright 2021 Energinet DataHub A/S
+﻿// Copyright 2021 Energinet DataHub A/S
 //
 // Licensed under the Apache License, Version 2.0 (the "License2");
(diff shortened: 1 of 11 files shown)


libs/api-dh-app/source/DataHub.WebApi/WeatherForecast.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 Energinet DataHub A/S
```
frontend-specialisten, line 1: Not sure why all these lines are changed, but the CI changes looks good to me :)

frontend-specialisten (approved):

libs/api-dh-app/source/DataHub.WebApi/WeatherForecast.cs
dstsyst: They where changed to "UTF-8 with BOM" :)

PR #98, opened 2021-11-08
feat(watt): support loading state in button component

dzhavat:
### Design System

- Support loading state in button component
- Adjust size of spinner component
- Adjust styling for text button 
- Reduce the number of stories and use controls to set individual properties

## References

 - https://github.com/Energinet-DataHub/MightyDucks/issues/215

## PLEASE NOTE

### DIFF
--- a/libs/ui-watt/src/lib/components/button/watt-button.component.html
+++ b/libs/ui-watt/src/lib/components/button/watt-button.component.html
@@ -23,26 +23,11 @@ limitations under the License.
 ></ng-container>
 
 <div #contentWrapper class="content-grid">
-  <ng-container *ngTemplateOutlet="contentWrapperTemplate"></ng-container>
-</div>
-
-<ng-template #contentWrapperTemplate>
-  <ng-container *ngIf="loading; else buttonContentTemplate">
-    <ng-container *ngTemplateOutlet="loadingTemplate"></ng-container>
-  </ng-container>
-</ng-template>
-
-<ng-template #loadingTemplate>
   <watt-spinner
+    *ngIf="loading"
     class="content-grid-item content-grid-item-spinner"
   ></watt-spinner>
 
-  <div class="content-grid content-grid-item">
-    <ng-container *ngTemplateOutlet="buttonContentTemplate"></ng-container>
-  </div>
-</ng-template>
-
-<ng-template #buttonContentTemplate>
   <watt-icon
     *ngIf="hasIcon"
     class="content-grid-item"
@@ -52,4 +37,4 @@ limitations under the License.
   <div class="content-grid-item">
     <ng-content></ng-content>
   </div>
-</ng-template>
+</div>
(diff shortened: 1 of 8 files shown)


libs/ui-watt/src/lib/components/button/watt-button.component.html
```diff
@@ -23,6 +23,26 @@
 ></ng-container>
 
 <div #contentWrapper class="content-grid">
+  <ng-container *ngTemplateOutlet="contentWrapperTemplate"></ng-container>
+</div>
+
+<ng-template #contentWrapperTemplate>
+  <ng-container *ngIf="loading; else buttonContentTemplate">
```
frontend-specialisten, line 30: When both templates is using "buttonContentTemplate" considering just having a ngIf on the spinner, it would be easier to follow.

libs/ui-watt/src/lib/components/button/watt-button.component.scss
```diff
@@ -33,6 +33,33 @@ watt-button {
...
+  &.watt-button-loading {
+    .content-grid {
+      flex-direction: column;
+
+      > .content-grid {
+        flex-direction: row;
+      }
+
+      .content-grid-item:not(.content-grid-item-spinner) {
+        height: 0;
```
frontend-specialisten, line 45: Is this necessary?

frontend-specialisten (changes requested):
Just some nitpicks, also I would like to have a story with an example of the loading functionality, so you can "play" with the loading functionality and see it in action.

libs/ui-watt/src/lib/components/button/watt-button.component.html
dzhavat: The two templates are not completely identical. The `loadingTemplate` wraps `buttonContentTemplate` in an extra `div.content-grid.content-grid-item` element. That is why `buttonContentTemplate` is moved on its own.

libs/ui-watt/src/lib/components/button/watt-button.component.scss
dzhavat: It is used for hiding the icon. Without it I get the following issue:

loading

dzhavat: My mistake. Not setting the `height` results in the content being hidden visually but still taking space on the page.

loading

frontend-specialisten: Wouldn't the "right" thing to do making the spinner absolute then? What if the button has multi-line text (not sure if that's posible), but then the height of the button will change?

dzhavat: > Wouldn't the "right" thing to do making the spinner absolute then

Good idea. Hadn't thought of it. Will try it out.

> What if the button has multi-line text (not sure if that's possible), 

I'm not sure what will happen on multi-line text. Haven't tested it.

dzhavat: Just tried with making the spinner absolute and it works better. The logic is also simpler. Thanks for the suggestion 👍

frontend-specialisten (approved):

PR #100, opened 2021-11-09
feat(watt): add icon button to Design System

dzhavat:
### Design System

- Add separate `<watt-icon-button>` component
- Expose the component from `WattButtonModule`
- Adjust Storybook to use the new component
- Fix property values in Storybook overview page

## References

- https://github.com/Energinet-DataHub/MightyDucks/issues/215

## PLEASE NOTE

### DIFF
--- a/libs/ui-watt/src/lib/components/button/watt-button.stories.ts
+++ b/libs/ui-watt/src/lib/components/button/watt-button.stories.ts
@@ -60,23 +60,28 @@ Overview.parameters = {
 };
 
 //👇 We create a “template” of how args map to rendering
-const ButtonTemplate: Story<WattButtonComponent> = (args) => ({
+const ButtonStory: Story<WattButtonComponent> = (args) => ({
   props: args,
   template: `<watt-button>Button</watt-button>`,
 });
 
-export const Button = ButtonTemplate.bind({});
+export const Button = ButtonStory.bind({});
 Button.args = {
   type: 'text',
 };
 
-const IconButtonTemplate: Story<WattIconButtonComponent> = (args) => ({
+const iconButtonTemplate = (args: Partial<WattIconButtonComponent>) =>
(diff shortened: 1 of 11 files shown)


dzhavat:
clarification: The reason for creating a separate component is because I didn't see how the existing `watt-button` could support an icon only and not end up with a `min-width` of 100px. One solution was to add an extra `@Input` property, e.g. `iconOnly` (boolean), to specifically that the button is just an icon. I didn't go that way because the `watt-button` already has many `@Input` properties and adding more would've made it more complicated to keep track of them at the same time.

libs/ui-watt/src/lib/components/button/watt-button.stories.ts
```diff
@@ -63,3 +69,24 @@ export const Button = ButtonTemplate.bind({});
 Button.args = {
   type: 'text',
 };
+
+const IconButtonTemplate: Story<WattIconButtonComponent> = (args) => ({
+  props: args,
+  template: `<watt-icon-button icon="${args.icon}" [disabled]="${args.disabled}"></watt-icon-button>`,
```
frontend-specialisten, line 75: This template doesn't seem to be reflected in the docs:
image

frontend-specialisten (changes requested):
Couldn't the icon button just be a type like text? Otherwise it looks good to me, only a single comment about the documentation example :)

dzhavat:
> Couldn't the icon button just be a type like text? Otherwise it looks good to me, only a single comment about the documentation example :)

Do you mean adding a fourth type `icon` to `WattButtonType`?

libs/ui-watt/src/lib/components/button/watt-button.stories.ts
dzhavat: Fixed.

frontend-specialisten (approved):

PR #107, opened 2021-11-11
feat(dh): add logo to sidebar and toolbar

frontend-specialisten:
## Description
add logo to sidebar and toolbar

## References

- #0000

## PLEASE NOTE


apps/dh/app-dh/project.json
```diff
@@ -30,6 +30,11 @@
             "input": "libs/dh/globalization/assets-localization/src/assets",
             "output": "assets"
           },
+          {
+            "glob": "**/*",
+            "input": "libs/dh/core/assets-core",
+            "output": "assets"
+          },
```
dzhavat, line 37: suggestion: Let's use `dh-shared-assets` library for the logos. As static files that don't change often, I don't see the need to scatter them between in multiple libraries.

libs/dh/core/shell/src/lib/shell/shell.component.html
```diff
@@ -15,9 +15,18 @@
 limitations under the License.
 -->
 <watt-shell>
-  <ng-container watt-shell-sidenav> Sidenav </ng-container>
+  <ng-container watt-shell-sidenav>
+    <div class="sidenav">
```
dzhavat, line 19: issue(non-blocking): The exact same class is used in `watt-shell` component as well. This is not a problem right now because that component is encapsulated but it might become an issue if we remove the encapsulation. Maybe we should use a `watt-` prefix for classes used in Watt just like Material uses `mat-`. What do you think?

libs/dh/core/shell/src/lib/shell/shell.component.html
```diff
@@ -15,9 +15,18 @@
...
+      <img
+        src="assets/logo-light.svg"
+        alt="DataHub logo"
+        class="logo--sidenav"
```
dzhavat, line 23: suggestion: I think we should agree on the methodology used for naming classes to be consisten. Does this follow BEM?

libs/dh/core/shell/src/lib/shell/shell.component.scss
```diff
@@ -14,10 +14,28 @@
...
 }
+
+.sidenav {
+  display: flex;
+  justify-content: center;
+}
+
+.logo {
+  &--toolbar {
+    @include watt.media('>=Large') {
```
dzhavat, line 34: praise: I didn't know using `>=` was also an option. Nice.

suggestion: Can you add it to the Breakpoint docs?

libs/dh/core/shell/src/lib/shell/shell.component.html
frontend-specialisten: Good point, I think that we should prefix watt

libs/dh/core/shell/src/lib/shell/shell.component.html
frontend-specialisten: Yes this is BEM `--sidenav` is ment as a modifier to logo

dzhavat (approved):

PR #111, opened 2021-11-12
feat(dh): add function to validate metering point id

dzhavat:
### DataHub

* Create `dh/metering-point/domain` library
* Add function to validate metering point id

## References

- https://github.com/Energinet-DataHub/MightyDucks/issues/186

## PLEASE NOTE


libs/dh/metering-point/domain/src/lib/is-metering-point-id.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { meteringPointIdPattern } from './metering-point-id-pattern';
+
+/**
+ * Validates that a metering point id is exactly 18 digits
```
frontend-specialisten, line 20: Consider removing the amount of digits from the comment, otherwise it has to be maintained if the regex changes

libs/dh/metering-point/domain/src/lib/is-metering-point-id.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export function isMeteringPointId(maybeMeteringPointId: string): boolean {
```
frontend-specialisten, line 22: Consider renaming the function to something like this, as we can't tell from this function if it's a `meteringPointId` however what it tells is if it's a valid metering point id
```suggestion
export function isValidMeteringPointId(maybeMeteringPointId: string): boolean {
```

libs/dh/metering-point/domain/src/lib/is-metering-point-id.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+describe(isMeteringPointId.prototype.name, () => {
+  it('is valid', () => {
+    const _18Digits = '123456789000000000';
+
+    expect(isMeteringPointId(_18Digits)).toBeTruthy();
+  });
+
+  describe('NOT valid', () => {
+    it('empty string', () => {
```
frontend-specialisten, line 27: Consider writing your spec's like you're telling a story ex.
```suggestion
    it('should invalidate empty strings', () => {
```

frontend-specialisten (changes requested):
Just some nitpicks :)

dzhavat:
Feedback is addressed.

frontend-specialisten:
@dzhavat Question, should we include a custom form validator that uses this check as well or should I just create it in my branch?

dzhavat:
@frontend-specialisten This branch was not initially intended to be related to metering point search but figured out that the validation logic might be relevant. So I'd say lets create the custom validator in a separate branch. I'm also going to use the validation function in another branch for a route guard.

frontend-specialisten (approved):

PR #123, opened 2021-11-16
feat(dh): add route guard to metering point overview page

dzhavat:
### DataHub

- Create `dh/metering-point/routing` library
- Create `dh/shared/test-util-metering-point` library
- Add routing guard for `metering-point/:metering-point-id` path
- Redirect to `metering-point/search` if metering point id is invalid

### Workspace

- Allow `routing` library to import from `domain`

## References

- https://github.com/Energinet-DataHub/greenforce-frontend/issues/113

## PLEASE NOTE


libs/dh/metering-point/routing/src/lib/dh-metering-point-overview.guard.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+        },
+      ],
+    });
+    angularLocation = TestBed.inject(Location);
+    router = TestBed.inject(Router);
+  });
+
+  let angularLocation: Location;
+  let router: Router;
+  const validMeteringPointId = '123456789000000000';
```
frontend-specialisten, line 65: Would be nice to have valid / invalid metering point id's in some test-util. I can see multiple places where these would be useful :)

frontend-specialisten (approved):
LGTM, just a comment which could be considered :)

libs/dh/metering-point/routing/src/lib/dh-metering-point-overview.guard.spec.ts
dzhavat: I was thinking to extracting the metering point id to a library as well but thought that it might become too distant. Anyway, I just pushed commits that does that. You can see how it looks/works.

PR #155, opened 2021-11-17
feat(dh): show metering point id from route

dzhavat:
### DataHub

- Show metering point id on the page from URL
- Show spinner while waiting for data to be shown

### Watt

- Extend spinner component to support diameter value

## References

- #113

## PLEASE NOTE


libs/dh/metering-point/data-access-route/src/lib/dh-metering-point-data-access-route.store.ts
```diff
@@ -0,0 +1,29 @@
...
+import { dhMeteringPointIdParam } from '@energinet-datahub/dh/metering-point/routing';
+
+@Injectable()
+export class DhMeteringPointDataAccessRouteStore {
+  meteringPointId$: Observable<string> = this.route.params.pipe(
+    map((params) => params[dhMeteringPointIdParam])
+  );
+
+  constructor(private route: ActivatedRoute) {}
+}
```
LayZeeDK, line 29: suggestion: I've experimented with this and it seems that there's not much point in creating a local router store as it's almost a 1:1 mapping of `ActivatedRoute`: https://github.com/LayZeeDK/ngrx-router-component-store/blob/main/packages/router-component-store/src/lib/local-router-store.ts

My suggestion is to use `ActivatedRoute` directly instead of wrapping it. At least until we have a global router store that all components and component stores can use.

dzhavat: Makes sense at this point in time. Later we can move it again since the metering point id should not be displayed directly from the URL but instead used for making an API request.

frontend-specialisten (approved):
LGTM

PR #156, opened 2021-11-18
feat(dh): add breadcrumbs to overview page

dzhavat:
### DataHub

- Add basic breadcrumbs component to metering point overview page
- Update translation files

### Watt

- Define styling for `a` elements
- Set color in typography mixins

## References

- https://github.com/Energinet-DataHub/greenforce-frontend/issues/113

## PLEASE NOTE


libs/dh/metering-point/feature-overview/src/lib/breadcrumbs/dh-breadcrumbs.module.ts
```diff
@@ -0,0 +1,31 @@
...
+import { WattIconModule } from '@energinet-datahub/watt';
+
+import { DhBreadcrumbsComponent } from './dh-breadcrumbs.component';
+
+@NgModule({
+  declarations: [DhBreadcrumbsComponent],
+  exports: [DhBreadcrumbsComponent],
+  imports: [CommonModule, RouterModule, TranslocoModule, WattIconModule],
+})
+export class DhBreadcrumbsModule {}
```
LayZeeDK, line 31: issue: please rename the class to `DhBreadcrumbsScam` to make intent clearer

libs/dh/metering-point/feature-overview/src/lib/breadcrumbs/dh-breadcrumbs.component.ts
```diff
@@ -0,0 +1,29 @@
...
+@Component({
+  selector: 'dh-breadcrumbs',
+  templateUrl: './dh-breadcrumbs.component.html',
+  styleUrls: ['./dh-breadcrumbs.component.scss'],
+})
+export class DhBreadcrumbsComponent {
+  meteringPointAbsolutePath = ['/', dhMeteringPointPath];
+
+  @Input() meteringPointId = '';
+}
```
LayZeeDK, line 29: issue: let's test the breadcrumb through an integration test for the metering points overview component

LayZeeDK (changes requested):

libs/dh/metering-point/feature-overview/src/lib/breadcrumbs/dh-breadcrumbs.component.html
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<ng-container *transloco="let transloco; read: 'meteringPoint.overview'">
+  <a [routerLink]="meteringPointAbsolutePath">{{
+    transloco('breadcrumbs.meteringPoints')
+  }}</a>
+  <watt-icon name="right"></watt-icon>
+  {{ meteringPointId }}
+</ng-container>
```
LayZeeDK, line 23: suggestion: consider using the transloco attribute directive or pipe so that some content is shown before translatons are loaded

libs/dh/metering-point/feature-overview/src/lib/breadcrumbs/dh-breadcrumbs.module.ts
```diff
@@ -0,0 +1,31 @@
...
+import { WattIconModule } from '@energinet-datahub/watt';
+
+import { DhBreadcrumbsComponent } from './dh-breadcrumbs.component';
+
+@NgModule({
+  declarations: [DhBreadcrumbsComponent],
+  exports: [DhBreadcrumbsComponent],
+  imports: [CommonModule, RouterModule, TranslocoModule, WattIconModule],
+})
+export class DhBreadcrumbsScam {}
```
LayZeeDK, line 31: question: I'm not sure if we agreed to use inline SCAMs or to have them separately in `.module.ts` files. Do you remember?

dzhavat: By now I'm confused as well. My understanding is that we're using inline SCAMs when the component and its HTML/SCSS consist of only a few lines but if they grow bigger, then they are moved to separate files.

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      imports: [DhMeteringPointOverviewModule],
+      routes: [
+        {
+          path: dhMeteringPointPath,
+          children: [
+            {
+              canActivate: [DhMeteringPointOverviewGuard],
+              component: DhMeteringPointOverviewComponent,
+              path: `:${dhMeteringPointIdParam}`,
+            },
```
LayZeeDK, line 41: suggestion: consider how we could test the real route here so that the actual route is covered by a test and we don't have to keep actual routes and test routes in sync

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    });
+  });
+
+  let view: RenderResult<SpectacularAppComponent>;
+  const meteringPointId = '571313180400014077';
+
+  it('displays a link to the Metering point URL', async () => {
+    await view.navigate(`/${dhMeteringPointPath}/${meteringPointId}`);
+
+    const link: HTMLAnchorElement = await screen.findByRole('link');
```
LayZeeDK, line 54: suggestion: consider whether we should add more semantic meaning to the HTML structure, for example: should the breadcrumb contain a `<nav>` element?

LayZeeDK: suggestion: consider what we are going to do when more links are added to this component. How can we identify this link amongst multiple `<a>` elements? Hint: include the metering point GSRN in the link query.

LayZeeDK (changes requested):

libs/dh/metering-point/feature-overview/src/lib/breadcrumbs/dh-breadcrumbs.module.ts
dzhavat: Yes, we are. In cases where the HTML/SCSS is of considerable amount, I'd suggest using external files for them. We have done something similar in other parts.

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
dzhavat: > Hint: include the metering point GSRN in the link query.

What did you mean by this? The very first link (and the only one for now) doesn't include the GSRN in it. It only points to the metering point page. The GSRN is a plain text for now.

dzhavat: Just pushed a commit that can support multiple links but only tests the first one for now.

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
dzhavat: Could a solution be to move all routing setup from `DhMeteringPointShellModule` to `metering-point/routing` library? Then expose a `*ChildRoutesModule` that we import back to shell library. Then the `*ChildRoutesModule` can also be imported in other places where routing is necessary. Or am I misunderstanding the idea?

LayZeeDK: When I was looking at the current solution last week, I didn't see a need for a routing library at all. I think we could have both the guard and the route path in its corresponding feature library but expose the route path. This would allow us to have integration tests in this library alone covering actual routes, route guards, modules, and components.

libs/dh/metering-point/feature-overview/src/lib/breadcrumb/dh-breadcrumb.component.html
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<nav class="breadcrumb">
+  <a [routerLink]="meteringPointAbsolutePath">{{
+    'meteringPoint.overview.breadcrumb.meteringPoints' | transloco
```
LayZeeDK, line 19: question: I see now that there's no `read` parameter for the Transloco pipe. Having the scope in a separate parameter using the Transloco structural directive seemed more clean in my opinion. What's your take on this?

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  let view: RenderResult<SpectacularAppComponent>;
+  const meteringPointId = '571313180400014077';
+
+  it('displays a link to the Metering point URL', async () => {
+    await view.navigate(`/${dhMeteringPointPath}/${meteringPointId}`);
+
+    const [topLevelLink]: HTMLAnchorElement[] = await screen.findAllByRole(
+      'link'
+    );
```
LayZeeDK, line 57: issue: querying for the first link in a component is a fragile way of getting a particular link. Query for the link text instead. We might need to figure out how to load Transloco translation texts in tests to do this.

LayZeeDK (changes requested):

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
LayZeeDK: > What did you mean by this? The very first link (and the only one for now) doesn't include the GSRN in it. It only points to the metering point page. The GSRN is plain text for now.

Never mind. I remembered the navigation structure incorrectly but see comment in a different thread regarding this.

libs/dh/metering-point/feature-overview/src/lib/breadcrumb/dh-breadcrumb.component.html
LayZeeDK: We should probably figure out how to use translation texts in tests before deciding on a preferred Transloco declarable. See other thread.

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
LayZeeDK: Querying for the link text case insensitively is the preferred solution when using Angular Testing Library. Another option could be to query for the first link contained in a `<nav>` element.

libs/dh/metering-point/feature-overview/src/lib/breadcrumb/dh-breadcrumb.component.html
dzhavat: > Having the scope in a separate parameter using the Transloco structural directive seemed more clean in my opinion. 

I agree. It was using the directive initially but I was wrapping the whole template. Maybe we can still use the directive but only wrap the link instead, so the other data is shown. A downside is that there's a jump if the translation takes some time to load and the is shown after the GSRN number is already there.

LayZeeDK: Documentation: https://ngneat.github.io/transloco/docs/unit-testing/

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
dzhavat: > Another option could be to query for the first link contained in a <nav> element.

Is there a way to use HTML elements in these query selectors? I couldn't find it.

libs/dh/metering-point/feature-overview/src/lib/breadcrumb/dh-breadcrumb.component.html
frontend-specialisten: @dzhavat @LayZeeDK lets have a talk about this after daily tomorrow, I've already done a temp translation testing module in my branch

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
LayZeeDK: There's only a `selector` option for the `findByText` query.

libs/dh/metering-point/shell/src/lib/dh-metering-point-shell.module.ts
```diff
@@ -37,12 +37,14 @@ const routes: Routes = [
...
   {
     path: `:${dhMeteringPointIdParam}`,
-    component: DhMeteringPointOverviewComponent,
     canActivate: [DhMeteringPointOverviewGuard],
-  },
-  {
-    path: `:${dhMeteringPointIdParam}/child/:child-id`,
-    component: DhMeteringPointChildOverviewComponent,
+    children: [
+      { path: '', component: DhMeteringPointOverviewComponent },
```
LayZeeDK, line 42: issue: this route must have `pathMatch: 'full'`, or the following child route can never be matched

libs/dh/metering-point/feature-overview/src/lib/breadcrumb/dh-breadcrumb.component.ts
```diff
@@ -0,0 +1,41 @@
...
+import { RouterModule } from '@angular/router';
+import { TranslocoModule } from '@ngneat/transloco';
+import { WattIconModule } from '@energinet-datahub/watt';
+
+import { dhMeteringPointPath } from '../routing/dh-metering-point-path';
+
+@Component({
+  selector: 'dh-breadcrumb',
+  templateUrl: './dh-breadcrumb.component.html',
+  styleUrls: ['./dh-breadcrumb.component.scss'],
```
frontend-specialisten, line 28: Suggestion add `changeDetection: ChangeDetectionStrategy.OnPush`

frontend-specialisten (approved):

libs/dh/metering-point/feature-overview/src/lib/breadcrumb/dh-breadcrumb.component.ts
dzhavat: Done.

dzhavat:
@LayZeeDK It looks like I cannot merge unless you also review the PR. Will you have time to do that tomorrow?

frontend-specialisten:
I can do it 👍

PR #157, opened 2021-11-18
refactor: rename "metering point id" to "gsrn number"

dzhavat:
## DO NOT MERGE

This PR depends on #156 

### Workspace

- Rename "metering point id" to "gsrn number" where applicable

## References

- N/A

## PLEASE NOTE


LayZeeDK:
We should all discuss whether this is the right way to go.

dzhavat:
We decided to use our own domain on the front-end where metering point id is used instead of GSRN number.

PR #160, opened 2021-11-19
style: configure ESLint plugin:ngrx/component-store-strict

timdeschryver:
## Description

Configures the NgRx ESLint plugin specific for component-store.

## References

- #0000

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,7 +1,7 @@
 {
   "root": true,
   "ignorePatterns": ["**/*"],
-  "plugins": ["@nrwl/nx", "ngrx"],
+  "plugins": ["@nrwl/nx"],
   "overrides": [
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
@@ -188,9 +188,7 @@
     },
     {
       "files": ["*.ts"],
-      "rules": {
-        "ngrx/updater-explicit-return-type": "error"
-      }
+      "extends": ["plugin:ngrx/component-store-strict"]
     },
     {
       "files": ["*.js", "*.ts"],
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -188,9 +188,7 @@
     },
     {
       "files": ["*.ts"],
-      "rules": {
-        "ngrx/updater-explicit-return-type": "error"
-      }
+      "extends": ["plugin:ngrx/component-store-strict"]
```
timdeschryver, line 191: this config is added in the latest version.
It includes all component store rules, and configures them to give an error (instead of a warning).

LayZeeDK:
@timdeschryver see #161 
How can we exclude `.ts` files in the `tools` folder and in `apps/**/e2e-*` folders?

I tried this:
```json
    {
      "files": ["*.ts"],
      "excludedFiles": ["apps/**/e2e-*/**/*.ts", "tools/**/*"],
      "extends": ["plugin:ngrx/component-store-strict"]
    },
```

timdeschryver:
@LayZeeDK that should do it.

The problem here is that the linter has problems to initialize because it can't find a tsconfig file (which is set in the config of the NgRx Plugin). To fix this, you can provide your own config file by overriding the `parserOptions.project` property. 
This is also how the angular eslint plugin configurations are configured, maybe there's a better way though.

For now, since the component-store rules don't rely on type informations (which is why the project is required), I will remove the setting from the config. This means that only effects will require the parser options, but I'll have to think/try on it a bit more to cover all cases.

There's a commit incoming for this.

LayZeeDK (approved):

LayZeeDK:
Thank you for contributing to the Danish Green Energy Hub and Energy Origin, @timdeschryver 🙇‍♂️

dzhavat:
Yes, thank you for taking the time! 🙂

timdeschryver:
Thanks for testing out the NgRx Plugin within a NX workspace 🙂

PR #176, opened 2021-11-25
Xjakk nemid mitid

j4k0bk:
## References

- #0000


LayZeeDK (changes requested):
Lint check and unit tests failing

libs/ett/core/shell/src/lib/ett-shell.component.ts
```diff
@@ -16,14 +16,26 @@
...
+import {
+  AuthOidcHttp,
+  GetProfileResponse,
+  UserProfile,
+} from '@energinet-datahub/ett/auth/data-access-api';
 import { WattShellModule } from '@energinet-datahub/watt';
+import { WattButtonModule } from '@energinet-datahub/watt';
+import { MatIconModule } from '@angular/material/icon'; // TODO Import from Watt?
+import { MatMenuModule } from '@angular/material/menu'; // TODO Import from Watt?
+import { map, catchError } from 'rxjs';
```
LayZeeDK, line 35: issue: remove unused import
```suggestion
import { map, } from 'rxjs';
```

libs/ett/core/shell/src/lib/ett-shell.component.ts
```diff
@@ -35,27 +47,129 @@ const selector = 'ett-shell';
...
+      });
+  }
+
+  private onLogoutComplete(success: boolean) {
+    if (success) {
+      this.router.navigate(['/login']);
+    }
+  }
+
+  private onLogoutFailed(error: any) {
```
LayZeeDK, line 133: issue: very rarely should we use the `any` type. Use `unknown` and figure out the actual type as needed through assertions or conditionals.

libs/ett/core/shell/src/lib/ett-shell.component.ts
```diff
@@ -35,27 +47,129 @@ const selector = 'ett-shell';
...
+  private onGetProfileComplete(response?: GetProfileResponse) {
+    if (response?.success) {
+      this.profile = response.profile;
+      this.profileLoading = false;
+      // TODO Replace with observables:
+      this.change.markForCheck();
+    }
+  }
+
+  private onGetProfileFailed(error: any) {
```
LayZeeDK, line 157: issue: very rarely should we use the any type. Use unknown and figure out the actual type as needed through assertions or conditionals.

LayZeeDK (changes requested):
Lint review

libs/ett/core/shell/src/lib/ett-shell.component.ts
```diff
@@ -35,27 +47,129 @@ const selector = 'ett-shell';
...
+          display: flex;
+
+          h1 {
+            flex: auto;
+          }
+
+          .menu {
+            align-self: flex-end;
+          }
+        }
```
LayZeeDK, line 62: issue: our components don't use emulated view encapsulation to save bandwidth, so we should use BEM-style selectors to keep styles from bleeding through to other components, for example `${selector}__toolbar`, `${selector} h1`, and `${selector}__menu`.

libs/ett/core/shell/src/lib/ett-shell.component.ts
```diff
@@ -35,27 +47,129 @@ const selector = 'ett-shell';
...
       }
     `,
   ],
   template: `
-    <watt-shell>
+    <div *ngIf="profileLoading">Loading</div>
+
+    <watt-shell *ngIf="!profileLoading">
```
LayZeeDK, line 69: issue: the existing test case is failing now because:
- the shell is not displayed until the user has authenticated
- this component now has quite a few concerns, one of them requires an HTTP request, so we need to [use the `HttpClientTestingModule`](https://angular.io/guide/testing-services#testing-http-services) or replace `AuthOidcHttp` with a test double

LayZeeDK, line 69: suggestion: the loading state could be handled using the `*rxLet` structural directive similar to the `EttAuthenticationLinkComponent`

LayZeeDK (changes requested):
Implementation review

libs/ett/auth/data-access-api/src/lib/oidc/auth-oidc-http.service.ts
```diff
@@ -49,12 +64,26 @@ export class AuthOidcHttp {
 
   login(returnUrl: string): Observable<AuthOidcLoginResponse> {
     return this.http.get<AuthOidcLoginResponse>(
-      `${environment.apiBase}/oidc/login`,
+      `${environment.apiBase}/auth/oidc/login`,
```
LayZeeDK, line 67: praise: Nice! Ideally, we use the same paths locally as in a deployed environment as is the case here.

libs/ett/auth/data-access-api/src/lib/oidc/auth-oidc-http.service.ts
```diff
@@ -41,6 +41,21 @@ export interface AuthOidcLoginResponse {
...
 }
 
+export interface AuthOidcLogoutResponse {
+  readonly success: boolean;
+}
+
+export interface UserProfile {
+  readonly id: string;
+  readonly name: string;
+  readonly company?: string;
```
LayZeeDK, line 51: question: Can `company` be `null`? Then the type should be:
```suggestion
  readonly company: string | null;
```

libs/ett/core/shell/src/lib/ett-shell.component.ts
```diff
@@ -36,26 +48,112 @@ const selector = 'ett-shell';
...
+        display: flex;
+      }
+
+      ${selector}__toolbar h1 {
+        flex: auto;
+      }
+
+      ${selector}__toolbar .menu {
+        align-self: flex-end;
+      }
```
LayZeeDK, line 63: suggestion: It should also be possible to use nesting like so:
```suggestion
      ${selector}__toolbar {
        width: 100%;
        display: flex;
        
        h1 {
          flex: auto;
        }
        
        .menu {
          align-self: flex-end;
        }
      }
```

LayZeeDK (changes requested):
Let me know if you'd like me to refactor and test this implementation.

LayZeeDK:
Stale and out of scope.

PR #183, opened 2021-12-06
feat(ui-watt): rename button type -> variant and add native button type

frontend-specialisten:
## Description

- Rename button type -> variant
- Add native button type for buttons
- Add button types for dh metering point search form
- Add missing HttpClientModule for empty state stories

## References

- #0000


libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts
```diff
@@ -77,16 +82,15 @@ describe(DhMeteringPointOverviewComponent.name, () => {
 
   it('displays a link to the Metering point URL', async () => {
     await featureRouter.navigateByUrl(`~/${meteringPointId}`);
-
-    const [topLevelLink]: HTMLAnchorElement[] = await screen.findAllByRole(
-      'link'
+    await waitFor(
```
dzhavat, line 85: suggestion: instead of wrapping the whole code block in a `waitFor`, can we use the third argument of `findAllByRole` method to specify timeout i.e. 

```ts
const [topLevelLink]: HTMLAnchorElement[] = await screen.findAllByRole('link', undefined, { timeout: 3000 });
```

As a side note, in #180 the breadcrumbs component is not dependent on the API response, so the test will be simplified further.

dzhavat (approved):

dzhavat (approved):

PR #187, opened 2021-12-07
chore: remove mkdir of app-dh configuration folder since it's already created

frontend-specialisten:
remove mkdir of app-dh configuration folder since it's already created

## References

- #0000

### DIFF
--- a/.github/workflows/app-dh-cd.yml
+++ b/.github/workflows/app-dh-cd.yml
@@ -133,7 +133,6 @@ jobs:
       - name: Add configuration file
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


tsconfig.base.json
```diff
@@ -15,7 +15,6 @@
     "skipLibCheck": true,
     "skipDefaultLibCheck": true,
     "baseUrl": ".",
-    "resolveJsonModule": true,
```
dzhavat, line 18: question: is this intentional? Wouldn't it break importing the translations?

frontend-specialisten: If you look at the whole file, it was added twice

dzhavat (approved):

PR #188, opened 2021-12-07
Update mp client in bff

thbaan:
## Description

Updated Metering Point client packages wherein Reading Occurrence is changed from a string to an enum.

## References

- #0000


libs/dh/shared/data-access-api/src/lib/generated/v1/model/reading-occurrence.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+
+export type ReadingOccurrence = 'P1Y' | 'P1M' | 'PT1H' | 'PT15M';
+
+export const ReadingOccurrence = {
+    P1Y: 'P1Y' as ReadingOccurrence,
+    P1M: 'P1M' as ReadingOccurrence,
+    Pt1H: 'PT1H' as ReadingOccurrence,
+    Pt15M: 'PT15M' as ReadingOccurrence
```
dzhavat, line 20: question: Everything looks fine. Only one non-blocking thing. Do you think this can be configured to be all capital letters?

dzhavat (approved):

libs/dh/shared/data-access-api/src/lib/generated/v1/model/reading-occurrence.ts
dzhavat: We'll adjust the configuration if it becomes an issue.

LayZeeDK:
Don't you just love it when bots do most of the work?

LayZeeDK (approved):

dzhavat:
> Don't you just love it when bots do most of the work?

It's perfect! :)

PR #189, opened 2021-12-09
Chore/update nx (Angular 13.x)

frontend-specialisten:
## References

- #0000


migrations.json
```diff
@@ -0,0 +1,107 @@
+{
```
dzhavat, line 1: question: is this file necessary?

dzhavat:
https://github.com/Energinet-DataHub/greenforce-frontend/blob/669b0beb1a4956ab6e41e7e65cbc76cb467a5b7a/libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.spec.ts#L93-L103

@LayZeeDK This test [fails](https://github.com/Energinet-DataHub/greenforce-frontend/runs/4470030405?check_suite_focus=true#step:10:2055) with `Class constructor ComponentStore cannot be invoked without 'new'` error pointing to `local-router-component-store.ts` file. Do you have an idea what might be causing it?

image

libs/ui-watt/src/lib/components/button/watt-button.stories.ts
```diff
@@ -73,7 +73,9 @@ Button.args = {
 const iconButtonTemplate = (args: Partial<WattIconButtonComponent>) =>
   `<watt-icon-button icon="${args.icon}" [disabled]="${args.disabled}" label="Something meaningful"></watt-icon-button>`;
 
-const IconButton: Story<WattIconButtonComponent> = (args) => ({
+const IconButton: Story<WattIconButtonComponent | WattButtonComponent> = (
```
dzhavat, line 76: question: Was there a problem with `WattButtonComponent` not being specified?

LayZeeDK:
Did we provide the local component store in the component tree? Otherwise, try the GlobalRouterStore, it has a root provider.

libs/ui-watt/src/lib/components/button/watt-button.stories.ts
frontend-specialisten: Compilation error, if not specified we can't disable the desired controls because they are not part of WattIconButtonComponent

frontend-specialisten:
> Did we provide the local component store in the component tree? Otherwise, try the GlobalRouterStore, it has a root provider.

@LayZeeDK Yes the component store is provided directly inside of the component

LayZeeDK:
I'll look into it tomorrow unless we have sprint planning.

dzhavat:
@frontend-specialisten We're already behind 😀 https://blog.nrwl.io/redesigned-starter-page-storybook-6-4-eslint-8-and-more-in-nx-13-3-5507d7378360

frontend-specialisten:
> @frontend-specialisten We're already behind 😀 https://blog.nrwl.io/redesigned-starter-page-storybook-6-4-eslint-8-and-more-in-nx-13-3-5507d7378360

@dzhavat I will run the migrations again 😅

dzhavat (approved):

PR #190, opened 2021-12-10
Docs/getting started

frontend-specialisten:
## References

- #0000


docs/getting-started.md
```diff
@@ -0,0 +1,80 @@
+# Prerequisites
+
+- [We recommend using VOLTA to manage your JavaScript command-line tools](volta.sh)
+- Node.js® version 14.18.1
+- Yarn - Please use Yarn v1 instead of npm to install dependencies  
+  `npm install --global yarn@1.22.17`
```
dzhavat, line 6: suggestion: In order to minimize setup differences between machines, I think we should promote Volta more and say that's how work on UI should be done. So my suggestion is to remove the lines about Node.js and Yarn and ask people to install Volta. My understanding is that Volta will read the `package.json` file and figure out on its own which versions of Node.js and Yarn to install.

docs/getting-started.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Visual Studio Code - recommended extensions
+
+- Angular Language Service
+- Prettier - Code formatter
+- Nx Console
+- EditorConfig for VS Code
+- markdownlint
+- ESLint
```
dzhavat, line 15: suggestion: Can we add these extensions to the `extensions.json` file under `.vscode` folder in the repo so VS Code prompts the developer to install them?

.gitignore
```diff
@@ -28,7 +28,6 @@
 *.userosscache
 *.sln.docstates
 *.idea*
-.vscode/
```
frontend-specialisten, line 31: FYI: We exclude other files in .vscode later in this file

docs/getting-started.md
frontend-specialisten: They are added now, should we just remove this section?

dzhavat: Yes, I think so. Can be replaced with a sentence saying that the extensions exist and how to install them (I think there's a command in VS Code that can install the recommended extensions if the file exists).

dzhavat (approved):

PR #193, opened 2021-12-14
Feature/526 metering point secondary master data

johevemi:
## Description

Create secondary master data for metering point

 effectiveDate (date formatting) - will be added in date formatting issue #543

## References

- #526


libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.ts
```diff
@@ -28,6 +28,7 @@ import { LetModule } from '@rx-angular/template';
 import { DhMeteringPointDataAccessApiStore } from '@energinet-datahub/dh/metering-point/data-access-api';
 import { WattSpinnerModule } from '@energinet-datahub/watt';
 
+import { UiSecondaryMasterDataComponentScam } from './ui-secondary-master-data/ui-secondary-master-data.component';
```
dzhavat, line 31: suggestion: Quick comment before continuing with more thorough review - in order to be consistent with the file naming, let's remove the `ui-` prefix from both folder, components and class names. The `ui-` prefix is used only when creating libraries that contain components used in the UI. The components themselves don't have to use that prefix.

johevemi: I have changed the prefix to dh and removed the prefix from the folder

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
```diff
@@ -0,0 +1,266 @@
...
+        </div>
+        <div>
+          <dt class="watt-label">
+            {{ t('productionObligation') }}
+          </dt>
+          <dd class="watt-text-m">
+            <ng-container
+              *ngIf="
+                (secondaryMasterData?.productionObligation | typeof) ===
+                  'boolean';
```
dzhavat, line 59: suggestion: Let's move this logic to a getter property inside the component instead of creating a dedicated pipe. The pipe solution is valid but the getter is a simpler solution that works better in this case.

libs/dh/globalization/assets-localization/src/assets/i18n/en.json
```diff
@@ -46,6 +46,92 @@
         "meterNumber": "Meternumber"
       }
     },
+    "secondaryMasterData": {
+      "technicalSpecificationTitle": "Tekniske Specifikationer",
```
dzhavat, line 50: suggestion: When we don't know the translations for some text, we usually add [EN] at the end, e.g. `<text-in-Danish> [EN]`. Let's do that for these as well.

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  beforeEach(() => {
+    fixture = TestBed.createComponent(DhSecondaryMasterDataComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
```
dzhavat, line 62: suggestion: you can delete these lines. The tests below are better.

dzhavat (changes requested):

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.ts
```diff
@@ -0,0 +1,87 @@
...
+export interface MeteringPointIdentityTranslationKeys {
+  disconnectionType: string;
+  connectionType: string;
+  assetType: string;
+  productId: string;
+  unit: string;
+}
+
+@Component({
+  selector: 'dh-ui-secondary-master-data',
```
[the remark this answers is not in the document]
johevemi: Yes

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
```diff
@@ -0,0 +1,266 @@
...
+      <dl>
+        <div>
+          <dt class="watt-label">
+            {{ t('netSettlementGroup') }}
+          </dt>
+          <dd class="watt-text-m">
+            <ng-container
+              *ngIf="secondaryMasterData?.netSettlementGroup; else emptyField"
+            >
+              {{ secondaryMasterData?.netSettlementGroup }}
```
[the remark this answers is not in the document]
johevemi: It says in the story that it should just display the value. I think we should let it be for now and come back to it at a later date

dzhavat (approved):

PR #194, opened 2021-12-14
Update metering point client

thbaan:


libs/dh/shared/data-access-api/src/lib/generated/v1/model/metering-point-simple-cim-dto.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+import { ConnectionState } from './connection-state';
+
+
+export interface MeteringPointSimpleCimDto { 
+    meteringPointId?: string;
+    gsrnNumber?: string;
+    connectionState?: ConnectionState;
+    effectiveDate?: string;
+}
```
dzhavat, line 20: issue: In order to implement the child metering point overview table, we need more data which is not currently part of the interface. Looking at the sketch, I can identify the following data:

image

1. Metering Point Type
2. GRSN Number
3. Effective Date (Start Date?)
4. End Date (?)
5. Connection State

The interface contains data about 2 (`gsrnNumber`), 3 (`effectiveDate`) and 5 (`connectionState`) but it misses 1 (`meteringPointType`) and 4 (new property).

Is the metering point domain able to supply the missing data?

dzhavat (changes requested):

ZavezX (approved):

ZavezX:
@thbaan Approved, just a little conflict in csproj file.

libs/dh/shared/data-access-api/src/lib/generated/v1/model/metering-point-simple-cim-dto.ts
ZavezX: properties added

PR #200, opened 2021-12-15
Fix/tabs

frontend-specialisten:
## Description
Fixing `Error: Error trying to diff 'new QueryList<WattTabComponent>()'. Only arrays and iterables are allowed` of the tabs component.

image

The issue can be reproduced by opening the console here, and notice you can't change between the tabs: 
https://61765fc47451ff003afe62ff-qbglkshdkv.chromatic.com/?path=/story/components-tabs--tabs

## References

- #0000


libs/ui-watt/src/lib/components/tabs/tabs.component.ts
```diff
@@ -33,7 +34,12 @@ import { WattTabComponent } from './tab/tab.component';
   templateUrl: './tabs.component.html',
   encapsulation: ViewEncapsulation.None,
 })
-export class WattTabsComponent {
+export class WattTabsComponent implements AfterContentInit {
   @ContentChildren(WattTabComponent)
-  public readonly tabElements: QueryList<WattTabComponent> = new QueryList<WattTabComponent>();
+  private readonly tabElements: QueryList<WattTabComponent> = new QueryList<WattTabComponent>();
```
dzhavat, line 39: nitpick: `readonly` is not necessary when the property is `private`.

libs/ui-watt/src/lib/components/tabs/tabs.component.ts
```diff
@@ -33,7 +34,12 @@ import { WattTabComponent } from './tab/tab.component';
...
+  tabs: Array<WattTabComponent> = [];
+
+  public ngAfterContentInit(): void {
```
dzhavat, line 42: nitpick: `public` is probably unnecessary.

dzhavat (approved):

dzhavat (approved):

PR #208, opened 2021-12-20
update charges client to 1.0.20

ZavezX:
## Description

- Update Charges Client to `v1.0.20`

## References

- #976


dzhavat:
praise: Removing `undefined` and `null` from properties that can't be any of those values is amazing. Makes it much easier to work with those properties. Only if we could do the same with the Metering point DTO 😁

PR #211, opened 2021-12-21
update charges client to version 1.0.21

ZavezX:
## Description

As a better solution for handling null values in auto generated typescript files is made in #210 the charges client is updated to revert back to record and package version is updated to latest build.

## References

- #976


frontend-specialisten (approved):
LGTM... however if this is a revert couldn't we just use an earlier version?

PR #214, opened 2021-12-22
feat(dh): configure Danish localization and add custom date pipe

dzhavat:
### DataHub

- Configure localization by setting locale to Danish
- Add custom `dhDate` pipe that displays formatted dates in Danish time-zone

## References

- https://github.com/Energinet-DataHub/geh-metering-point/issues/543


ZavezX (approved):
Looks good.

PR #218, opened 2021-12-23
Lki/fix nullable properties in genereated ts models

ZavezX:
## Description

Extend schema to hold nullable information for reference objects in openapi.


ZavezX:
We will wait for open api to be updated to support our needs to get rid of null and undefined attributes on non-nullable reference types.

PR #223, opened 2022-01-03
Bug/581 net settlement group should be number

johevemi:
## Description

Correct netSettlementGroup to display number instead of text
Correct spelling mistake

image

## References

#581 #582

### DIFF
--- a/libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.ts
+++ b/libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.ts
@@ -16,7 +16,10 @@
  */
 import { CommonModule } from '@angular/common';
 import { Component, Input, NgModule } from '@angular/core';
-import { MeteringPointCimDto, NetSettlementGroup } from '@energinet-datahub/dh/shared/data-access-api';
+import {
+  MeteringPointCimDto,
+  NetSettlementGroup,
+} from '@energinet-datahub/dh/shared/data-access-api';
 import { DhSharedUiDateTimeModule } from '@energinet-datahub/dh/shared/ui-date-time';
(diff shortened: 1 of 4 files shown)


libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.ts
```diff
@@ -75,13 +76,17 @@ export class DhSecondaryMasterDataComponent {
       meteringPoint?.productId ?? ''
     }`;
     const unit = `meteringPoint.unit.${meteringPoint?.unit ?? ''}`;
+    const netSettlementGroupAsNumber = `meteringPoint.secondaryMasterData.netSettlementGroupAsNumber.${
+      meteringPoint?.netSettlementGroup ?? ''
+    }`;
```
dzhavat, line 81: suggestion: It feels strange to map the `netSettlementGroup` value to a translation string. Especially if the display value is always going to be a number (e.g. 6).

We can instead use a getter property and map the `netSettlementGroup` value to a number in there before displaying it on the page. For example:

```ts
  get netSettlementGroupAsNumber(): number | undefined {
    switch (this.#secondaryMasterData?.netSettlementGroup) {
      case NetSettlementGroup.Zero:
        return 0;
      case NetSettlementGroup.One:
        return 1;
      case NetSettlementGroup.Two:
        return 2;
      case NetSettlementGroup.Three:
        return 3;
      case NetSettlementGroup.Six:
        return 6;
      case NetSettlementGroup.NinetyNine:
        return 99;
      default:
        return;
    }
  }
```

johevemi: okay i will change it to this

ZavezX (approved):

dzhavat (approved):

PR #238, opened 2022-01-06
set min height on tab body

ZavezX:
## Description

Does what it says in the title👆

RJgI4088y3

## References

- #548


MrDach (commented):
Won't this still result in the same bug whenever the content of a tab exceeds the min-height?
Didn't see that you started this so I attempted a different solution here:
https://github.com/Energinet-DataHub/greenforce-frontend/pull/242

ZavezX:
> Won't this still result in the same bug whenever the content of a tab exceeds the min-height? Didn't see that you started this so I attempted a different solution here: #242

Your solution is the better one, so I'll close this one.

PR #250, opened 2022-01-07
Make table styling global

MrDach:
## Description

Moved the table styling into _table.scss in the design system.
Also added a minor tweak to the padding of cells in the charges table.

## References

- #542


libs/dh/metering-point/feature-overview/src/lib/child-metering-point-tab-content/dh-child-metering-point-tab-content.component.scss
```diff
@@ -39,7 +31,3 @@
 .multi-line-cell {
```
frontend-specialisten, line 31: This class also sounds like being much related to the styling of the table, consider moving this as well.

frontend-specialisten (changes requested):
Looks good, only a single consideration :)

frontend-specialisten (approved):

PR #252, opened 2022-01-10
Charges spinner styling

MrDach:
## Description

Added styling to the loading spinner in the charges tab.

Note: This could be combined with the styling from the overview loading spinner but for now it has just been duplicated.

## References

- #552


libs/dh/metering-point/feature-overview/src/lib/charges/dh-charges.component.html
```diff
@@ -15,7 +15,7 @@
 limitations under the License.
 -->
 <ng-container *rxLet="isLoading$ as isLoading">
-  <div *ngIf="isLoading" class="overview__spinner">
+  <div *ngIf="isLoading" class="charges__spinner">
```
frontend-specialisten, line 18: Are there any reasons to prefix with charges? Consider removing it as the component properly has isolated styling anyways :)

frontend-specialisten (changes requested):

libs/dh/metering-point/feature-overview/src/lib/charges/dh-charges.component.html
MrDach: It was prefixed with overview where I hijacked it from but makes sense - I'll remove it :)

frontend-specialisten (approved):

PR #256, opened 2022-01-11
feat(dh): unify usage of em dash fallback pipe

dzhavat:
### DataHub

- unify usage of displaying an empty fallback value
- rename `emptyValue` pipe to `emDashFallback`
- update `yesNo` pipe to return on `null`/`undefined`
- adjust tests

## References

- https://github.com/Energinet-DataHub/geh-metering-point/issues/570


libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
```diff
@@ -160,10 +136,8 @@
           {{ t('capacity') }}
         </dt>
         <dd class="watt-text-m">
-          <ng-container *ngIf="secondaryMasterData?.capacity; else emptyField">
-            {{ secondaryMasterData?.capacity }}
-            {{ t('units.kilowatt') }}
-          </ng-container>
+          {{ secondaryMasterData?.capacity | emDashFallback }}
```
frontend-specialisten, line 139: Question: wouldn't this change cause kilowatt to be shown even though capacity is not defined?

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
```diff
@@ -193,12 +165,8 @@
           {{ t('ratedCapacity') }}
         </dt>
         <dd class="watt-text-m">
-          <ng-container
-            *ngIf="secondaryMasterData?.ratedCapacity; else emptyField"
-          >
-            {{ secondaryMasterData?.ratedCapacity }}
-            {{ t('units.kilowatt') }}
-          </ng-container>
+          {{ secondaryMasterData?.ratedCapacity | emDashFallback }}
```
frontend-specialisten, line 168: Question: wouldn't this change cause kilowatt to be shown even though ratedCapacity is not defined?

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
```diff
@@ -212,12 +180,8 @@
           {{ t('ratedCurrent') }}
         </dt>
         <dd class="watt-text-m">
-          <ng-container
-            *ngIf="secondaryMasterData?.ratedCurrent; else emptyField"
-          >
-            {{ secondaryMasterData?.ratedCurrent }}
-            {{ t('units.ampere') }}
-          </ng-container>
+          {{ secondaryMasterData?.ratedCurrent | emDashFallback }}
```
frontend-specialisten, line 183: Question: wouldn't this change cause ampere to be shown even though ratedCurrent is not defined?

libs/dh/metering-point/feature-overview/src/lib/identity/dh-metering-point-identity.component.html
```diff
@@ -19,28 +19,23 @@
...
   <h1 [title]="t('gsrnNumber')">{{ meteringPoint?.gsrnNumber }}</h1>
 
   <div class="watt-text-s">
     <span [title]="t('connectionState')">
-      <ng-container *ngIf="meteringPoint?.connectionState; else emptyField">
-        <dh-status-badge
-          [connectionState]="meteringPoint?.connectionState"
-        ></dh-status-badge>
-      </ng-container>
+      <dh-status-badge
```
frontend-specialisten, line 36: Question: before if connectionState wasn't defined a fallback value would be shown. Will this now going to be just blank and is that intentional?

frontend-specialisten (changes requested):

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
dzhavat: Yes. good catch. One solution could be to make the `capacity` a param to the `units.kilowatt` translation. Another solution is to make a getter and return the display value from it or return `undefined` if `capacity` is `null`/`undefined`. A third solution is to also create a directive that displays the em dash.

Do you have a preference? I think I'll try the getter first.

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
dzhavat: Yes. good catch. I think the options are similar to my previous comment above (under `capacity`).

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
dzhavat: Yes. good catch. I think the options are similar to my previous comment above (under `capacity`).

libs/dh/metering-point/feature-overview/src/lib/identity/dh-metering-point-identity.component.html
dzhavat: According to the `MeteringPointCimDto` interface, the `connectionState` property will always have a value. That was the reason for removing the fallback. From the code it loos like `connectionState` Input property might end up with `undefined` but that is mostly due to how the `meteringPoint` property is typed.

Considering all of this, do you think it's still relevant to display a fallback?

libs/dh/metering-point/feature-overview/src/lib/secondary-master-data/dh-secondary-master-data.component.html
dzhavat: Went with the first option.

libs/dh/metering-point/feature-overview/src/lib/identity/dh-metering-point-identity.component.html
frontend-specialisten: No I think it's totally fine :)

frontend-specialisten (approved):

PR #266, opened 2022-01-12
Feat/msal interceptors

frontend-specialisten:
## Description

Add MSAL interceptor, so our HTTP requests includes a bearer token

## References

https://github.com/Energinet-DataHub/geh-aggregations/issues/559


libs/dh/auth/feature-msal/src/lib/b2c-config.ts
```diff
@@ -66,21 +55,25 @@ export function MSALInstanceFactory(
...
 
-export function MSALGuardConfigFactory(): MsalGuardConfiguration {
+export function MSALGuardConfigFactory(
+  config: DhB2CEnvironment
+): MsalGuardConfiguration {
   return {
     interactionType: InteractionType.Redirect,
     authRequest: {
-      scopes: [...apiConfig.scopes],
+      scopes: [config.clientId],
```
dzhavat, line 76: question: I wonder about the value difference between the old `apiConfig.scopes` and `config.clientId`. Usually the scope is given in the form of a `user.read` value but the `clientId` is quite different. Do you think there's a problem with that?

frontend-specialisten: I don't think so, but really we can't do any kind of tests before the backend is implementing there security. However what I found out was setting the scope as the clientId is the only way for us to receive and accessToken (at least not at this moment, until they create custom scopes). 

See https://docs.microsoft.com/en-us/azure/active-directory-b2c/access-tokens#openid-connect-scopes

`00000000-0000-0000-0000-000000000000 - Using the client ID as the scope indicates that your app needs an access token that can be used against your own service or web API, represented by the same client ID.`

https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2315#issuecomment-773411964

frontend-specialisten: So really I see this as a first draft, until we can do some actually testing on this because I don't really have lots of experience with B2C

dzhavat: Nice finding about the usage of `clientId` in scopes.

dzhavat (approved):

PR #273, opened 2022-01-13
Add sonarcloud to dh frontend app

Sondergaard:
## Description

Add Sonar Cloud to DH3 Frontend App

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -150,11 +150,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+          fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
       - name: SonarCloud Scan
         uses: SonarSource/sonarcloud-github-action@master
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
 
   e2e:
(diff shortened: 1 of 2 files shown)


.github/workflows/frontend-ci.yml
```diff
@@ -144,6 +144,19 @@ jobs:
...
 
+  sonarcloud:
+    name: SonarCloud
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
+      - name: SonarCloud Scan
+        uses: SonarSource/sonarcloud-github-action@master
```
LayZeeDK, line 155: suggestion: Use a version range tag instead of `latest` to prevent accidental breakage
```suggestion
        uses: SonarSource/sonarcloud-github-action@v1.6
```

sonar-project.properties
```diff
@@ -0,0 +1,12 @@
+sonar.projectKey=geh-frontend-app
+sonar.organization=energinet-datahub
+
+# This is the name and version displayed in the SonarCloud UI.
+# sonar.projectName=DH3 Frontend App
+# sonar.projectVersion=1.0
+
+# Path is relative to the sonar-project.properties file. Replace "\" #by "/" on Windows.
+sonar.sources=./apps/dh/app-dh,./libs/dh
```
LayZeeDK, line 9: suggestion: Consider adding Watt and shared code
```suggestion
sonar.sources=./apps/dh/app-dh,./libs/dh,./libs/ui-watt,./libs/gf
```

LayZeeDK (changes requested):

sonar-project.properties
Sondergaard: I would rather have them as separate projects in sonarcloud. I assume that they have there own build workflow in GitHub - I might be wrong since this repo is new to me.

LayZeeDK: Watt itself (`libs/ui-watt`) is not currently independently buildable as its consumed directly by frontend applications in this monorepo. It does, however, have a deployment workflow for its living documentation (Storybook).

Shared libraries (`libs/gf`) are used by the DataHub frontend so it should be scanned by SonarCloud either here or elsewhere.

Sondergaard:
@LayZeeDK - your comment / commits have been applied. Do you have any other comments?

LayZeeDK (approved):
🚀

PR #279, opened 2022-01-14
refactor(dh): move identity and master data components to feature library

dzhavat:
### DataHub

- Move `identity`, `primary-master-data` and `secondary-master-data` to separate library

## References

- https://github.com/Energinet-DataHub/geh-metering-point/issues/580


frontend-specialisten (approved):
LGTM, only comment would be if this really should be just UI instead of feature since overview is still requesting the data?

dzhavat:
Good point. I was thinking whether it's a good idea to keep loading of metering point data in `dh-metering-point-overview.component.ts` but inject `DhMeteringPointDataAccessApiStore` in `dh-metering-point-identity-and-master-data.component.ts` and get the metering point data that way. This will make the library more like a feature.

I think that will work. Should we try it?

frontend-specialisten (approved):

PR #283, opened 2022-01-14
refactor(dh): move charges and child metering points tabs to feature library

dzhavat:
### DataHub

- Move `charges` and `child-metering-points` tabs to a separate library
- Move routing artifacts to a `metering-point/routing` library

## References

- https://github.com/Energinet-DataHub/geh-metering-point/issues/580


frontend-specialisten (approved):
LGTM, however it seems like we have some testing depth of the "charges" component.

PR #286, opened 2022-01-14
Set up basic table example with sorting

MrDach:
### Watt Design System

* Set up example of Angular Material table with sorting in Watt design system.
### Workspace

* Exclude `+storybook` directories from SonarCloud coverage report

## References

- https://github.com/Energinet-DataHub/geh-aggregations/issues/543


.github/workflows/frontend-ci.yml
```diff
@@ -158,6 +158,7 @@ jobs:
             -Dsonar.projectKey=geh-frontend-app
             -Dsonar.organization=energinet-datahub
             -Dsonar.sources=./apps/dh/app-dh,./libs/dh,./libs/ui-watt,./libs/gf
+            -Dsonar.exclusions=**/+storybook
```
frontend-specialisten, line 161: Question: Does this mean every sonar related scanning is disabled for our storybook related files. Shouldn't it only be the coverage we wanted to exclude?

dzhavat: Good point. There was a setting to exclude only coverage. Will give it a try.

frontend-specialisten (approved):

PR #305, opened 2022-01-24
Chore/update storybook

frontend-specialisten:
## Description
Update storybook to latest version due to security warning of immer:
https://github.com/storybookjs/storybook/issues/16093

`react-dom` has been added to the resolutions as it otherwise won't be installed, by some reasons: 
https://github.com/storybookjs/storybook/issues/13194

_If you have any suggestions how to fix this, it would be great!_  

## References

https://github.com/Energinet-DataHub/greenforce-frontend/security/dependabot/yarn.lock/immer/closed


libs/ui-watt/project.json
```diff
@@ -25,8 +25,9 @@
     "storybook": {
       "executor": "@nrwl/storybook:storybook",
       "options": {
-        "uiFramework": "@storybook/angular",
         "port": 4400,
+        "projectBuildConfig": "app-dh",
```
dzhavat, line 29: suggestion: By looking at a spec file in Nx (https://github.com/nrwl/nx/blob/dcb839c49d102f5d50b5d54875750120c2090faf/packages/storybook/src/migrations/update-13-4-6/__snapshots__/set-project-build-config.spec.ts.snap#L109), it seems like `projectBuildConfig` should be set to something like `ui-watt:build-storybook`.

libs/ui-watt/project.json
```diff
@@ -41,8 +42,9 @@
       "executor": "@nrwl/storybook:build",
       "outputs": ["{options.outputPath}"],
       "options": {
+        "outputPath": "dist/designsystem",
+        "projectBuildConfig": "app-dh",
```
dzhavat, line 46: suggestion: By looking at a spec file in Nx (https://github.com/nrwl/nx/blob/dcb839c49d102f5d50b5d54875750120c2090faf/packages/storybook/src/migrations/update-13-4-6/__snapshots__/set-project-build-config.spec.ts.snap#L109), it seems like `projectBuildConfig` should be set to something like `ui-watt:build-storybook`.

libs/ui-watt/.storybook/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "emitDecoratorMetadata": true
   },
-
+  "files": ["../../../apps/dh/app-dh/src/polyfills.ts"],
```
dzhavat, line 6: question: why is this needed?

package.json
```diff
@@ -141,6 +141,7 @@
     "typescript": "4.4.4"
   },
   "resolutions": {
+    "react-dom": "17.0.2",
```
dzhavat, line 144: nitpick: I don't have a better solution to this other than maybe adding the dependency to `devDependencies`. But I'm also not sure if that will fix the warning.

dzhavat (approved):

Issue #295, opened 2022-01-18
Support state colors of icon component


MadsBloendAndersen:
### Description
Expand design system with all state colors

State colors
- Danger 
- Warning 
- Success
- Info

![image](https://user-images.githubusercontent.com/72008712/149950225-bad3a737-1bb4-4d13-bd82-96cba126b144.png)

### Acceptance criteria
- [ ] design system is expanded with all state colors of icon component with documentation and showcase

PR #313, opened 2022-01-25
feat(watt): extend icon to support state colors

dzhavat:
### Watt Design System
- Extend icon to support state colors
- Update Storybook to showcase examples

image

### DataHub
- Refactor component to use state property on icon component

## References

- https://github.com/Energinet-DataHub/greenforce-frontend/issues/295

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -158,7 +158,8 @@ jobs:
             -Dsonar.projectKey=geh-frontend-app
             -Dsonar.organization=energinet-datahub
             -Dsonar.sources=apps/dh/app-dh,libs/dh,libs/ui-watt,libs/gf
-            -Dsonar.coverage.exclusions=libs/ui-watt/**/+storybook/*,libs/ui-watt/**/*.stories.ts
+            -Dsonar.exclusions=**/*.spec.ts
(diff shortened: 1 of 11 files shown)


frontend-specialisten (dismissed):

frontend-specialisten (approved):

MadsBloendAndersen:
@dzhavat is it on purpose that there is a checkmark inside each state? Shouldn't the icon also reflect the state?

dzhavat:
@MadsBloendAndersen Yes, the checkbox icon is used intentionally to show that we can set any state on an icon. I can also understand your argument that the icon reflects the state (in our case it will be the other way, i.e. the state will reflect the icon because the icon name is required but the state is not. That's a detail). So the `success` icon always shows in "Success" state, etc. Good point. Maybe we should consider it. What do you think @frontend-specialisten

frontend-specialisten:
I think that we properly could have some more sensible default states of some of the icons like succes, warnings etc. However I only think that they should be defaults so you still have the flexibility to overwrite it.

dzhavat:
@MadsBloendAndersen @frontend-specialisten I just opened #326 which adds default state to these four icons (`success`, `danger`, `warning` and `info`) with the option to overwrite it.

PR #325, opened 2022-01-27
fix(dh): handle cancelling of signup flow

frontend-specialisten:
## Description

Handle cancelling of signup flow by reloading the page, to get the application / b2c instance in fresh state to ensure the user is redirected back to the login page. 

## References

https://github.com/Energinet-DataHub/geh-aggregations/issues/575

### DIFF
--- a/libs/dh/auth/feature-msal/src/lib/b2c-config.ts
+++ b/libs/dh/auth/feature-msal/src/lib/b2c-config.ts
@@ -46,13 +46,20 @@ export function MSALInstanceFactory(
     system: {
       loggerOptions: {
         loggerCallback: (logLevel: LogLevel, message: string) => {
-          console.log(message);
+          reloadOnLoginFailed(message);
         },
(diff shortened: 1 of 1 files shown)


libs/dh/auth/feature-msal/src/lib/b2c-config.ts
```diff
@@ -55,6 +55,15 @@ export function MSALInstanceFactory(
   });
 }
 
+function reloadOnLoginFailed(error: string) {
+  const loginFailed = error.includes(
+    'Error - Guard - error while logging in, unable to activate'
```
dzhavat, line 60: question: is this error message always in English? Also for browsers that are set to a different locale?

frontend-specialisten: @dzhavat yes the translation is an internal error message from the MSAL guard, so it's not a message for the users

dzhavat (approved):

PR #329, opened 2022-01-28
Updated EnergyOrigin release workflow

MartinSchmidt:
## Description
- Updated EnergyOrigin release workflow
- Moved EO infrastructure to build folder.
- Updated workflow build containers and releases helm chart with version from the Chart.yaml an values.yaml files in the chart folder.
- Added version increment check for container and chart when the EO app is affected.
- Add `APP_FOLDER` parameter to host container image defaulting to workflow build artifacts folder

## References
- https://github.com/Energinet-DataHub/energy-origin/issues/313


LayZeeDK (changes requested):
question: Can we change the container name to `greenforce-frontend-eo-app`?

.github/workflows/ett.yml
```diff
@@ -117,119 +117,52 @@ jobs:
...
-    needs: app
     outputs:
-      release_tag: ${{ steps.build_and_push.outputs.RELEASE_VERSION }}
-    runs-on: ubuntu-latest
+      chart_version: ${{ steps.get_semver.outputs.semver }}
+      containers: >-
+        [
+          { 
+            "deployment_name": "app", 
+            "dockerfile": "infrastructure/eo/host/Dockerfile",
```
LayZeeDK, line 129: suggestion: If we have time, we would like to change the path `infrastructure/eo/host` to `build/infrastructure/eo/host` but then I think we need to change a configuration setting in `eo-base-environment` as well.

MartinSchmidt: Once it has moved to the new workflow, this dependency is removed from the base-environment, and then this can be easily changed locally.

MartinSchmidt:
> question: Can we change the container name to `greenforce-frontend-eo-app`?

I have added the ability to change the name, and changed it to: **eo-frontend-app**

LayZeeDK:
@MartinSchmidt 
Missing:
- PR description of what the change is supposed to do
- Running and working demonstration of workflows before this is merged

MartinSchmidt:
> @MartinSchmidt Missing:
> 
>     * [x]  PR description of what this change is supposed to do
> 
>     * [ ]  I suggest we use kebab-case rather than snake_case similar to workflow configurations (`runs-on`) and names of parameters of most first class GitHub Actions (`setup-node`, `node-version`, `restore-keys`, `if-no-files-found`)
> 
>     * [ ]  Running and working demonstration of workflows before this is merged
> 
>     * [x]  Resolve merge conflict of `docs/getting-started.md`

In regards to casing, most of the GreenForce art uses snake_case when writing the names of steps and jobs in the workflows but kebab-case when writing the names on the workflow and action files, I suggest we keep to that convention, maybe even writes it into an official guideline.

MartinSchmidt:
> @MartinSchmidt Missing:
> 
>     * [x]  PR description of what this change is supposed to do
> 
>     * [x]  I suggest we use kebab-case rather than snake_case similar to workflow configurations (`runs-on`) and names of parameters of most first class GitHub Actions (`setup-node`, `node-version`, `restore-keys`, `if-no-files-found`)
> 
>     * [ ]  Running and working demonstration of workflows before this is merged
> 
>     * [x]  Resolve merge conflict of `docs/getting-started.md`

Ran the job in a *Fork*, only last step fails because I do not have a base-env to update, but that part is testen in the eo-auth repo.

image

LayZeeDK:
- [ ] Merge https://github.com/Energinet-DataHub/greenforce-frontend/pull/377 first to fix semantic version check?

LayZeeDK (changes requested):

.github/workflows/ett.yml
```diff
@@ -117,119 +117,81 @@ jobs:
...
-
-  container_merge:
-    name: 'Merge: Host container'
-    if: ${{ github.ref == 'refs/heads/main' }}
-    needs: app
     outputs:
-      release_tag: ${{ steps.build_and_push.outputs.RELEASE_VERSION }}
-    runs-on: ubuntu-latest
+      chart_version: ${{ steps.get_semver.outputs.semver }}
+      containers: >-
```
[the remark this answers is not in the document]
MartinSchmidt: Because of the way yaml works: https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines, removes the feed line from the input.

    >, |: "clip": keep the line feed, remove the trailing blank lines.
    >-, |-: "strip": remove the line feed, remove the trailing blank lines.
    >+, |+: "keep": keep the line feed, keep trailing blank lines.

LayZeeDK (approved):
🍰

PR #330, opened 2022-01-28
Eo288 topbar button and logo update

Klemmensen:
## Description
- Adds new styling for the start button inside the topbar.

## References
- https://github.com/Energinet-DataHub/energy-origin/issues/288


libs/eo/landing-page/shell/src/lib/eo-landing-page-header.component.ts
```diff
@@ -57,12 +57,14 @@ const selector = 'eo-landing-page-header';
...
+          // This is equivalent to a normal size watt button width
           min-width: 6.25rem;
-          height: 2.75rem;
+          height: calc(5 * var(--watt-space-s)); // 40px
+          padding: var(--watt-space-s) var(--watt-space-m) var(--watt-space-s)
+            var(--watt-space-m); // 8px 16px 8px 16px
 
           // Adjustment locally to fit the position of the text vertically
-          line-height: 2.75rem;
+          line-height: calc(3 * var(--watt-space-s)); // 24px;
```
LayZeeDK, line 67: suggestion: If line height is no longer the button size, use 1.5 text size. Verify that the following works and adjust the comment above.
```suggestion
          line-height: 1.5;
```

libs/eo/landing-page/shell/src/lib/eo-landing-page-header.component.ts
```diff
@@ -57,12 +57,14 @@ const selector = 'eo-landing-page-header';
           background: var(--watt-color-primary);
           color: var(--watt-color-primary-contrast);
 
-          // This is equivalent to a normal size watt button
+          // This is equivalent to a normal size watt button width
           min-width: 6.25rem;
-          height: 2.75rem;
+          height: calc(5 * var(--watt-space-s)); // 40px
```
LayZeeDK, line 62: suggestion: (1) Use multiples of the XS space unit unless the size is relative to another component. (2) Let's refrain from using comments with values that can get out of sync.
```suggestion
          height: calc(10 * var(--watt-space-xs));
```

libs/eo/landing-page/shell/src/lib/eo-landing-page-header.component.ts
```diff
@@ -57,12 +57,14 @@ const selector = 'eo-landing-page-header';
           background: var(--watt-color-primary);
           color: var(--watt-color-primary-contrast);
 
-          // This is equivalent to a normal size watt button
+          // This is equivalent to a normal size watt button width
           min-width: 6.25rem;
```
LayZeeDK, line 61: suggestion: inline the comment so that it's clear that it's not for the whole section of properties. Alternatively, use a named Sass or CSS variable to make intent clear.
```suggestion
          min-width: 6.25rem; // This is equivalent to a normal size watt button width
```

libs/eo/landing-page/shell/src/lib/eo-landing-page-header.component.ts
```diff
@@ -57,12 +57,14 @@ const selector = 'eo-landing-page-header';
...
-          height: 2.75rem;
+          height: calc(5 * var(--watt-space-s)); // 40px
+          padding: var(--watt-space-s) var(--watt-space-m) var(--watt-space-s)
+            var(--watt-space-m); // 8px 16px 8px 16px
```
LayZeeDK, line 64: question: What does this match? Add a comment referring to the Inset Squish or whatever style its matching.
issue: Remove the comment with a value that can get out of sync.

LayZeeDK (changes requested):

libs/eo/landing-page/shell/src/lib/eo-landing-page-header.component.ts
```diff
@@ -57,12 +57,14 @@ const selector = 'eo-landing-page-header';
           background: var(--watt-color-primary);
           color: var(--watt-color-primary-contrast);
 
-          // This is equivalent to a normal size watt button
+          // This is equivalent to a normal size watt button width
           min-width: 6.25rem;
-          height: 2.75rem;
+          height: calc(5 * var(--watt-space-s)); // 40px
+          padding: var(--watt-space-s) var(--watt-space-m) var(--watt-space-s)
```
frontend-specialisten, line 62:
```suggestion
          @include watt.space-inset-squish-m;
```

libs/eo/landing-page/shell/src/lib/eo-landing-page-header.component.ts
frontend-specialisten: Question: Why is a hardcoded height needed? :)

libs/eo/landing-page/shell/src/lib/eo-landing-page-header.component.ts
frontend-specialisten: Consideration: The intension of the spacing variables has only been to use them for spacing (padding / margin). Using the variables for other stuff like line-height can potentially end up given us unintended side-effects. Just to keep that in mind :)

frontend-specialisten (changes requested):

LayZeeDK (approved):
🚀

frontend-specialisten (approved):
I would prefer the mixin used, however it's not a deal breaker for me :)

LayZeeDK:
@frontend-specialisten Are the Inset Squish mixins exported so that they available outside of the Watt library? It seemed we were only able to use the Inset Squish classes.

frontend-specialisten:
@LayZeeDK they should be, if they're not it's a clear mistake. So if thats the case I will create a bug and fix it.

Issue #293, opened 2022-01-18
Add message box


MadsBloendAndersen:
### Description
expand design system to be able to show message box with all state colors.

![image](https://user-images.githubusercontent.com/86852536/150989914-9bcaf1c8-8e1d-4887-9271-344f1f6a5acc.png)

### Acceptance criteria
- [x] add message box is added to design system with documentation and showcase
- [x] Single and multiple messages should be supported, see #297

### Tech notes:
For inspiration `bootstrap` has a similar "component" called `alert`: https://getbootstrap.com/docs/5.1/components/alerts/

PR #336, opened 2022-02-02
feat(watt): add validation message component

dzhavat:
### Watt Design System
- Add `watt-validation-message` component
- Showcase and document component in Storybook

**Note**: The text color is set to primary because setting the text to a state color on light state color background doesn't have enough contrast to meet the accessibility standards.

image

## References

- https://github.com/Energinet-DataHub/greenforce-frontend/issues/293

### DIFF
--- a/libs/ui-watt/src/lib/components/validation-message/watt-validation-message.component.scss
+++ b/libs/ui-watt/src/lib/components/validation-message/watt-validation-message.component.scss
@@ -1,0 +1,43 @@
+@use '@energinet-datahub/watt/utils' as watt;
+
+:host {
+  border-radius: 4px;
+  display: block;
+  @include watt.space-inset-squish-xs;
+
+  &.watt-validation-message-info {
+    background-color: var(--watt-color-state-info-light);
+
+    & .watt-label,
+    & .watt-text-xs {
+      color: var(--watt-color-state-info);
+    }
+  }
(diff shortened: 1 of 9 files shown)


libs/ui-watt/src/lib/components/validation-message/watt-validation-message.component.scss
```diff
@@ -0,0 +1,40 @@
...
+@use '@energinet-datahub/watt/utils' as watt;
+
+:host {
+  align-items: center;
+  border-radius: 4px;
+  display: flex;
+  @include watt.space-inset-squish-s;
+
+  &.watt-validation-message--info {
+    background-color: var(--watt-color-state-info-light);
```
frontend-specialisten, line 26: Question: wouldn't the correct thing todo here not setting the color to the "contrast / text color" ex. `--watt-color-state-info-contrast`?

frontend-specialisten (approved):

libs/ui-watt/src/lib/components/validation-message/watt-validation-message.component.scss
dzhavat: Good point. I forgot about them. Let's wait until UX review the colors in the design system and make the necessary adjustments, if any, afterwards.

PR #345, opened 2022-02-04
Enrich charge links with actual actor data

BjarkeMeier:
## Description

DRAFT STATUS BECAUSE:
- Uses the alpha version of the charge link client v2
- Is missing enrichment of actor data

Use the new charge link client v2, which doesn't provide actor data as it is not the responsibility of the charges domain. Instead, the actor data is fetched from the actor register and used to enrich the charge link data for the front-end.

## References

- #0000


BjarkeMeier:
@HenrikSommer , @prtandrup , I'm closing this PR. I assume you're not interested in it.

prtandrup:
> @HenrikSommer , @prtandrup , I'm closing this PR. I assume you're not interested in it.

Thanks @BjarkeMeier, that's fine. Architect has given go-ahead with serving the data from Charges, so we'll be doing that, when story is prioritized.

Issue #152, opened 2021-11-10
Spacing issues on Button suffix in Text Field component  


stephaniegitha:
Focus border of Button suffix (fx clear button) in Text Field component (Size: Default) overflows   

![Screenshot 2021-11-10 at 09.45.26.png](https://images.zenhubusercontent.com/60db08a8a79d712269c585cb/b4f54b06-fe01-4705-bf80-4339b96b94ba)

Too much margin to the right of button suffix in Text Field component

![Screenshot 2021-11-10 at 09.45.37.png](https://images.zenhubusercontent.com/60db08a8a79d712269c585cb/ba5168ab-879d-4b23-8625-43d2152fe189)

PR #376, opened 2022-02-24
style(form-field): fix height and spacing of preffix / suffix icon bu…

frontend-specialisten:
## Description
fix height and spacing of preffix / suffix icon buttons in form-fields

## References

- #152

### DIFF
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -206,12 +206,6 @@ jobs:
         run: yarn install --frozen-lockfile --non-interactive
       - name: Install Playwright
         run: npx playwright install --with-deps
-      - name: 'DH: Start API'
-        if: ${{ env.is-main-branch == 'true' || needs.dh-affected.outputs.is-affected == 'true' }}
-        run: |
-          yarn nx run api-dh:ci-configuration
-          yarn nx serve api-dh:serve:ci
-
       - name: 'PR: End-to-end test affected products'
         if: ${{ env.is-pull-request == 'true' }}
         run: yarn affected:e2e --configuration=ci --base=origin/$BASE_BRANCH
@@ -378,12 +372,6 @@ jobs:
             ${{ runner.os }}
       - name: Install package dependencies
         run: yarn install --frozen-lockfile --non-interactive
-
-      - name: 'DH: Start API'
-        if: ${{ env.is-main-branch == 'true' || needs.dh-affected.outputs.is-affected == 'true' }}
-        run: |
-          yarn nx run api-dh:ci-configuration
-          yarn nx serve api-dh:serve:ci
 
       - name: 'PR: Test affected projects'
         if: ${{ env.is-pull-request == 'true' }}
--- a/.github/workflows/frontend-ci.yml
+++ b/.github/workflows/frontend-ci.yml
@@ -206,6 +206,12 @@ jobs:
         run: yarn install --frozen-lockfile --non-interactive
       - name: Install Playwright
         run: npx playwright install --with-deps
+      - name: 'DH: Start API'
+        if: ${{ env.is-main-branch == 'true' || needs.dh-affected.outputs.is-affected == 'true' }}
+        run: |
+          yarn nx run api-dh:ci-configuration
+          yarn nx serve api-dh:serve:ci
+
       - name: 'PR: End-to-end test affected products'
         if: ${{ env.is-pull-request == 'true' }}
         run: yarn affected:e2e --configuration=ci --base=origin/$BASE_BRANCH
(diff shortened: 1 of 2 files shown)


libs/ui-watt/src/lib/components/form-field/form-field.component.scss
```diff
@@ -78,12 +85,27 @@ watt-form-field {
...
+    .watt-icon-button {
+      height: 40px;
+    }
+  }
+
   [wattPrefix] {
     margin-right: 8px;
+
+    .watt-icon-button {
+      margin-left: -12px;
```
dzhavat, line 99: issue: this is applying height to icons inside a `watt-icon-button` component. Often the prefix icon is a `watt-icon` component. Consider that the icon might be inside a `watt-icon-button` component. 

question: Will there be a case where the prefix icon is a `watt-icon-button`?

libs/ui-watt/src/lib/components/form-field/form-field.component.scss
```diff
@@ -78,12 +85,27 @@ watt-form-field {
...
+    }
   }
 
   [wattSuffix] {
     margin-left: 8px;
+
+    .watt-icon-button {
+      margin-right: -12px;
```
dzhavat, line 107: question: Wouldn't it be better to move the whole component 12px to the right instead of just the button? Right now the button inside the component is moved to the right and when inspected, the component looks like this:

image

libs/ui-watt/src/lib/components/form-field/form-field.component.scss
```diff
@@ -31,6 +31,13 @@ watt-form-field {
     .mat-form-field-flex {
       height: 56px;
     }
+
+    [wattPrefix],
+    [wattSuffix] {
+      .watt-icon-button {
+        height: 52px;
```
dzhavat, line 38: issue: this is applying height to icons inside a `watt-icon-button` component. Often the prefix icon is a `watt-icon` component (e.g. inside `dh-metering-point-search-form.component.html`). This style won't be applied in that case. Will there be a case where the prefix icon is a `watt-icon-button`?

libs/ui-watt/src/lib/components/form-field/form-field.component.scss
```diff
@@ -78,12 +85,27 @@ watt-form-field {
     display: none;
   }
 
+  [wattPrefix],
+  [wattSuffix] {
+    .watt-icon-button {
+      height: 40px;
```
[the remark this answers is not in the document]
frontend-specialisten: The search form is also using a `watt-icon-button` when the icon is clickable (the reset). If it's just an indicator like the search icon, it won't have this issue as it's not clickable and won't get any outline (it's not even focusable). This is why I'm only targeting the `watt-icon-button` :)

libs/ui-watt/src/lib/components/form-field/form-field.component.scss
frontend-specialisten: If we're targeting the whole `suffix`, icons would also be moved because it doesn't have any padding.

dzhavat: What about targeting `watt-icon-button` elements with `wattSuffix` e.g. `watt-icon-button[wattSuffix]`?

frontend-specialisten: Unfortunately `watt-icon-button[wattSuffix]` won't work as this won't change the height of the button itself.

dzhavat: This was more related to the `margin-right`. Instead of moving the inner button, we can move the whole `watt-icon-button` element. Wouldn't that work?

frontend-specialisten: Selector regarding margin-right and left has now been changed

.github/workflows/frontend-ci.yml
```diff
@@ -379,12 +373,6 @@ jobs:
       - name: Install package dependencies
         run: yarn install --frozen-lockfile --non-interactive
 
-      - name: 'DH: Start API'
-        if: ${{ env.is-main-branch == 'true' || needs.dh-affected.outputs.is-affected == 'true' }}
-        run: |
-          yarn nx run api-dh:ci-configuration
-          yarn nx serve api-dh:serve:ci
-
```
dzhavat, line 387: question: Are these changes necessary in this PR?

frontend-specialisten: No that was a mistake, changes has now been reverted

dzhavat (approved):

PR #388, opened 2022-03-04
Updated workflow to check is-affected

MartinSchmidt:
## Description
Only run version check if is-affected is true.


LayZeeDK (approved):
Chef's kiss

Issue #338, opened 2022-02-03
Add Dropdown component to Design System


mknic:
**Description**
Add dropdown component to Design System

**AC**
- [ ] Align with UX on MVP for a dropdown component
- [ ] Dropdown component is added to design system
- [ ] Showcase and documented in storybook

**Notes**
- [Component in Figma](https://www.figma.com/file/fwEDQkgEcpwfUojXl0EIC5/Watt-Design-System?node-id=209%3A3)

mknic:
Remaining estimate is 3.

PR #422, opened 2022-03-18
feat(watt): add dropdown to Watt Design System

dzhavat:
### Watt Design System
- Add `watt-dropdown` component
- Styling adjustments to `watt-form-field` component
- Showcase and documentation in Storybook

Single selection

image

Multi-selection

image

## References

- #338


libs/ui-watt/src/lib/components/dropdown/+storybook/storybook-dropdown-template-driven-forms.stories.ts
```diff
@@ -0,0 +1,178 @@
...
+ 
+ <watt-dropdown [(ngModel)]="singleSelectionModel" [options]="options"></watt-dropdown>
+ 
+ 5. Wrap the dropdown component in a "watt-form-field"
+
+ <watt-form-field>
+  <watt-dropdown [(ngModel)]="singleSelectionModel" [options]="options"></watt-dropdown>
+ </watt-form-field>`;
+
+export const singleSelect: Story<WattDropdownComponent> = (
```
frontend-specialisten, line 78: Question: Should we have some showcases for `large` form-fields as well or is this out of scope for now?

libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.ts
```diff
@@ -0,0 +1,324 @@
...
+@Component({
+  selector: 'watt-dropdown',
+  templateUrl: './watt-dropdown.component.html',
+  styleUrls: ['./watt-dropdown.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class WattDropdownComponent
+  implements ControlValueAccessor, OnInit, OnDestroy
+{
+  /** Subject that emits when the component has been destroyed. */
```
frontend-specialisten, line 56: Suggestion: Let's figure out an convention how we want our comments formatted, so they can be more consistent.

frontend-specialisten (approved):
Praise: Looking good! Just have a single question and a nitpick, I'm approving this anyways since these changes won't break anything for the users, so I don't see any reasons for holding this back so ex. Titans can start using it. My comments can easily be implemented in other pull-request(s) but ofc. it's up to you.

libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.ts
dzhavat: Good point. I'll then suggest `//` for comments that explain a property, code chunk, etc. and TypeDoc kind of comments for API artifacts (property on a component, interface, type, etc.). What do you think?

Will also delete the comment above. We use this pattern in many places and we should be familiar with it by now.

libs/ui-watt/src/lib/components/dropdown/+storybook/storybook-dropdown-template-driven-forms.stories.ts
dzhavat: I think we should include it. It's also a good opportunity to test how the dropdown will work in a `large` form-field. Looks like Titans are only using `normal` form-fields for now. So we can make a follow-up PR with a showcase for `large` form-fields.

frontend-specialisten: Totally fine by me, do you create an issue for this follow-up for the backlog? :)

frontend-specialisten (approved):

frontend-specialisten (approved):

libs/ui-watt/src/lib/components/dropdown/+storybook/storybook-dropdown-template-driven-forms.stories.ts
dzhavat: Just created the issue in the backlog.

PR #428, opened 2022-03-22
Added Organizations from Market Participant

defectiveAi:


defectiveAi:
@frontend-specialisten PR is ready for first review. Feedback is more than welcome, this is the first time our team has made UI. Also, SQ fails because of code-duplication in autogenerated code, not sure what we can do about that?

libs/dh/market-participant/data-access-api/src/lib/dh-market-participant-data-access-api.store.ts
```diff
@@ -0,0 +1,69 @@
...
+  readonly setActors = (organizations: OrganizationWithActor[]) => {
+    this.patchState({ organizations, isLoading: false });
+  };
+
+  readonly mapActors = (organizations: OrganizationDto[]) => {
+    return organizations.reduce((running, x) => {
+      return (
+        (x.actors.length > 0 &&
+          running.concat(
+            x.actors.map((a) => ({ organization: x, actor: a }))
```
frontend-specialisten, line 63:
```suggestion
            x.actors.map((actor) => ({ organization: x, actor }))
```

libs/dh/market-participant/feature-organization/.browserslistrc
```diff
@@ -0,0 +1,16 @@
...
+# For additional information regarding the format and rule options, please see:
+# https://github.com/browserslist/browserslist#queries
+
+# For the full list of supported browsers by the Angular framework, please see:
+# https://angular.io/guide/browser-support
+
+# You can see what browsers were selected by your queries by running:
+#   npx browserslist
+
+last 1 Chrome version
```
frontend-specialisten, line 11: Suggestion: remove this file, as it shouldn't be necessary

libs/dh/market-participant/shell/.browserslistrc
```diff
@@ -0,0 +1,16 @@
...
+# For additional information regarding the format and rule options, please see:
+# https://github.com/browserslist/browserslist#queries
+
+# For the full list of supported browsers by the Angular framework, please see:
+# https://angular.io/guide/browser-support
+
+# You can see what browsers were selected by your queries by running:
+#   npx browserslist
+
+last 1 Chrome version
```
frontend-specialisten, line 11: Suggestion: Remove this file as this shouldn't be necessary

libs/dh/market-participant/data-access-api/.browserslistrc
```diff
@@ -0,0 +1,16 @@
...
+# For additional information regarding the format and rule options, please see:
+# https://github.com/browserslist/browserslist#queries
+
+# For the full list of supported browsers by the Angular framework, please see:
+# https://angular.io/guide/browser-support
+
+# You can see what browsers were selected by your queries by running:
+#   npx browserslist
+
+last 1 Chrome version
```
frontend-specialisten, line 11: Suggestion: Remove this file as it shouldn't be necessary

libs/dh/market-participant/feature-organization/src/lib/dh-market-participant-organization.component.html
```diff
@@ -0,0 +1,135 @@
...
+    </ng-container>
+
+    <!-- Actor GLN -->
+    <ng-container matColumnDef="{{ this.columnIds[1] }}">
+      <th mat-header-cell *matHeaderCellDef>
+        {{ transloco('columns.gln') }}
+      </th>
+      <td mat-cell *matCellDef="let element">
+        <div *ngIf="element.actor !== undefined">
+          {{ element.actor.gln.value }}
```
frontend-specialisten, line 44: Consider: Is `element.actor.gln` always set? Otherwise be careful about deeply nested objects. You can use the safe navigation operator to be avoid issues like this `element?.actor?.gln?.value`. Also if some of the values can be null or undefined consider if fallback values should be shown. (this affects all the values in the template)

frontend-specialisten (changes requested):

libs/dh/market-participant/shell/.browserslistrc
defectiveAi: Removed

libs/dh/market-participant/data-access-api/.browserslistrc
defectiveAi: Removed

libs/dh/market-participant/feature-organization/src/lib/dh-market-participant-organization.component.html
defectiveAi: GLN is always set, while actor can be missing. I've added fallbacks in places.

frontend-specialisten (approved):

PR #433, opened 2022-03-25
chore(watt): fix lint error

dzhavat:
### Watt Design System
- Fixes a lint error

## References

- N/A

### DIFF
--- a/libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
+++ b/libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
@@ -21,7 +21,7 @@ import { RouterModule } from '@angular/router';
 
 import { WattNavListModule } from './watt-nav-list.component';
 
-const energientDkUrl = 'https://energinet.dk';
+const energinetDkUrl = 'https://energinet.dk';
 
 describe(WattNavListModule.name, () => {
   it('exports shared Watt Design System nav list', async () => {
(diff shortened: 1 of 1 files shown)


libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
```diff
@@ -154,11 +156,11 @@ describe(WattNavListModule.name, () => {
       const link = await screen.findByRole('link', {
         name: /energinet/i,
       });
-      expect(link).toHaveAttribute('href', 'http://energinet.dk');
+      expect(link).toHaveAttribute('href', energinetDkUrl);
```
LayZeeDK, line 159: issue: This link uses the `http://` protocol so it can't use the same variable as the instances above where the `https://` protocol (notice the `s`) is used.

LayZeeDK (changes requested):

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
dzhavat: Good catch. Could it be because the changes are only in a spec file, Nx didn't detect the library as effected and didn't run the tests? https://github.com/Energinet-DataHub/greenforce-frontend/runs/5716634156?check_suite_focus=true#step:9:24
Should would've been catched otherwise.

LayZeeDK (approved):
Thank you for catching this.

PR #453, opened 2022-04-07
updated message archive ui

xnetsrak:
## Description

updated message archive ui
## References

- #0000


libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.html
```diff
@@ -16,35 +16,26 @@
...
+          transloco("documenttype")
         }}</watt-label>
 
-        <select
-          #searchInputDocumentType
-          wattInput
-          name="documentType"
+        <watt-dropdown
           [(ngModel)]="this.searchCriteria.rsmName"
+          [formControl]="rsmFormControl"
```
dzhavat, line 29: issue: it's not recommended to use `ngModel` and `formControl` on the same field. The `watt-dropdown` component works with either, so you can remove one of them. By looking at the file as a whole, you're mostly using the template-driven approach, so you can delete the `formControl`. Both here and the `rsmFormControl` property.

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.html
```diff
@@ -16,35 +16,26 @@
...
+          [options]="rsmFormFieldOptions"
+          [placeholder]="transloco('ChooseDocumentType')"
+          [noOptionsFoundLabel]="'-'"
```
dzhavat, line 32: issue: the `noOptionsFoundLabel` only works with multi-select. Setting it on single selection dropdown will do nothing.

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.html
```diff
@@ -55,86 +46,62 @@ <h1>{{ transloco('title') }}</h1>
...
       </watt-form-field>
 
       <watt-form-field size="normal">
         <watt-label class="watt-text-m">{{
-          transloco('businessReason')
+          transloco("businessReason")
         }}</watt-label>
+        <watt-dropdown
+          [(ngModel)]="this.searchCriteria.processType"
+          [formControl]="processTypeFormControl"
```
dzhavat, line 83: issue: it's not recommended to use `ngModel` and `formControl` on the same field. The `watt-dropdown` component works with either, so you can remove one of them. By looking at the file as a whole, you're mostly using the template-driven approach, so you can delete the `formControl`. Both here and the `processTypeFormControl` property.

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.html
```diff
@@ -55,86 +46,62 @@ <h1>{{ transloco('title') }}</h1>
...
+          [options]="processTypeFormFieldOptions"
+          [placeholder]="transloco('ChooseBusinessReason')"
+          [noOptionsFoundLabel]="'-'"
```
dzhavat, line 86: issue: the `noOptionsFoundLabel` only works with multi-select. Setting it on single selection dropdown will do nothing.

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.html
```diff
@@ -55,86 +46,62 @@ <h1>{{ transloco('title') }}</h1>
...
+        >
+        </watt-dropdown>
+      </watt-form-field>
 
-        <select
-          #searchInputBusinessReasonCode
+      <!-- Reference id -->
+      <watt-form-field size="normal">
```
dzhavat, line 92: suggestion: the `watt-checkbox` component is not designed to be used inside `watt-form-field`. You can remove the `watt-form-field` wrapper.

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.html
```diff
@@ -145,8 +112,9 @@ <h1>{{ transloco('title') }}</h1>
         />
       </watt-form-field>
 
+      <!-- Date to -->
       <watt-form-field size="normal">
```
dzhavat, line 116: suggestion: the default size of `watt-form-field` is "normal" so you don't have to specify it.

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.html
```diff
@@ -55,86 +46,62 @@ <h1>{{ transloco('title') }}</h1>
...
+          [(ngModel)]="this.searchCriteria.receiverId"
           wattInput
           type="text"
-          name="messageId"
+          name="receiverId"
         />
       </watt-form-field>
 
       <watt-form-field size="normal">
         <watt-label class="watt-text-m">{{
```
dzhavat, line 64: suggestion: when you use `watt-label` inside `watt-form-field` it's better to not apply additional `watt-text-*` classes. Can you remove them from here and other similar places in the file? The text in the UI will look like this: 

image

dzhavat (changes requested):

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.ts
```diff
@@ -117,11 +119,26 @@ export class DhMessageArchiveLogSearchComponent implements OnDestroy {
...
-      this.initDateTo().toISOString().split('.')[0] + 'Z';
+  private buildRsmOptions() {
+    const entries = Object.entries(DocumentTypes);
+    entries.forEach((entry) => {
+      this.rsmFormFieldOptions.push({
+        value: entry[0],
+        displayValue: entry[1] + ' - ' + entry[0],
+      });
+    });
+  }
```
dzhavat, line 132: suggestion: instead of modifying the `rsmFormFieldOptions` property by pushing array items to it, consider building the array in the method and returning it. This way you keep the function more "pure". So it could be something like:

```suggestion
  private buildRsmOptions() {
    return Object.entries(DocumentTypes).map((entry) => {
      return {
        value: entry[0],
        displayValue: entry[1] + ' - ' + entry[0],
      };
    });
  }
```

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.ts
```diff
@@ -117,11 +119,26 @@ export class DhMessageArchiveLogSearchComponent implements OnDestroy {
...
+
+  private buildProcessTypesOptions() {
+    const entries = Object.entries(ProcessTypes);
+    entries.forEach((entry) => {
+      this.processTypeFormFieldOptions.push({
+        value: entry[0],
+        displayValue: entry[0] + ' - ' + entry[1],
+      });
+    });
```
dzhavat, line 141: suggestion: instead of modifying the `processTypeFormFieldOptions` property by pushing array items to it, consider building the array in the method and returning it. This way you keep the function "pure" and with less side-effects. So it could be something like:

```suggestion
  private buildProcessTypesOptions() {
    return Object.entries(ProcessTypes).map((entry) => {
      return {
        value: entry[0],
        displayValue: entry[0] + ' - ' + entry[1],
      };
    });
```

dzhavat (dismissed):

dzhavat (approved):

PR #454, opened 2022-04-08
EO: Add content to Origin of Energy page

MadsThy:
## Description
image

Added static content to the page.

## Feature environment
Available at https://eo-u-431addcontent.westeurope.cloudapp.azure.com/origin-of-energy

## References

- https://github.com/Energinet-DataHub/energy-origin-issues/issues/431


LayZeeDK:
Try something like [this](https://stackoverflow.com/questions/36992922/chart-js-v2-how-to-make-tooltips-always-appear-on-pie-chart) to put tooltips on top of the chart slices.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  template: `<canvas
+    baseChart
+    [data]="pieChartData"
+    [type]="pieChartType"
+    [options]="pieChartOptions"
+    [plugins]="pieChartPlugins"
+  >
+  </canvas>`,
+  styles: [``],
+})
```
Klemmensen, line 33: fix/question:

Since this is state of the component, we could consider moving this part into a ComponentStore. It might blow up the task, so talk with Lars on how we should do this for now.

See example of state management inside, i.eg. 

- eo-privacy-policy.store.ts
- eo-media.presenter.ts

Klemmensen (changes requested):

apps/eo/app-eo/project.json
```diff
@@ -56,22 +56,22 @@
             {
               "type": "allScript",
               "maximumWarning": "500kb",
-              "maximumError": "1mb"
```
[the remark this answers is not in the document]
MadsThy: This is because of recent changes to the project, where new packages have added  to the overall size of the bundle :) It now excees the 1mb that was alotted, and is at 1.01. So in order for the CI to allow me to build, i have to increase the maximum allowed size.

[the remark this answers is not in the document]
LayZeeDK: ...or stop increasing the bundle size. Increasing the size limit is the road to Hell for our users. What's the cause? Is something not lazy loaded?

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
```diff
@@ -14,45 +14,160 @@
...
+        width: 360px; /* Magic UX number */
+        gap: var(--watt-space-l);
+        display: flex;
+        flex-direction: column;
+      }
+
+      .global-goals-box {
+        border: 1px solid #f9d557; /* Color not yet added to Watt */
+        max-height: 100px; /* Magic UX number */
+
```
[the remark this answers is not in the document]
MadsThy: I agree :) But the issue here is that the text that UX wants in here, is actually 104px tall. If i take that and add the 1px on top and buttom, the height would be 106 in total.

I will have a talk with UX about it, but otherwise i need to "squish" the box size for this paragraph in order for it to fit :)

libs/eo/core/shell/jest.config.js
```diff
@@ -25,6 +25,13 @@ module.exports = {
     },
   },
   coverageDirectory: '../../../../coverage/libs/eo/core/shell',
+  /**
+   * https://stackoverflow.com/questions/42260218/jest-setup-syntaxerror-unexpected-token-export
+   * NG2-charts has an error with lodash-es, Jest tests fail if this is not added.
+   */
+  moduleNameMapper: {
+    'lodash-es': 'lodash',
+  },
```
LayZeeDK, line 34: suggestion: Try moving this to the root `jest.config.js` to verify whether project-specific `jest.config.js` configurations extend from it. Otherwise, we probably have to add it to all libraries that depend on a library using `ng2-charts`.

libs/eo/core/shell/jest.config.js
```diff
@@ -25,6 +25,13 @@ module.exports = {
     },
   },
   coverageDirectory: '../../../../coverage/libs/eo/core/shell',
```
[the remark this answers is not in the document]
MadsThy: I agree. Should we maybe make a tech-debt story for this so we keep it for the future and move forward with this story? :)

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
```diff
@@ -14,45 +14,160 @@
...
+          <img
+            eoMediaImage
+            [eoMediaImageMaxWidthPixels]="100"
+            src="/assets/images/origin-of-energy/globalgoal-7.2.svg"
+            alt="Global goal 7.2"
+          />
+        </eo-media>
+        <mat-card class="tip-card">
+          <div class="tip-card-header">
+            <img
```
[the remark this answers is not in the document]
MadsThy: Fixed :)

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
```diff
@@ -14,45 +14,183 @@
...
+
+      .chart-card {
+        max-width: 584px; /* Magic UX number */
+        width: 100%;
+      }
+
+      .chart-box {
+        margin-top: var(--watt-space-m);
+        margin-right: calc(var(--watt-space-xl) - var(--watt-space-s));
+        margin-left: calc(var(--watt-space-xl) - var(--watt-space-s));
```
LayZeeDK, line 44: issue: Unless the design dimensions are too *magic* and the UX designer insists on magic spacing dimensions, we should use the Watt spacing primitives for spacing, that is Inset, Squished Inset, Stretched Inset, Stack, and Inline.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
```diff
@@ -14,45 +14,160 @@
...
+            You can increase your share of renewable energy by shifting your
+            consumption to periods with more renewable energy in the grid.
+          </p>
+        </mat-card>
+      </div>
+    </div>
+    <div class="description-row">
+      <mat-card class="description-card watt-space-inline-l">
+        <p><strong>Renewable energy</strong></p>
+        <img
```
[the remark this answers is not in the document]
MadsThy: Fixed :)

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
```diff
@@ -14,45 +14,183 @@
...
+
+      .chart-tips {
+        width: 360px; /* Magic UX number */
+        gap: var(--watt-space-l);
+        display: flex;
+        flex-direction: column;
+      }
+
+      .global-goals-box {
+        border: 1px solid #f9d557; /* Color not yet added to Watt */
```
LayZeeDK, line 55: suggestion: Refer to Watt's *warning* color both here and in the SVG until colors in the Watt implementation are aligned with the specification in Figma. Otherwise, we have to come back here and fix it. We could also add a note to the enabler we have on aligning Energy Origin colors after the Watt implementation has been aligned with Storybook.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
```diff
@@ -14,45 +14,160 @@
...
 })
-export class EoOriginOfEnergyShellComponent {
-  iconSize = WattIconSize;
-}
+export class EoOriginOfEnergyShellComponent {}
 
 @NgModule({
   declarations: [EoOriginOfEnergyShellComponent],
   exports: [EoOriginOfEnergyShellComponent],
-  imports: [WattIconModule, EoInlineMessageScam],
```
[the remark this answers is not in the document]
MadsThy: You caught it correctly :) The CommonModule import was a leftover from when i hadn't put the pie-chart into it's own component. Removed it now.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
```diff
@@ -15,44 +15,189 @@
...
+          The hourly declaration describes the origin of the energy you have
+          consumed within a given period as well as the corresponding emissions.
+        </p>
+        <p>
+          The declaration is calculated as a weighted average based on your
+          hourly electricity consumption and the corresponding hourly residual
+          mix in your bidding zone.
+        </p>
+      </mat-card>
+    </div>`,
```
LayZeeDK: suggestion: Consider splitting this component into separate components. It's grown big and a shell component is not supposed to have much content or styling, mostly just organizing the order and layout of other components.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { render, screen } from '@testing-library/angular';
+import { EoPieChartComponent, EoPieChartScam } from './eo-pie-chart.component';
+
+describe(EoPieChartComponent.name, () => {
+  beforeEach(async () => {
+    await render('<eo-pie-chart role="piechart" ></eo-pie-chart>', {
```
LayZeeDK, line 22: suggestion: Remove extra space
```suggestion
    await render('<eo-pie-chart role="piechart"></eo-pie-chart>', {
```

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+      imports: [EoPieChartScam],
+    });
+
+    hostElement = screen.getByRole('piechart');
```
LayZeeDK, line 26: issue: `piechart` is not a valid ARIA role

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
MadsThy: My own view on this, would be that i would rather manage this components state at component-level :) I was thinking we could reuse the component around the app by providing inputs for the colors, labels and so on.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  selector: 'eo-pie-chart',
+  template: `<canvas
+    baseChart
+    [data]="pieChartData"
+    [type]="pieChartType"
+    [options]="pieChartOptions"
+    [plugins]="pieChartPlugins"
+  >
+  </canvas>`,
+  styles: [``],
```
LayZeeDK, line 32: issue: At the very least, a component must have a `block` box model so that it contains it content (`inline` is the default box model for custom element names).
```suggestion
  styles: [
    `
      :host {
        display: block;
      }
    `
  ],
```

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
...
+        rotation: 180,
+        data: [81, 19],
+        backgroundColor: ['#7FB069', '#616161'],
+        hoverBackgroundColor: ['#7FB069', '#616161'],
+        borderWidth: 0,
+      },
+    ],
+  };
+  public pieChartType: ChartType = 'pie';
+  public pieChartPlugins = [DatalabelsPlugin];
```
LayZeeDK, line 74: suggestion: I suggest not using the default `public` access modifier except for shorthand property initializers in a constructor. We could enforce this using the [explicit-member-accessibility](https://typescript-eslint.io/rules/explicit-member-accessibility) lint rule but this should be done in a separate PR and aligned with other teams as it also affects  Watt and DataHub.
```suggestion
  pieChartOptions: ChartConfiguration['options'] = {
    responsive: true,
    animation: false,
    plugins: {
      tooltip: { enabled: false },
      legend: {
        display: false,
      },
      datalabels: {
        textAlign: 'center',
        color: ['black', 'white'],
        font: {
          family: 'Open Sans',
          weight: 'bold',
          size: 16,
        },
        formatter: (value, ctx) => {
          if (ctx.chart.data.labels) {
            return `${value}%\n${ctx.chart.data.labels[ctx.dataIndex]}`;
          }
          return '';
        },
      },
    },
  };

  pieChartData: ChartData<'pie', number[], string | string[]> = {
    labels: ['Renewable', 'Other'],
    datasets: [
      {
        rotation: 180,
        data: [81, 19],
        backgroundColor: ['#7FB069', '#616161'],
        hoverBackgroundColor: ['#7FB069', '#616161'],
        borderWidth: 0,
      },
    ],
  };
  pieChartType: ChartType = 'pie';
  pieChartPlugins = [DatalabelsPlugin];
```

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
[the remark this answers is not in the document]
MadsThy: UX has said i should allow it to grow, so i will fix it to be 106px tall :)

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
...
+          family: 'Open Sans',
+          weight: 'bold',
+          size: 16,
+        },
+        formatter: (value, ctx) => {
+          if (ctx.chart.data.labels) {
+            return `${value}%\n${ctx.chart.data.labels[ctx.dataIndex]}`;
+          }
+          return '';
+        },
```
LayZeeDK, line 56: suggestion: Add a few unit tests for this logic.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
...
+        display: false,
+      },
+      datalabels: {
+        textAlign: 'center',
+        color: ['black', 'white'],
+        font: {
+          family: 'Open Sans',
+          weight: 'bold',
+          size: 16,
+        },
```
LayZeeDK, line 50: suggestion: Store these values in variables containing `watt` in their name so that we at the very least know why they are used. Ideally, we would compute the values from a rendered element relying on global styles.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
+/**
```
LayZeeDK, line 1: issue: A chart is usually a molecule, not an atom.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+    },
+  };
+
+  public pieChartData: ChartData<'pie', number[], string | string[]> = {
+    labels: ['Renewable', 'Other'],
+    datasets: [
+      {
+        rotation: 180,
+        data: [81, 19],
```
LayZeeDK, line 66: issue: If we want to have this component in our shared Atomic Design collection, it should be reusable so we would at the very least need to accept labels and data points as input properties. Alternatively, keep this component in the Origin of Energy domain for now. Avoid Hasty Abstractions (AHA).

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
...
+        backgroundColor: ['#7FB069', '#616161'],
+        hoverBackgroundColor: ['#7FB069', '#616161'],
```
LayZeeDK, line 68: question: Are these Watt colors? If so, we should compute them from Watt's CSS Custom Properties.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
```diff
@@ -14,45 +14,160 @@
...
+      <div class="chart-tips">
+        <eo-media [eoMediaMaxWidthPixels]="360" class="global-goals-box">
+          <p class="watt-space-inset-m">
+            <strong>Global goals 7.2</strong><br />
+            increase the share of renewable energy globally
+          </p>
+          <img
+            eoMediaImage
+            [eoMediaImageMaxWidthPixels]="100"
+            src="/assets/images/origin-of-energy/globalgoal-7.2.svg"
```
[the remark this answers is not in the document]
MadsThy: I think the alt text would be a text states what is on the picture :) I would possibly even add "icon" to it so it says "Global goal 7.2 icon", making it better for reading out loud by page readers.

LayZeeDK (changes requested):

apps/eo/app-eo/project.json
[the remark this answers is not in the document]
MadsThy: I observed this after i pushed an update with ng2-charts, chart.js, and chartjs-plugin-datalabels :) I had just merged in an update from main that included "inputmask" package as well, and that caused my build to fail on this.

I am unsure how to procede from here. Any advice @LayZeeDK  or @Klemmensen ? I don't know how the app should behave after adding 4 new packages in total.

libs/eo/core/shell/jest.config.js
MadsThy: @LayZeeDK, @Klemmensen and i actually had a talk about this, and it might be better for us to move to ngx-chart instead in the future. Lots of people have had problems in their projects when using ng2-charts, because it has a dependency with lodash-es. NGX-chart does not.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
MadsThy: Sure. I can do that :)

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
MadsThy: I will put it in the Origin of Energy domain for now then, seeing as we are still trying to find a good fit for charting :)

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
MadsThy: Only 616161 is in Watt right now :) Should i just make that one compute then, and keep the other static?

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
LayZeeDK: Call the function `component.pieChartOptions.plugins.datalabels.formatter` in a unit test.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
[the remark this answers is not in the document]
Klemmensen: Yes agree, that is what it is for - Also the reason I'm asking, searchability was a loooong time ago also possible by branding image ALT tags with i.eg. the domain name/company name or what ever is relevant :-)

I am not sure if the value of that still counts?

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
LayZeeDK: It's not necessary for now but the need will increase if we combine data sources or add more complex business logic. Let's deal with it if something like that happens. I have real world examples of splitting chart concerns into presenters.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
MadsThy: I have remove this, and replaced it with watt-space-inset-squish-l after agreeing with UX about it :)

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+      datalabels: {
+        textAlign: 'center',
+        color: ['black', 'white'],
+        font: {
+          family: 'Open Sans',
+          weight: 'bold',
+          size: 16,
+        },
+        formatter: (value, ctx) => {
```
LayZeeDK, line 51: issue: TypeScript is not able to infer proper types so we must add type annotations.
image

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
Klemmensen: Sure, I think as long as state management is somewhat consistent across the app, we are good - But Component Store is actually meant to be on component level.

apps/eo/app-eo/project.json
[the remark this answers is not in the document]
LayZeeDK: We must analyze the output bundles using source-map-analyzer or [@ngx-builders/analyze](https://github.com/ngx-builders/source-map-analyzer).

We can also evaluate which budgets are going to grow with the app or how we can prevent this: [Configuring size budgets](https://angular.io/guide/build#configuring-size-budgets).

libs/eo/core/shell/jest.config.js
Klemmensen: If we can test "as is", with ngx charts, it is also a benefit. Some exploration could be done here :-D

LayZeeDK: Well, we should also use `lodash-es` over `lodash` to keep a minimal bundle size impact.

Another major difference is that `@swimlane/ngx-charts` renders D3 SVG charts whereas `ng2-charts` renders a canvas element.

In favor of SVG:
- Higher testability as we can interact with SVG elements in our integration and end-to-end tests
- More style customizations as we can apply CSS to SVGs
- Possibly OK for accessibility given enough control of the SVG structure or if this is taken into the account by the chart library

In favor of canvas:
- Canvas performance cannot be beat using SVG. With many charts on one page or many data points in the same chart, performance drops or grinds to a halt when using SVG-based charts for the same reason as having tens of thousands of HTML elements on the page does.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
LayZeeDK: Something like this:
```ts
const cssValue = getComputedStyle(element).getPropertyValue('--my-property')
```

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
LayZeeDK: I didn't install the added libraries yet. Let me confirm this issue.

LayZeeDK: This is still an issue after installing the chart library packages.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
dzhavat: I'm sorry for commenting without being invited but just noticed this comment in my inbox, and I believe I can share something useful.

In Watt we have the `WattColorHelperService` that can help you access Watt colors in TypeScript files. So you don't have to compute them yourself.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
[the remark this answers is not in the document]
MadsThy: Fixed :)

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
MadsThy: Will skip this for now then, but bear it in mind for when the chart is gonna become more generic :) 

Thanks both!

libs/eo/core/shell/jest.config.js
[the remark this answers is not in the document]
MadsThy: Resolving this comment, as we will take it into consideration in the future :)

libs/eo/core/shell/jest.config.js
MadsThy: Moved the moduleNameMapper to project level, and removed it from specific areas inside the project.

We will look at NG2 vs NGX in the future.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
MadsThy: Changed the variable for warning-yellow in the project SCSS variables.

Also fixed the usage of this variable in the inline-message component, because these two places were the only ones affected by this color change.

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy-shell.component.ts
MadsThy: Split up the component into smaller components

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
MadsThy: Moved the chart into the energy of origin lib, so it is no longer part of atomic-design

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
MadsThy: Created private variables for taking the root styles from memory that you mentioned here.

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-pie-chart/eo-pie-chart.component.ts
MadsThy: Added types from ChartJS datalabel package

Klemmensen (dismissed):

Klemmensen (dismissed):

LayZeeDK (dismissed):
🚀

libs/ui-watt/src/lib/foundations/_variables.scss
```diff
@@ -57,7 +57,7 @@
   --watt-color-state-danger-contrast: var(--watt-color-neutral-white);
   --watt-color-state-danger-light: #f7e2e0;
   --watt-color-state-danger-light-contrast: var(--watt-color-neutral-black);
-  --watt-color-state-warning: #f8ae3c;
+  --watt-color-state-warning: #f9d557;
```
LayZeeDK: issue: We can't change Watt without coordinating with GreenForce.

LayZeeDK (changes requested):

LayZeeDK (approved):

Issue #418, opened 2022-03-16
Time-picker input mask


dzhavat:
**AC**

- [x] Ensure correct format when typing in time inputs
- [x] Should support hours and minutes
- [x] Add showcase and documentation in Storybook

mknic:
Remaining: 3

PR #457, opened 2022-04-11
feat(watt): add time-range-input to Design System

dzhavat:
### DataHub
- add "Droid Sans Mono" font
- increase the size budget for the `initial` bundle to 1.3 Mb

### Watt Design System
- add `watt-time-input-range` component
- add showcase in Storybook

time-picker

## References

- #418


frontend-specialisten (dismissed):
Looks good, the only missing piece as I see it, is to re-use the logic and styles. When this is merged it can be done in the date-mask branch.

frontend-specialisten (approved):

PR #468, opened 2022-04-19
refactor(dh): simplify sending of route param to NgRx effect

dzhavat:
### DataHub
- refactor sending of route param is sent to NgRx Component Store effect

## References

- N/A


frontend-specialisten (approved):
Praise: Nice, simplyfications.

PR #480, opened 2022-04-21
added actor id to show condition

xnetsrak:
## Description

added actor id to show condition
## References

- #0000


libs/dh/message-archive/feature-log-search/src/lib/searchresult/dh-message-archive-log-search-result.component.html
```diff
@@ -36,14 +36,24 @@
     <!-- sender gln -->
     <ng-container matColumnDef="sender">
       <th mat-header-cell *matHeaderCellDef>{{ transloco("sender") }}</th>
-      <td mat-cell *matCellDef="let element">{{ element.senderGln }}</td>
+      <td mat-cell *matCellDef="let element">
+        <container-element [ngSwitch]="element.httpData">
+          <label *ngSwitchCase="'request'" type="info">{{
```
dzhavat, line 41: suggestion: The `label` element is usually used in relation to other elements like `input`, `textarea`, etc. I can see in the rest of the file that the `watt-badge` component was used for a similar case. Could it be that `watt-badge` should be used here too?

libs/dh/message-archive/feature-log-search/src/lib/searchresult/dh-message-archive-log-search-result.component.html
```diff
@@ -36,14 +36,24 @@
...
+            element.senderGln
+          }}</label>
+          <label *ngSwitchCase="'response'" type="success">{{
+            element.receiverGln
+          }}</label>
+          <label *ngSwitchDefault type="info">{{ "??" }}</label>
```
dzhavat, line 47: nitpick: the interpolation (`{{ .. }}`) is not necessary when we want to output a simple text string.

I.e. `<element>??</element>`

libs/dh/message-archive/feature-log-search/src/lib/searchresult/dh-message-archive-log-search-result.component.html
xnetsrak: no badge here it should only be text

dzhavat: Then unless you need specific styling, you can replace `label` with `ng-container` and remove the `type` property. `ng-container` is a generic element in Angular that will not render the element in the HTML DOM tree.

dzhavat (approved):

PR #482, opened 2022-04-21
Fix origin of energy stacking and FAQ page links

MadsThy:
## Description

Origin of Energy page was missing some padding between text, and the FAQ page was miscolored on the links to the items on the page.

image
image

### DIFF
--- a/libs/eo/faq/shell/src/lib/eo-faq-shell.component.ts
+++ b/libs/eo/faq/shell/src/lib/eo-faq-shell.component.ts
@@ -32,7 +32,7 @@ import { eoFaqRoutePath } from '@energinet-datahub/eo/faq/routing';
         text-decoration: none;
       }
 
-      h2 {
+      h3 {
         margin-top: var(--watt-space-xl);
         margin-bottom: var(--watt-space-s);
       }
@@ -65,7 +65,7 @@ import { eoFaqRoutePath } from '@energinet-datahub/eo/faq/routing';
       </a>
     </div>
 
-    <h2 id="what-is-energy-origin">What is Energy Origin?</h2>
+    <h3 id="what-is-energy-origin">What is Energy Origin?</h3>
     <p>
       The aim of the Energy Origin platform is to create transparency on the
       origin of energy for all energy consumers and to facilitate a market for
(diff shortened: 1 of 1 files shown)


libs/eo/faq/shell/src/lib/eo-faq-shell.component.ts
```diff
@@ -24,43 +24,48 @@ import { eoFaqRoutePath } from '@energinet-datahub/eo/faq/routing';
...
+        text-decoration: none;
+      }
+
+      h2 {
         margin-top: var(--watt-space-xl);
         margin-bottom: var(--watt-space-s);
       }
     `,
   ],
   template: `
```
Klemmensen, line 41: Remember that you will be formatting **a** tags from now and until the end of days, if you do not handle this in the build steps :-)

libs/eo/faq/shell/src/lib/eo-faq-shell.component.ts
```diff
@@ -24,43 +24,48 @@ import { eoFaqRoutePath } from '@energinet-datahub/eo/faq/routing';
...
     </div>
-    <div class="watt-space-stack-m">
-      <a href="${eoFaqRoutePath}#where-can-i-read-more"
-        >Where can I read more?</a
-      >
+    <div class="watt-space-stack-m faq-link">
+      <a href="${eoFaqRoutePath}#where-can-i-read-more">
+        Where can I read more?
+      </a>
     </div>
```
Klemmensen, line 67: question:

This (h2 vs h3) was previously agreed and approved with designer, has the design changed?

Klemmensen (changes requested):

libs/eo/faq/shell/src/lib/eo-faq-shell.component.ts
MadsThy: What do you mean? 😄

Klemmensen: Linting is splitting the tags up during build - This is automated. So manual updates, will be reverted on the next build where linting will kick in again.

So, you should update the linting rules in the build pipeline, and not the tags in the templates :-)

MadsThy: I will find a linter rule in the future to avoid having this as a topic :P

libs/eo/faq/shell/src/lib/eo-faq-shell.component.ts
MadsThy: Checking with UX again. Will get back to you :)

MadsThy: UX changed their design to H3 now, so the code has been changed to reflect this now :)

Klemmensen (approved):

PR #483, opened 2022-04-21
Align colors with UX

MadsThy:
## Description

Aligning colors from UX in Figma, with storybook

## References

- https://app.zenhub.com/workspaces/fenris---team-board-20-616bc40121d71900140955f8/issues/energinet-datahub/greenforce-frontend/356


libs/dh/shared/assets/src/assets/login-page/login-unified.html
```diff
@@ -60,7 +60,7 @@
     <style>
```
frontend-specialisten, line 60: Please revert this, as this static page doesn't include watt. This is at Mighty ducs backlog, so will be refactored later on. Same with `libs/dh/shared/assets/src/assets/signup-page/signup-unified.html`

frontend-specialisten (changes requested):
Beautiful, only a few comments where you need to revert some of your boy scouting (sorry). Also when I compare to figma it seems that the text of `states` colors has changed a bit over time. If you would update that also, that would be great!

libs/dh/shared/assets/src/assets/login-page/login-unified.html
MadsThy: Reverted the files :)

frontend-specialisten (approved):

frontend-specialisten (approved):

PR #504, opened 2022-04-29
Introduce new WATT button

MadsThy:
## Description

The design for the WATT button was a little off, and a little too complicated, so i have introduced a new button "watt-button-alt", which i think should replace the current buttons.

## References

https://app.zenhub.com/workspaces/fenris---team-board-20-616bc40121d71900140955f8/issues/energinet-datahub/energy-origin-issues/454

### DIFF
--- a/libs/ui-watt/src/lib/components/button/watt-button-alt.component.ts
+++ b/libs/ui-watt/src/lib/components/button/watt-button-alt.component.ts
@@ -1,0 +1,33 @@
+import { Component, Input, ViewEncapsulation } from '@angular/core';
+import { WattIcon } from '../../foundations/icon';
+
+export type WattButtonVariant = 'primary' | 'secondary' | 'text' | 'icon';
+export type WattButtonSize = 'normal' | 'large';
+
+@Component({
+  selector: 'watt-button-alt',
+  template: `
+    <button mat-button [ngClass]="variant" [disabled]="disabled">
+      <watt-spinner
+        *ngIf="loading"
+        [diameter]="18"
+        class="content-grid-item content-grid-item-spinner"
+      ></watt-spinner>
+      <watt-icon *ngIf="hasIcon()" [name]="icon"></watt-icon>
+      <ng-content></ng-content>
+    </button>
+  `,
+  styleUrls: ['./watt-button-alt.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class WattButtonAltComponent {
+  @Input() icon?: WattIcon;
+  @Input() variant: WattButtonVariant = 'primary';
+  @Input() size: WattButtonSize = 'normal';
+  @Input() disabled = false;
+  @Input() loading = false;
+
+  hasIcon(): boolean {
+    return !!this.icon;
+  }
+}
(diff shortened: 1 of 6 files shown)


frontend-specialisten (changes requested):
Looks good, only a few comments :) 
- I would love to see some specs
- It would be nice if we could sort the pages for the button so ovweview was on top
- Regarding the strategy of first having a `watt-button-alt` I think this should just replace the old button components taken breaking changes and the project current size in consideration. The only breaking change I see is the `icon button` which should be manageable to just do a search and replace, maybe I've overseen something? 

(I'm not a huge fan of temporary stuff merged to main, who knows how long it's going to exist so for me that would be preferable to avoid that if possible).

MadsThy:
> Looks good, only a few comments :)
> 
> * I would love to see some specs
> * It would be nice if we could sort the pages for the button so overview was on top
> * Regarding the strategy of first having a `watt-button-alt` I think this should just replace the old button components taken breaking changes and the project current size in consideration. The only breaking change I see is the `icon button` which should be manageable to just do a search and replace, maybe I've overseen something?
> 
> (I'm not a huge fan of temporary stuff merged to main, who knows how long it's going to exist so for me that would be preferable to avoid that if possible).

- Will do a SPEC file for this button, no problem!
- I will change the sorting so Overview goes on top :)
- If you agree that i can change things, i would love to merge this button first, and then do a PR where i change the buttons around all over the place, and then a PR that deletes the old button :) Would that work for you @frontend-specialisten ? :)

libs/ui-watt/src/lib/components/button/watt-button-alt.component.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Component, Input, ViewEncapsulation } from '@angular/core';
+import { WattIcon } from '../../foundations/icon';
+
+export type WattButtonVariant = 'primary' | 'secondary' | 'text' | 'icon';
+export type WattButtonSize = 'normal' | 'large';
+
+@Component({
+  selector: 'watt-button-alt',
+  template: `
+    <button mat-button [ngClass]="[variant, size]" [disabled]="disabled">
```
dzhavat, line 26: issue: quick observation (haven't done a complete review) - adding classes like `primary`, `secondary`, `normal`, etc. might end up being targeted by unrelated CSS because they are too general. Consider renaming the classes to `watt-button--primary`, `watt-button--secondary`, etc.

frontend-specialisten: Agreed, as long the style encapsulation is none we should prefix our classes

MadsThy: Fixed :)

dzhavat: Nice work so far! My previous comment should've been more specific but the `size` class has the same issues - `normal` and `large` are too general. Maybe `watt-button--normal` and `watt-button--large`?

MadsThy: Oh right :) My bad! Will get it up now :)

MadsThy: Done @dzhavat 😄

PR #511, opened 2022-05-05
feat(ui-watt): add single selection of datepicker

frontend-specialisten:
## Description

- Rename `watt-date-range-input` -> `watt-datepicker`
- Add support for single date mask / datepicker

## References

- #373

PR #517, opened 2022-05-06
feat(watt): add date picker to date-range component 

dzhavat:
### DataHub
- increase the `initial` bundle budget size to 1.5 Mb

### Watt Design System
- add support for selecting a date range by using a picker

date-picker

## Notes

- `.` suffix of dates in calendar will be removed by #420

## References

- #412


libs/ui-watt/src/lib/components/input/date-range-input/watt-date-range-input.component.ts
```diff
@@ -32,13 +32,18 @@ import {
...
-import { Subject, takeUntil } from 'rxjs';
+import { combineLatest, map, Subject, takeUntil, tap } from 'rxjs';
+import { parse } from 'date-fns';
+import { formatInTimeZone } from 'date-fns-tz';
 
 import { WattInputMaskService } from '../shared/watt-input-mask.service';
 import { WattRangeInputService } from '../shared/watt-range-input.service';
 import { WattRange } from '../shared/watt-range';
 
+const dateTimeFormat = 'dd-MM-yyyy';
```
frontend-specialisten, line 45: Suggestion: This should properly be consolidated with `getInputFormat` to support localization, but we can do this in #420.

frontend-specialisten (approved):

libs/ui-watt/src/lib/components/input/date-range-input/watt-date-range-input.component.ts
dzhavat: Agree. Let's look at it when working on that issue.

frontend-specialisten (approved):
Specs with integration tests will be added in #511

PR #518, opened 2022-05-09
feat(ui-watt): add chips component

ManBearTM:
## Description

Simple implementation of chips, covering only the "choice chips variant". 

## References

- #487

### DIFF
--- a/libs/ui-watt/src/lib/components/chips/watt-chips.component.ts
+++ b/libs/ui-watt/src/lib/components/chips/watt-chips.component.ts
@@ -27,7 +27,7 @@ export interface WattChipsOption {
   value: string;
 }
 
-export type WattChipsSelection = WattChipsOption | null;
+export type WattChipsSelection = string | null;
 
 @Component({
   encapsulation: ViewEncapsulation.None,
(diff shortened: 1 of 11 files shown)


libs/ui-watt/src/lib/components/chips/watt-chips.component.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Simple implementation of Material "Choice chips".
+ * @see https://material.io/components/chips#choice-chips
+ */
+@Component({
+  encapsulation: ViewEncapsulation.None,
+  selector: 'watt-chips',
+  styleUrls: ['./watt-chips.component.scss'],
+  templateUrl: './watt-chips.component.html',
```
frontend-specialisten, line 42: Suggestion: Add `changeDetection: ChangeDetectionStrategy.OnPush`

Here are some articles you can take a look at: 
https://angular.io/api/core/ChangeDetectionStrategy
https://www.thinktecture.com/en/angular/whats-the-hype-onpush/
https://blog.angular-university.io/onpush-change-detection-how-it-works/

libs/ui-watt/src/lib/components/chips/watt-chips.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+} as Meta<WattChipsComponent>;
+
+export const Overview: Story<WattChipsComponent> = (args) => ({
+  props: args,
+});
+
+Overview.args = {
+  selection: 'day',
+  options: [
+    { label: 'Dag', value: 'day' },
```
frontend-specialisten, line 39: Suggestion: Change the labels to be english as we haven't used danish other places (at least what I know of). I think it's a good idea that `label` and `value` aren't the same. Maybe just change the value to be some kind of id instead?

frontend-specialisten (changes requested):
Praise: Looks good! Only some minor comments :)

libs/ui-watt/src/lib/components/chips/watt-chips.component.ts
ManBearTM: 0f6e24b7bdcc3c1e7b1b4743df4c68b8d0284371

libs/ui-watt/src/lib/components/chips/watt-chips.stories.ts
ManBearTM: 04ee9769471f6ba315058895cef8f91fb266a544

frontend-specialisten (approved):

PR #522, opened 2022-05-10
feat(ui-watt): watt-dropdown showResetOption

michaelmajgaard:
watt-dropdown now updates if options are set again after init. also added option to remove the empty option in single select mode


libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.ts
```diff
@@ -112,6 +114,11 @@ export class WattDropdownComponent
    */
   @Input() multiple = false;
 
+  /**
+   * Sets support for deselecting a selected option in single select mode.
+   */
+  @Input() emptyOption = true;
```
dzhavat, line 120: suggestion: rename to `showResetOption`

libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.html
```diff
@@ -36,7 +36,7 @@
     ></ngx-mat-select-search
   ></mat-option>
 
-  <mat-option *ngIf="!multiple">—</mat-option>
+  <mat-option *ngIf="!multiple && emptyOption">—</mat-option>
```
dzhavat, line 39: suggestion: add test that verifies that the reset option can be removed. I can help with that if you want.

dzhavat (changes requested):

libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.ts
```diff
@@ -132,15 +139,21 @@ export class WattDropdownComponent
...
     this.initializePropertiesFromParent();
     this.bindParentValidatorsToControl();
     this.bindControlToParent();
   }
 
+  /**
+   * @ignore
+   */
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.options.currentValue !== changes.options.previousValue) {
```
dzhavat, line 152: issue: this will result in an error if a property other than `options` is updated dynamically. Make access to `currentValue` and `previousValue` optional.

```suggestion
    if (changes.options?.currentValue !== changes.options?.previousValue) {
```

libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.html
dzhavat: Just pushed a couple of tests.

dzhavat:
issue: can you merge `main` and resolve the conflicts? The latest version should be `0.2.43`.

suggestion: update PRs title to something like "feat(watt): update dropdown component" to be somewhat similar to other PRs.

libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.html
michaelmajgaard: So did I :D

PR #527, opened 2022-05-11
(Watt) Implement save_alt icon from Material

MadsThy:
## Description

Implement the save_alt icon from Material Design


frontend-specialisten (approved):
You'll just need to update the version of EO, and then it's ready to go :)

frontend-specialisten (approved):

PR #529, opened 2022-05-11
Add concurrency to actions

MadsThy:
## Description

This adds concurrency to all actions, and makes them cancel the jobs if a PR has commits added quickly after each other


frontend-specialisten (approved):
LGTM

Issue #417, opened 2022-03-16
Time-picker [Implementation]


dzhavat:
**AC**

- [ ] We need to test the optimal time increment/steps with UX
- [ ] Showcase and document in Storybook
- [ ] Should be able to select a time range

**Design**

![image.png](https://images.zenhubusercontent.com/612cb8944df0eb85278b93db/81705013-b8cb-486f-aa98-6709b0523a3a)

**Technical notes**
- Material Design doesn't include range slider which supports multiple handles
- No validation. This is implemented under another user story
- No input mask. This is implemented under another user story

PR #541, opened 2022-05-18
feat(ui-watt): add slider based on ngx-slider

ManBearTM:
## Description
Implementation of range slider based on ngx-slider.

## References

- #417


libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  @Input() max = 100;
+
+  /** The currently selected range value. */
+  @Input() value = [this.min, this.max];
+
+  /**
+   * Emits value whenever it changes.
+   * @ignore
+   */
+  @Output() valueChange = new EventEmitter<number[]>();
```
frontend-specialisten, line 50: Consideration: maybe align the output to be more like the date/time-picker instead of returning an array to be more explicit? So something like `{min: x, max: y}`

frontend-specialisten (changes requested):
Praise: Looks good! only a single comment, and a single consideration

libs/ui-watt/src/lib/components/slider/watt-slider.stories.ts
```diff
@@ -0,0 +1,60 @@
...
+      <watt-slider [(value)]="value"></watt-slider>
+      <p>{{format(value[0])}} - {{format(value[1])}}</p>
+    `,
+  };
+};
+
+TimePicker.args = {
+  max: 96,
+  value: [25, 75],
+};
```
dzhavat, line 60: praise: this looks quite nice. Good job 💪

question: do we want to show a time picker-looking functionality as part of the slider stories? We know that the slider component is going to be integrated with the time picker component, so we should probably leave keep this story when we do that. Another point if favor of that is that including the functionality here might lead to the wrong impression that this is how a time picker component should be implemented in the apps.

What do you think? Also @frontend-specialisten

dzhavat (changes requested):

libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
ManBearTM: Makes sense, will do 🙂

libs/ui-watt/src/lib/components/slider/watt-slider.stories.ts
ManBearTM: My initial idea was to remove this story once the time picker is in place, so sounds like it aligns with your thoughts as well? I created this story primarily for testing how it would work for selecting a time range. I could also just remove it right away if that is better, I am fine either way 🙂

frontend-specialisten: I think it's totally fine as @ManBearTM has planned. Currently we don't have any usecases for the slider alone, so I don't see any problems having this until the timepicker has been finished :)

libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
```diff
@@ -0,0 +1,60 @@
...
+})
+export class WattSliderComponent {
+  /** The lowest permitted value. */
+  @Input() min = 0;
+
+  /** The greatest permitted value. */
+  @Input() max = 100;
+
+  /** The currently selected range value. */
+  @Input() value = [this.min, this.max];
```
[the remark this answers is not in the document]
ManBearTM: 44fb76d3857b197f89f6284c51f46cdcda2226ed

libs/ui-watt/src/lib/components/slider/watt-slider.component.scss
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+watt-slider {
+  display: block;
+
+  .ngx-slider {
+    height: 8px;
+    margin: 10px 0;
```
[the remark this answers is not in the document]
ManBearTM: Question: Does the spacing system only apply to "whitespace" such as margin and padding or is it for all "distances" including width and height? The slider handles are `24px` which doesn't exist as a spacing variable and any of the other variables does not fit the design very well.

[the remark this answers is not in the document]
ManBearTM: 189466b30bf408b45513e4793352a0df364583bd

libs/ui-watt/src/lib/components/slider/watt-slider.component.html
```diff
@@ -0,0 +1,29 @@
...
+  [highValue]="value[1]"
+  (valueChange)="onChange([$event, value[1]])"
+  (highValueChange)="onChange([value[0], $event])"
+  [options]="{
+    floor: min,
+    ceil: max,
+    minRange: 1,
+    hideLimitLabels: true,
+    hidePointerLabels: true
+  }"
```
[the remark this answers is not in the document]
ManBearTM: ec4674d200f853db27decf2d5e9f0ebee4d7d9d5

libs/ui-watt/src/lib/components/slider/watt-slider.stories.ts
dzhavat: Let's leave it as is. We'll move it as part of implementing the timepicker.

libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
```diff
@@ -0,0 +1,79 @@
...
+  encapsulation: ViewEncapsulation.None,
+  selector: 'watt-slider',
+  styleUrls: ['./watt-slider.component.scss'],
+  templateUrl: './watt-slider.component.html',
+})
+export class WattSliderComponent {
+  /** The lowest permitted value. */
+  @Input() min = 0;
+
+  /** The greatest permitted value. */
```
dzhavat, line 46: nitpick:

```suggestion
  /** The highest permitted value. */
```

libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  get options(): Options {
+    return {
+      floor: this.min,
+      ceil: this.max,
+      minRange: 1,
+      hideLimitLabels: true,
+      hidePointerLabels: true,
+    };
+  }
```
dzhavat, line 69: suggestion:

```suggestion
  options: Options = {
    floor: this.min,
    ceil: this.max,
    minRange: 1,
    hideLimitLabels: true,
    hidePointerLabels: true,
  };
```

dzhavat (changes requested):

build/infrastructure/eo/chart/values.yaml
```diff
@@ -4,7 +4,7 @@ eo-base-helm-chart:
       replicaCount: 2
       image:
         repository: ghcr.io/energinet-datahub/eo-frontend-app
-        tag: 0.2.51
+        tag: 0.2.52
```
dzhavat, line 7: issue: can you merge `main` and resolve the conflicts here and in `Chart.yaml`?

libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
ManBearTM: The reason I'm using a getter here is to allow setting the `floor` and `ceil` values using `min` and `max`. If I just set options directly like your suggestion, the `floor` and `ceil` will just always be the default values (`0` and `100` respectively). That is also why I initially had options inlined in the template. Alternatively we could do something like [this](https://angular-slider.github.io/ngx-slider/demos#dynamic-options-slider), but then we'd have to listen to changes to `min` and `max` somehow and that just seems needlessly complicated IMHO. Am I missing something here? 🙂

dzhavat: No, you're not. You're right! Sorry for the confusion :)

dzhavat (approved):

libs/ui-watt/src/lib/components/slider/watt-slider.component.scss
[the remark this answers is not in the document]
frontend-specialisten: @ManBearTM The intention is whitespace so not `width` or `height`, however I would apply it to `top` or `left` as well :)

frontend-specialisten (approved):

Issue #506, opened 2022-05-02
Fix b2c healthcheck


frontend-specialisten:
Reason: U-002 is down (404 Error) https://ambitious-coast-027d0aa03.azurestaticapps.net/

- [ ] adjust U-002 from healthcheck to newly created URL: https://wonderful-field-057109603.1.azurestaticapps.net
- [ ] Add https://itlev.datahub.dk/ if not already added

PR #546, opened 2022-05-19
fix(dh): healthchecks

frontend-specialisten:
## Description

Fix healthchecks of dh to correspond with the new URL of U-002 and add itlev environment 

## References

- #506

### DIFF
--- a/apps/dh/e2e-dh/src/integration/b2c.spec.ts
+++ b/apps/dh/e2e-dh/src/integration/b2c.spec.ts
@@ -21,11 +21,10 @@ const environments = [
     name: 'U001',
     url: 'https://jolly-sand-03f839703.azurestaticapps.net',
   },
-  // TODO: Re-add this website once it has been fixed and returns 200 again
-  /*   {
+  {
     name: 'U002',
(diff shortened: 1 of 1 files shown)


apps/dh/e2e-dh/src/integration/b2c.spec.ts
```diff
@@ -38,6 +37,10 @@ const environments = [
     name: 'B002',
     url: 'https://purple-forest-07e41fb03.azurestaticapps.net',
   },
+  {
+    name: 'itlev',
+    url: 'https://itlev.datahub.dk',
+  },
```
dzhavat, line 43: question: Not sure if this is a valid concern but right now anyone can find these URLs on GitHub and create a user. Then they can do anything in the app, including changing names, triggering actions, reading information. Especially with the ITLev environment, isn't that risky?

frontend-specialisten: I see your point, however everything should go through APIM so they shouldn't be able to do stuff they aren't allowed to. Anyways better safe than sorry, so I will remove it for now.

PR #569, opened 2022-05-30
Add License check for Angular dependencies

MadsThy:
## Description

This PR adds support for automatic checking of licenses for the dependencies we use in production code.

It only checks the DIRECT dependencies and not DEV dependencies. It uses the .allowed-licenses.json file to uphold the list of allowed licenses.

After this PR has been merged, the license-check step will be added as a required check for every PR going forward, so we don't suddenly introduce software in production that has licenses that we don't allow.


.github/workflows/frontend-ci.yml
```diff
@@ -336,6 +336,41 @@ jobs:
         if: ${{ env.is-pull-request == 'false' }}
         run: yarn lint --max-parallel=$NX_MAX_PARALLEL
 
+  License_check:
+    name: License check
```
frontend-specialisten, line 340: Consider a more explicit naming on this, since we already have "checks" which adds and checks for our own license.

frontend-specialisten (changes requested):

.github/workflows/frontend-ci.yml
MadsThy: How about "Code Dependencies License Check"? :)

MadsThy: or "Production Dependencies License Check"

frontend-specialisten: Fine by me. 
`Allowed licenses check` could also be an option, whatever you choose just that we easily know what's it all about and don't confuse it with our existing license check :)

PR #573, opened 2022-06-01
feat(watt): add custom date adapter and move shared localization

frontend-specialisten:
## Description

- Add custom date-adapter 
- Move `danish-datetime.module` to watt
- Move `configuration-danish-locale` and `test-util-local` to shared resources (gf)
- Refactor `dh` to new structure of shared localization resources
- Remove duplicates of localization modules in storybook
- Change locale of the dateadapter on language selection (dh)

## References

- #420


libs/dh/globalization/feature-language-picker/src/lib/language-picker/dh-language-picker.component.ts
```diff
@@ -31,10 +33,18 @@ import { map, Observable } from 'rxjs';
 })
 export class DhLanguagePickerComponent {
   activeLanguage$: Observable<DisplayLanguage> =
-    this.transloco.langChanges$.pipe(map(toDisplayLanguage));
+    this.transloco.langChanges$.pipe(
+      map(toDisplayLanguage),
+      tap((language) => {
+        (this.dateAdapter as WattDateAdapter).setActiveLocale(language);
```
dzhavat, line 39: suggestion: what about using a service to set the locale instead of exposing `DateAdapter` and `WattDateAdapter`?

We can provide the service as part of `WattDanishDatetimeModule.forRoot()`. Then we inject it here and in `WattDateAdapter` where we listen to language changes and set the locale. Wouldn't that work?

(So something similar to `TranslocoService`)

libs/ui-watt/src/lib/configuration/watt-date-adapter.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.setLocale((language === 'da' ? daLocale : enLocale) as Locale);
+  }
+
+  /**
+   * This is necessary to remove the dots from the date (ordinals) for danish locale in the calendar view.
+   * due to `Intl.DateTimeFormat`
+   */
+  getDateNames(): string[] {
+    const dateNames = super.getDateNames();
+    return this.locale.code === 'da'
```
dzhavat, line 39: suggestion: move the `da` to a variable at the start of the file.

We should also think about injecting these values from the app instead of copying them to Watt. Not as part of this PR but maybe we should look at a pattern similar to how Transloco is configured.

libs/ui-watt/src/lib/components/input/shared/watt-picker-base.ts
```diff
@@ -388,6 +388,6 @@ export abstract class WattPickerBase
    * @ignore
    */
   private setSingleValue(value: string, input: HTMLInputElement) {
-    input.value = value;
+    input.value = !value ? '' : value;
```
dzhavat, line 391: question: what would `value` be other than string?

dzhavat (changes requested):

libs/dh/globalization/feature-language-picker/src/lib/language-picker/dh-language-picker.component.ts
frontend-specialisten: How will you listen for language changes without making Watt depending on transloco? (this is the reason why I did it like this, and to keep it simple).

dzhavat: I don't think Watt should depend on Transloco. My suggestion was to create a pattern similar to what Transloco is doing - to provide a simple service that is injected in Watt and here and use that to send values between app and Watt. So importing `DateAdapter` and `WattDateAdapter` wouldn't be necessary.

libs/ui-watt/src/lib/components/input/shared/watt-picker-base.ts
```diff
@@ -387,7 +387,7 @@ export abstract class WattPickerBase
   /**
    * @ignore
    */
-  private setSingleValue(value: string, input: HTMLInputElement) {
-    input.value = value;
+  private setSingleValue(value: string | undefined, input: HTMLInputElement) {
+    input.value = !value ? '' : value;
```
ManBearTM, line 391: suggestion: use default parameter instead:
```suggestion
  private setSingleValue(value = '', input: HTMLInputElement) {
    input.value = value;
```

ManBearTM (approved):
🥇

dzhavat (changes requested):

libs/ui-watt/src/lib/components/input/shared/watt-picker-base.ts
dzhavat: Just checked the setup that you're describing with the form control from search metering point. The reason for seeing `undefined` is because the form control's value is set to `undefined` in the `setInitialValue` method (the `q` route param is `undefined`). Is that a realistic setup though?

frontend-specialisten: I think it's better that we handle it (now we know it can actually happen), rather than just ignoring it and throw an error.

frontend-specialisten: Types should be adjusted now.

dzhavat (approved):

PR #595, opened 2022-06-03
Add loading to origin of energy

MadsThy:
## Description

- Add a loading indicator to the Origin of Energy pie chart
- Adjust API for pie chart data


C-Christiansen (approved):
HP

PR #614, opened 2022-06-08
Correct for NaN

MadsThy:
## Description

Bugfixing before it becomes an issue, to check for null or NaN in API response


Exobitt (approved):
Lgtm

PR #624, opened 2022-06-09
Change origin of energy api

MadsThy:
## Description

Backend API for origin of energy changed. This change is to reflect that.

### DIFF
--- a/libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy.service.ts
+++ b/libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy.service.ts
@@ -33,12 +33,12 @@ export interface EoOriginOfEnergyResponse {
         waste: number;
         straw: number;
         oil: number;
-        naturalgas: number;
+        naturalGas: number;
         coal: number;
-        biogas: number;
+        bioGas: number;
         solar: number;
-        windonshore: number;
-        windoffshore: number;
+        windOnshore: number;
+        windOffshore: number;
       };
     }
   ];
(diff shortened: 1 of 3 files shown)


libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy.service.ts
```diff
@@ -23,22 +23,22 @@ import {
 import { map, Observable } from 'rxjs';
 
 export interface EoOriginOfEnergyResponse {
-  energysources: [
+  energySources: [
```
Exobitt, line 26: Why is this camelCase and not the entries?
Like naturalgas etc.

Exobitt (changes requested):

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy.service.ts
MadsThy: Because we are not in control of what comes back inside the ratios object, but everything outside it is ours to control, and there we use camelCase :)

I asked Claus to use a toLowerCase() function on the entries in ratios, because otherwise they would have returned "Straw" instead of "straw", and NaturalGas instead of "naturalGas". And since i dont want him to make a lot of if statements, we opted for toLowerCase :)

Exobitt (approved):
lgtm

Exobitt (changes requested):

libs/eo/origin-of-energy/shell/src/lib/eo-origin-of-energy.service.ts
```diff
@@ -23,22 +23,22 @@ import {
...
       dateTo: number;
       renewable: number;
-      source: {
+      ratios: {
         wood: number;
         waste: number;
         straw: number;
         oil: number;
-        naturalGas: number;
+        naturalgas: number;
```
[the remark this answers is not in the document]
MadsThy: Agreed :) I will ask Claus if they can solve it easily

Exobitt (approved):
lgtm

PR #625, opened 2022-06-10
gridareas for actor shown in organisation overview

michaelmajgaard:


libs/dh/market-participant/feature-organization/src/lib/overview/dh-market-participant-organization-overview.component.html
```diff
@@ -69,7 +69,7 @@
           </div>
         </ng-container>
         <ng-container *ngIf="element.actor === undefined">
-          <div class="actor-roles-cell-container">
+          <div class="list-cell-container">
```
frontend-specialisten, line 72: Nitpick: Since `div.list-cell-container` is the only content of the `ng-container`, the if statement could just be applied to the `div` directly and remove `ng-container`.

**Applying this suggestion, remember to remove the `ng-container` to avoid double if statements**

```suggestion
          <div class="list-cell-container" *ngIf="element.actor === undefined">
```

libs/dh/market-participant/feature-organization/src/lib/overview/dh-market-participant-organization-overview.component.html
```diff
@@ -82,8 +82,18 @@
         {{ t("columns.gridAreas") }}
       </mat-header-cell>
       <mat-cell *matCellDef="let element">
-        <!-- gridAreas is currently always undefined, but always returning the dash is correct for now. -->
-        {{ null | dhEmDashFallback }}
+        <ng-container *ngIf="element.actor !== undefined">
```
frontend-specialisten, line 85: Nitpick: This `ng-container` seems to be redundant as well. Consider just applying the `*ngIf` to the child instead. 
One of the cases where you wanna use `ng-container` is when an element/node already has a structural directive applied ex. `*ngFor` or `*ngIf` (because only a single structural directive, can be applied for an element/node) but this is not the case for `div.market-roles-cell`.

frontend-specialisten (approved):
LGTM, only two suggestions you can consider to apply before merging :)

frontend-specialisten (approved):

frontend-specialisten (approved):

PR #628, opened 2022-06-13
Feature flag service and config

frontend-specialisten:
## Description

- Add `dh app environment` configuration, like our other configuration files (BFF, B2C). 
- Add `feature-flags` library which is based on environment (environment is going to be replaced by CD in another issue).
- Add a test which ensures no feature flags are older than 14 days
 

## Note

- #405


libs/dh/shared/environments/src/lib/app-environment/dh-app-environment.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface DhAppEnvironmentConfig {
+  current: string;
+}
+
+export enum DhAppEnvironments {
+  local = 'localhost',
+  dev = 'u-001',
+  test = 't-001',
+  preProd = 'b-001',
+  prod = 'b-002',
```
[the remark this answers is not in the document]
frontend-specialisten: That's a good question, let's remove it for now and add it if the environment is being requested.

libs/dh/shared/environments/src/lib/app-environment/dh-app-environment.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export const dhAppEnvironmentToken = new InjectionToken<DhAppEnvironmentConfig>(
+  'dhAppEnvironmentToken',
+  {
+    factory: (): DhAppEnvironmentConfig => {
+      if (environment.production) {
+        throw new Error('No DataHub app environment config provided.');
+      }
```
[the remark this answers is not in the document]
frontend-specialisten: If no environment is found in production build we throw an error, this is how we have done it with our other configs as well. However we could also handle the error in this scenario maybe just setting it to `prod`?

libs/dh/shared/environments/src/lib/app-environment/dh-app-environment.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export interface DhAppEnvironmentConfig {
+  current: DhAppEnvironments;
+}
+
+export enum DhAppEnvironments {
+  local = 'localhost',
+  dev = 'u-001',
+  test = 't-001',
+  prod = 'b-002',
```
dzhavat, line 30: suggestion: it'd be nice to still include `b-001` (preProd) in this list but don't expose it publicly. So either using a TypeScript util or maybe having two lists (one exposed and one internal)?

libs/dh/shared/feature-flags/src/lib/feature-flags.ts
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { DhAppEnvironments } from '@energinet-datahub/dh/shared/environments';
+
+export type DhFeatureFlag = {
+  created: string;
+  disabledEnvironments: DhAppEnvironments[];
```
dzhavat, line 21: suggestion: the essential purpose for using feature flags, to me, is to "enable the intentional development, deployment and release of new functionality without affecting the existing one". When I look that, I think having an "enabled unless disabled" approach seems too loose. When it comes to "hiding" something behind a feature flag, it should be as hard as possible to accidentally "leak" it unintentionally. On the other hand, the act of "revealing" it  should be as intentional as possible.

The problem with `disabledEnvironments` is that forgetting to add an env to the list accidentally releases "the thing". Adding a new environment (or chaining the name of an existing one) will lead to the same result. 

Having said that, I think our approach should be the opposite, i,e, `enabledInEnvironments`. It will be more verbose but less error prone. Forgetting to add an env to the list will do nothing, "the thing" will not be released.

What do you think?

dzhavat (changes requested):

libs/dh/shared/environments/src/lib/app-environment/dh-app-environment.ts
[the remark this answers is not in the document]
ManBearTM: Your call, I'm fine either way. I just wanted to understand 🙂

libs/dh/shared/environments/src/lib/app-environment/dh-app-environment.ts
```diff
@@ -0,0 +1,47 @@
...
+export const dhAppEnvironmentToken = new InjectionToken<DhAppEnvironmentConfig>(
+  'dhAppEnvironmentToken',
+  {
+    factory: (): DhAppEnvironmentConfig => {
+      if (environment.production) {
+        throw new Error('No DataHub app environment config provided.');
+      }
+
+      // Used for unit and integration tests
+      return dhLocalAppEnvironment as unknown as DhAppEnvironmentConfig;
```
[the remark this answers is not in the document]
frontend-specialisten: Reason is because we're loading the config from json, typescript sees it as a `string` which is not `DhAppEnvironment`. If you remove it you should see an error.

libs/dh/shared/feature-flags/src/lib/feature-flags.ts
frontend-specialisten: I see your point, however as you said it would properly be more verbose since I guess it typically should be enabled for dev and test but disabled for prod. Also the fact that "the users" should actively use the feature-flags functionality, and if they don't it will also be released (maybe by accident) and that would be pretty hard for us to guard us against that?

libs/dh/shared/environments/src/lib/app-environment/dh-app-environment.ts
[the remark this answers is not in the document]
frontend-specialisten: `resolveJsonModule` is actually already set to true in `tsconfig.base.json`, so properly it's something else which causes the issue. If you wanna give it a try @dzhavat or @ManBearTM please do :)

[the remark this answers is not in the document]
dzhavat: Ah, it's type incompatibility. `DhAppEnvironmentConfig.current` is `DhAppEnvironment` whereas the `dhLocalAppEnvironment.current` is seen as `string` by TS.

libs/dh/shared/feature-flags/src/lib/feature-flags.ts
dzhavat: > Also the fact that "the users" should actively use the feature-flags functionality, and if they don't it will also be released (maybe by accident) 

How will something be released by accident in this case?

The way I see it, having a `disabled unless enabled` approach means that "the user" must intentionally add an env to the list in order to enable "the thing" for that env. So releasing something by accident in that case means adding an env to the list that shouldn't be there (either intentionally or not). In the opposite approach (i.e. "enabled unless disabled") , NOT adding an env will mean releasing "the thing". 

So is "adding something [extra] to a list" vs "forgetting to add something to a list" more error prone? I'd argue it's the latter.

As for verbosity, it seems like it's the same. In one case is adding "dev" and "test" env, in the other case is adding "pre-prod" and "prod" env.

libs/dh/shared/feature-flags/.eslintrc.json
```diff
@@ -0,0 +1,36 @@
...
+  "extends": ["../../../../.eslintrc.json"],
+  "ignorePatterns": ["!**/*"],
+  "overrides": [
+    {
+      "files": ["*.ts"],
+      "extends": [
+        "plugin:@nrwl/nx/angular",
+        "plugin:@angular-eslint/template/process-inline-templates"
+      ],
+      "rules": {
```
dzhavat, line 11: suggestion: what do you think of adding `"no-dupe-keys": ["error"],` here in order to raise a lint error in case a duplicate feature flag is used?

libs/dh/shared/feature-flags/src/lib/feature-flags.ts
frontend-specialisten: We keep it as is, and gather feedback from the users.

libs/dh/shared/feature-flags/.eslintrc.json
frontend-specialisten: Nice one!

ManBearTM (dismissed):

libs/dh/shared/feature-flags/src/lib/feature-flags.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+test.each(
+  // Avoid "Error: `.each` called with an empty Array of table data."
+  featureFlagCases.length > 0 ? featureFlagCases : [['dummy feature flag', 0]]
+)(
+  `The feature flag: "%s" must not be older than ${maxAgeOfDays} days, but is %s days old!`,
+  (_, ageOfFeatureFlag) => {
+    expect(ageOfFeatureFlag).toBeLessThanOrEqual(maxAgeOfDays);
+  }
+);
```
dzhavat, line 46: suggestion: let's move the conditional to two separate tests.

```suggestion
if (featureFlagCases.length === 0) {
  test('empty feature flags list', () => {
    expect(featureFlagCases).toStrictEqual([]);
  });
}

// Note: `test.each` must not be called with an empty array
if (featureFlagCases.length > 0) {
  test.each(featureFlagCases)(
    `The feature flag: "%s" must not be older than ${maxAgeOfDays} days, but is %s days old!`,
    (_, ageOfFeatureFlag) => {
      expect(ageOfFeatureFlag).toBeLessThanOrEqual(maxAgeOfDays);
    }
  );
}
```

frontend-specialisten: The dummy test only exist because our feature flags are empty which will cause jest to fail, so splitting it don't make sense here.

libs/dh/shared/feature-flags/src/lib/feature-flags.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+import { DhAppEnvironment } from '@energinet-datahub/dh/shared/environments';
+
+export type DhFeatureFlagEnvironment = Exclude<
+  DhAppEnvironment,
+  DhAppEnvironment.preProd
+>;
+
+export type DhFeatureFlag = {
+  created: string;
```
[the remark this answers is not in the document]
frontend-specialisten: Example has now been added, we also have an upcoming story of documenting everything.

dzhavat (dismissed):

dzhavat:
praise: good work! Exciting to see how the other teams are going to use this feature.

dzhavat (approved):

libs/dh/shared/feature-flags/src/lib/feature-flags.ts
```diff
@@ -37,8 +37,8 @@ export const DhFeatureFlagsConfig = makeFeatureFlags({
    * Feature flag example:
    *
    * 'example-feature-flag': {
-   *  created: '01-01-2022',
-   *  disabledEnvironments: [DhAppEnvironment.prod],
+   *     created: '01-01-2022',
+   *     disabledEnvironments: [DhAppEnvironment.prod],
    * },
    *
```
dzhavat, line 43: suggestion: move the code block on top of the variable.

suggestion: change variable's name to start with a lower-case for consistency.

dzhavat (approved):

Issue #563, opened 2022-05-30
Update dependencies


AnnSophieChris:
**Description:**
Update dependencies from #490

**AC:**
- [ ] Ensure feature teams and Fenris accepts PR
- [ ] Go thought the list of dependencies with Bjørn

**Notes**
- `@ngworker/spectacular` cannot be updated to v0.3.0 because it's not compatible with Angular v14
- `cypress` v10 requires some refactoring. Team Fenris will take care of that. They use Cypress for E2E testing
- `jest` is dependent on Nx. They will release a version that migrates to Jest v28 at some point
- `jest-preset-angular` is dependent on Jest.
- `ts-jest` is dependent on Jest

AnnSophieChris:
Estimate pending completion of #490

PR #636, opened 2022-06-14
chore: update NX to latest version

dzhavat:
### Workspace
- update NX to latest version
- update Angular and related packages to v14
- update Playwright to latest version
- update NOTICE file
- run migrations to adjust files
  - add `$schema` to files
  - set target to `es2020` for libraries
  - use untyped form controls
  - adjust configuration in Nx file

### DataHub
- adjust path for Watt assets
- fix custom Playwright executor

### Energy Origin
- broaden error message check in e2e tests
- adjust path for Watt assets
- bump project version

### Watt
- delete temporary `ui-watt` assets directory
- fix errors related to typed form controls
- fix Storybook types
- remove unnecessary `entryComponents` property

## References

- #563


apps/dh/api-dh/project.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "..\\..\\..\\node_modules\\nx\\schemas\\project-schema.json",
```
MadsThy, line 2: What the actual fudge is this pathing? 🤣

frontend-specialisten (approved):

apps/eo/app-eo/project.json
```diff
@@ -19,7 +20,7 @@
             "output": "assets"
           },
           {
-            "input": "libs/ui-watt/src/assets/ui-watt",
```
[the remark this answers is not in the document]
dzhavat: Yes, it's correct. The reason it was here was due to a bug in Storybook in v6.4.x. Storybook is not updated to v6.5.x where the bug is fixed. So this workaround is not necessary anymore. See https://github.com/Energinet-DataHub/greenforce-frontend/pull/636/files/1d370e454a4fc7033e2255da095bb31a71fb28bb#diff-4e9ced2e5b8238ea24312e391957b7780f06290fc90ca186e220d1e74b919ff7

apps/dh/api-dh/project.json
MadsThy, line 2: Suggestion: 
```suggestion
  "$schema": "node_modules/nx/schemas/project-schema.json",
```

And this one goes for all the nx project-schema changes :)

libs/ui-watt/src/lib/components/input/+storybook/storybook-input-wrapper.component.ts
```diff
@@ -57,7 +57,7 @@ import { FormControl } from '@angular/forms';
     </watt-error>
     <watt-hint *ngIf="hasHint">Some hint</watt-hint>
     <watt-hint *ngIf="hasHint" align="end"
-      >{{ exampleFormControl.value.length }} / 256</watt-hint
+      >{{ exampleFormControl.value?.length }} / 256</watt-hint
```
MadsThy, line 60: What do we want to leave the user with here if this value is null? :)

libs/ui-watt/src/lib/components/input/datepicker/+storybook/watt-datepicker-reactive-forms.stories.ts
```diff
@@ -69,7 +69,7 @@ const template = `
 
 const initialValue = '22-11-3333';
 
-export const withFormControl: Story<WattDatepickerComponent> = (args) => ({
```
MadsThy, line 72: Please clean up this files imports if you are removing this one from usage :)

package.json
```diff
@@ -39,7 +39,7 @@
     "lint": "nx workspace-lint && nx run-many --target=lint --max-warnings=0 --all --parallel",
     "lint:fix": "nx workspace-lint && nx run-many --target=lint --max-warnings=0 --fix --all --parallel",
     "ng": "nx",
-    "ngcc": "ngcc --properties es2015 module browser main",
+    "ngcc": "ngcc --properties es2020 module browser main",
```
MadsThy, line 42: Have people generally signed off on us building for es2020 instead of es2015? :) I don't remember having this discussion with anyone, so i am interested in what people think.

package.json
```diff
@@ -82,37 +82,38 @@
...
     "@ngworker/spectacular": "0.2.0",
-    "@nrwl/cli": "14.1.9",
-    "@nrwl/cypress": "14.1.9",
-    "@nrwl/eslint-plugin-nx": "14.1.9",
-    "@nrwl/jest": "14.1.9",
-    "@nrwl/linter": "14.1.9",
-    "@nrwl/nx-cloud": "14.0.5",
-    "@nrwl/storybook": "14.1.9",
-    "@nrwl/workspace": "14.1.9",
+    "@nrwl/cli": "14.3.2",
```
MadsThy, line 96: Remember to update NOTICE.md when upgrading packages :)

MadsThy (changes requested):
Left some comments :)

apps/dh/api-dh/project.json
dzhavat: Just pushed a commit with changes. The `../../../` part is still needed though :)

libs/ui-watt/src/lib/components/input/+storybook/storybook-input-wrapper.component.ts
dzhavat: This is a showcase in Storybook and we know for sure that the `value` will not be `null`/`undefined`.

package.json
dzhavat: Done.

libs/ui-watt/src/lib/components/input/+storybook/storybook-input-wrapper.component.ts
MadsThy: Then why the optional chaining? 😄 

(This is a very minor thing. Just wondering)

apps/dh/api-dh/project.json
MadsThy: I just checked, and ./node_modules does not work. Resolving this one :)

dzhavat: Because the lookup starts from where the file is placed. Those on root start with `./node_modules/...`.

libs/ui-watt/src/lib/components/input/+storybook/storybook-input-wrapper.component.ts
dzhavat: Reactive forms in Angular v14 are strongly typed. So a `FormControl` used to be untyped but now it is and has a type `string | null`. If we want the previous behaviour, we'll have to use `new UntypedFormControl()`. I didn't do it here because it's Storybook. But I can also make it use the old behaviour and we can migrate to a typed `FormControl` at a later point.

package.json
dzhavat: I this specific case it's used for compiling dependencies to es2020, not apps. The version for apps is set in `tsconfig.base.json` which is still set to `es2015`.

Generally, Angular itself changed the target from `es2015` to `es2020` in v14 (which I haven't set in this PR). Maybe we should.

As for our apps, DataHub supports only the latest versions of evergreen browsers until we want another strategy. Have you discussed browser support for Energy Origin?

libs/ui-watt/src/lib/components/input/+storybook/storybook-input-wrapper.component.ts
MadsThy: Fair enough :) Then lets resolve it here and leave it be. Was just wondering 😄

package.json
MadsThy: We have not discussed browser support no 😄 But thanks! Yet another reason why I think we should move on the "guild/chapter" idea! :)

From my point of view we should upgrade it all to es2020.

apps/dh/api-dh/source/DataHub.WebApi/openapitools.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "node_modules/@openapitools/openapi-generator-cli/config.schema.json",
+  "$schema": "../../../../../node_modules/@openapitools/openapi-generator-cli/config.schema.json",
```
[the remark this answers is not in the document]
dzhavat: I changed it manually because the path was not correct.

MadsThy (dismissed):

package.json
```diff
@@ -50,22 +50,22 @@
...
+    "@angular/material-date-fns-adapter": "14.0.1",
+    "@angular/platform-browser": "14.0.1",
+    "@angular/platform-browser-dynamic": "14.0.1",
+    "@angular/router": "14.0.1",
     "@azure/msal-angular": "2.2.0",
     "@azure/msal-browser": "2.23.0",
     "@ngneat/transloco": "3.2.0",
-    "@ngrx/component-store": "13.0.2",
-    "@nrwl/angular": "14.1.9",
+    "@ngrx/component-store": "13.2.0",
```
frontend-specialisten, line 67: I can see that NGRX seems to be in release candidate of version `14`, should we create a task for that now so we remember to have an eye on that?

package.json
```diff
@@ -82,37 +82,38 @@
...
   "devDependencies": {
-    "@angular-devkit/build-angular": "13.3.7",
-    "@angular-eslint/eslint-plugin": "13.0.1",
-    "@angular-eslint/eslint-plugin-template": "13.0.1",
-    "@angular-eslint/template-parser": "13.0.1",
-    "@angular/cli": "13.3.7",
-    "@angular/compiler-cli": "13.3.11",
-    "@angular/language-service": "13.3.11",
+    "@angular-devkit/build-angular": "14.0.1",
+    "@angular-eslint/eslint-plugin": "13.2.1",
```
frontend-specialisten, line 86: Is this what NX migrates to? As I know @anglular-eslint is currently running alpha on version `14`, but if this is manual we should properly consider creating a task of keeping an eye on this future release as well.

tools/executors/playwright/playwright.impl.ts
```diff
@@ -136,6 +136,7 @@ async function runPlaywright(
     {
       commands: [playwrightCommand],
       parallel: true,
+      __unparsed__: [],
```
frontend-specialisten, line 139: Praise: good catch!

frontend-specialisten (dismissed):

package.json
dzhavat: I think it will naturally come up during the next dependency update. Which I will suggest is something we do regularly every x weeks to avoid spending a week updating.

package.json
dzhavat: I don't remember whether it was an automatic migration. But will also suggest to update our deps more often so the new version will naturally come up during the next update.

tools/executors/playwright/playwright.impl.js
```diff
@@ -437,17 +437,20 @@ function runPlaywright(baseUrl, opts, context) {
...
+              '"'
+            );
           }
           if (opts.exclude) {
-            playwrightCommand +=
-              ' --grep-invert="' + escapeRegExp(opts.exclude) + '"';
+            playwrightCommand += ' --grep-invert="'.concat(
+              escapeRegExp(opts.exclude),
+              '"'
+            );
```
MadsThy, line 453:
```suggestion
            playwrightCommand = `${playwrightCommand} --grep-invert='${escapeRegExp(opts.exclude)}'`
```

MadsThy (changes requested):
Found some small things i'd like to have us change :)

libs/ui-watt/src/lib/components/input/+storybook/storybook-input-wrapper.module.ts
```diff
@@ -17,13 +17,13 @@
 import { NgModule } from '@angular/core';
 import { FormsModule, ReactiveFormsModule } from '@angular/forms';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+
```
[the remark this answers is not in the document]
dzhavat: We actually follow a "soft" rule where we try to organize the imports:

- 3rd party imports
- 1st party library imports
- 1st party local imports

... with an empty line between each group. We haven't yet found a good automatic way to do this, so we rely on doing it manually if/when we notice it. :P

tools/executors/playwright/playwright.impl.js
```diff
@@ -437,17 +437,20 @@ function runPlaywright(baseUrl, opts, context) {
         case 0:
           projectname = context.projectName;
           sourceRoot = context.workspace.projects[projectname].sourceRoot;
-          playwrightCommand =
```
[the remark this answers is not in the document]
MadsThy: Fair enough :) Then leaving it be.

MadsThy (dismissed):

MadsThy (approved):

PR #646, opened 2022-06-20
chore: update dependencies

dzhavat:
### Workspace 
- update dependencies
- fix RxAngular config

### DataHub
- refactor tests for custom pipes to remove an unnecessary property

## References

- #563


MadsThy (dismissed):
LGTM :)

libs/gf/configuration-danish-locale/src/lib/danish-locale.integration.spec.ts
```diff
@@ -21,10 +21,8 @@ import { formatInTimeZone } from 'date-fns-tz';
 import { DanishLocaleModule } from './danish-locale.module';
 import { spaceToNonBreakingSpace } from './space-to-non-breaking-space';
 
-const dummyPipeName = 'testDummy';
-
 @Pipe({
-  name: dummyPipeName,
+  name: 'testDummyPipeName',
```
ManBearTM, line 25: Name is different than before, don't know if intentional or not. Either way, it probably doesn't matter 🙂

ManBearTM (dismissed):

libs/gf/configuration-danish-locale/src/lib/danish-locale.integration.spec.ts
dzhavat: Had to revert this change because the library we're using to test pipes is not compatible with the newest version of Angular.

The pipe name was changed hoping to make its purpose more clear.

PR #650, opened 2022-06-20
Make functional cookie accept button work

MadsThy:
## Description

People were complaining that the "functional only" button didn't close the popup, so i am making it do so.


Exobitt (approved):
lgtm

PR #661, opened 2022-06-23
Remove decimals from pie chart and put type on metering points

MadsThy:
## Description

Small bugfixes to:

- Remove decimals from the pie chart
- Add type filtering for metering points


Exobitt (approved):
lgtm

PR #664, opened 2022-06-23
feat(ui-watt): modal configuration

### LINKED ISSUE #620: Support for Modal configuration
**Description**

**AC**
- [ ] Should be configurable so it can't be dismissed
- [ ] Should show/hide close button based on above configuration

**Tech notes**
- https://material.angular.io/components/dialog/api#MatDialogConfig

ManBearTM:
## References

Closes #620

### DIFF
--- a/libs/ui-watt/src/lib/components/modal/watt-modal.component.html
+++ b/libs/ui-watt/src/lib/components/modal/watt-modal.component.html
@@ -22,7 +22,7 @@ limitations under the License.
     type="button"
     (click)="close(false)"
   >
-    <watt-icon name="close"></watt-icon>
+    <watt-icon [label]="closeLabel" name="close"></watt-icon>
   </button>
   <h3 class="watt-modal-title">{{ title }}</h3>
   <div class="watt-modal-content"><ng-content></ng-content></div>
--- a/libs/ui-watt/src/lib/components/modal/watt-modal.component.html
+++ b/libs/ui-watt/src/lib/components/modal/watt-modal.component.html
@@ -15,6 +15,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <ng-template #modal>
+  <button
+    *ngIf="!disableClose"
+    mat-icon-button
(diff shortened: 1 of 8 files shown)


libs/ui-watt/src/lib/components/modal/watt-modal.component.html
```diff
@@ -15,6 +15,15 @@
...
 <ng-template #modal>
+  <button
+    *ngIf="!disableClose"
+    mat-icon-button
+    class="watt-modal-close"
+    type="button"
+    (click)="close(false)"
+  >
+    <watt-icon [label]="closeLabel" name="close"></watt-icon>
+  </button>
```
dzhavat, line 26: nitpick: we should discuss this with UX (can also be after merge) because the button's look is different than what we have in Watt today. There's no button in Watt that can really be used as a drop-in replacement without making additional adjustments. So we either need to extend Watt to support such a button or use the existing icon button in order to not deviate from Watt's direction.

dzhavat (approved):

libs/ui-watt/src/lib/components/modal/watt-modal.stories.ts
```diff
@@ -65,4 +65,5 @@ export const Overview: Story<WattModalComponent> = (args) => ({
 Overview.args = {
   title: 'Install Angular',
   size: 'large',
+  disableClose: false,
 };
```
dzhavat, line 69: issue: merge `main` and bump EO 🙃

dzhavat (approved):

libs/ui-watt/src/lib/components/modal/watt-modal.component.html
ManBearTM: Good point! I talked with UX and it is possible to use the `<watt-button>` here, but it requires changing the size of the icon variant so it matches Figma as well as using the `:focus-visible` selector for focus state. But these are potentially breaking changes, so I will create a different PR for that.

dzhavat (approved):

PR #669, opened 2022-06-24
fix(ui-watt): disable slider toggle when input is disabled

ManBearTM:
## Description
Forgot to account for disabled state. This just makes sure the slider cannot
be toggled while the input is disabled.

## References

- #0000

### DIFF
--- a/libs/ui-watt/src/lib/components/input/timepicker/watt-timepicker.component.html
+++ b/libs/ui-watt/src/lib/components/input/timepicker/watt-timepicker.component.html
@@ -57,6 +57,7 @@ limitations under the License.
     mat-icon-button
     class="watt-timepicker-slider-toggle"
     type="button"
+    [disabled]="disabled"
     [attr.aria-pressed]="sliderOpen"
     (click)="toggleSlider()"
     (focusin)="onFocusIn()"


libs/ui-watt/src/lib/components/input/timepicker/watt-timepicker.component.html
```diff
@@ -57,6 +57,7 @@
     mat-icon-button
     class="watt-timepicker-slider-toggle"
     type="button"
+    [disabled]="disabled"
```
dzhavat, line 60: praise: good catch! 👍

dzhavat (approved):

libs/ui-watt/src/lib/components/input/timepicker/watt-timepicker.module.spec.ts
```diff
@@ -171,6 +173,15 @@ describe(WattTimepickerModule.name, () => {
       expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
     });
 
+    it('disables slider button', async () => {
+      await setup({ template, disabled: true });
+      const sliderToggle = screen.queryByRole('button') as HTMLButtonElement;
+
+      userEvent.click(sliderToggle);
+
+      expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
```
dzhavat, line 182: question: what do you think about using the `[.toBeDisabled`](https://github.com/testing-library/jest-dom#tobedisabled) matcher on the button in this case? The test wouldn't know what would have happen had the button been clicked.

ManBearTM: Good thinking, added an extra `expect` to assert for disabled state.

PR #671, opened 2022-06-27
fix(ui-watt): prefer watt-button

ManBearTM:
**IMPORTANT: DO NOT MERGE UNTIL APPROVED BY PRODUCT TEAMS** 

## Description
This PR aims to use `<watt-button>` in the timepicker, datepicker and modal component. To accomplish this, a few changes had to be made:

- Set the width of the icon button variant to be `44px`. This aligns with the design in Figma, so could be considered a bug fix.
- Use `:focus-visible` selector instead of `:focus` for all buttons. This means focus state is only visible to keyboard users.

These changes are potentially breaking (although only in a styling-sort-of-way), so it has to be tested and cleared with the other teams. There might also be accessibility concerns regarding `:focus-visible` that I'm not aware of, so that should also be looked into.


dzhavat:
praise: looks nice. Hopefully enough to start the process of adjusting the icon button to look better in such contexts.

issue: EO bump 😒

libs/ui-watt/src/lib/components/input/timepicker/watt-timepicker.module.spec.ts
```diff
@@ -177,7 +177,9 @@ describe(WattTimepickerModule.name, () => {
       await setup({ template, disabled: true });
       const sliderToggle = screen.queryByRole('button') as HTMLButtonElement;
 
-      userEvent.click(sliderToggle);
+      userEvent.click(sliderToggle, undefined, {
+        skipPointerEventsCheck: true,
+      });
```
dzhavat, line 182: praise: nice solution. Didn't know this was possible.

dzhavat (approved):

dzhavat (approved):

PR #676, opened 2022-06-29
refactored emit out of ngOnChanges for organization- and actor-master-data component + added UTs

michaelmajgaard:


libs/dh/market-participant/data-access-api/src/lib/dh-market-participant-edit-organization-data-access-api.store.ts
```diff
@@ -174,7 +164,21 @@ export class DhMarketParticipantEditOrganizationDataAccessApiStore extends Compo
       .pipe(
         tap((organization) =>
           this.patchState({
-            organization,
+            organization: organization,
+            changes: {
```
frontend-specialisten, line 168: Seems a bit odd that `changes` and `organization` are identical? When that's the case why even have the `changes` state?

Anyways this can also be simplified to this (as it seems all the props from organization is used): 
```suggestion
            changes: {...organization}
```

michaelmajgaard: Fixed

frontend-specialisten (approved):

Issue #408, opened 2022-03-16
Documentation for feature flags


dzhavat:
**AC**

- [ ] Document feature flag service
- [ ] Document feature flag directive
- [ ] Document feature flag configuration and environment overwrites
- [ ] Notify feature teams

mknic:
@dzhavat @ManBearTM @frontend-specialisten the wholesale domain would be the place we start to implement the feature flag mechanism and I think that experience together with them would provide good feedback to our documentation. Nevertheless, it's nice having this issue and I definitely agree with creating documentation before we announce the readiness of this feature.

PR #679, opened 2022-06-29
docs(dh): add docs for feature flags

frontend-specialisten:
## Description

Add docs for feature flags

## References

- #408


libs/dh/shared/feature-flags/README.md
```diff
@@ -1 +1,60 @@
...
+```ts
+'example-feature-flag': {
+    created: '01-01-2022',
+    disabledEnvironments: [DhAppEnvironment.prod],
+  }
+```
+
+The `created` property is danish locale `dd-mm-yyyy`. Feature flags use this property internally to check for old feature flags.
+
+The `disabledEnvironments` property is an array of the environments you wan't to disable the feature, for ex., `prod.`
```
dzhavat, line 26: suggestion:

```suggestion
The `disabledEnvironments` property is an array of the environments you want to disable a feature in, e.g. `prod`.
```

libs/dh/shared/feature-flags/README.md
```diff
@@ -1 +1,60 @@
...
+The feature flags require two properties, `created` and `disabledEnvironments`.
+
+**Example:**
+
+```ts
+'example-feature-flag': {
+    created: '01-01-2022',
+    disabledEnvironments: [DhAppEnvironment.prod],
+  }
+```
```
dzhavat, line 22: nitpick:suggestion: let's use the same feature flag name mentioned later in the doc.

```suggestion
```ts
'my-awesome-feature-flag': {
  created: '01-01-2022',
  disabledEnvironments: [DhAppEnvironment.prod],
}
```
```

**Review by ManBearTM — COMMENTED** [review]
Awesome! I just have one question 🙂

File: libs/dh/shared/feature-flags/README.md [hunk 3]
```diff
@@ -1 +1,59 @@
...
+
+The `created` property is danish locale `dd-mm-yyyy`. Feature flags use this property internally to check for old feature flags.
+
+The `disabledEnvironments` property is an array of the environments you want to disable a feature in, e.g. `prod`.
+
+## Using a feature flag
+
+Feature flags can be used by the `*dhFeatureflag` directive or the `DhFeatureFlagsService.`
+
+<mark>Notice: Unknown feature flags not added to the feature flag configuration are treated as enabled.</mark>
```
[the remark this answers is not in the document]
frontend-specialisten: @ManBearTM You're absolutely right unless you "cheat" with `as DhFeatureFlags`. This is ofc. not the intended usage but the behavior would be as mentioned in the doc. However, we could skip documenting this, as it's not intended usage what do you think?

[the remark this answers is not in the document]
frontend-specialisten: Actually there's also the use-case where no feature flags are added to the configuration. then `DhFeatureFlags` will get the type `never`.

libs/dh/shared/feature-flags/README.md
```diff
@@ -1 +1,59 @@
...
+
+```ts
+  constructor(private featureFlagsService: DhFeatureFlagsService) {}
+```
+
+2. Use the `isEnabled` method on the service:
+
+```ts
+this.featureFlagsService.isEnabled('my-awesome-feature-flag');
+```
```
ManBearTM, line 59: Perhaps add a short section about how to remove the flags again, something like this:
```md
## Remove a feature flag

When it is time to remove a feature flag, it is recommended to begin by removing the entry from
`dhFeatureFlagsConfig`. The compiler will then error in all the files using that feature flag, making
it easier to identify which files need to be cleaned up.
```

frontend-specialisten: Good point!

ManBearTM (approved):

dzhavat (approved):

Issue #495, opened 2022-04-27
Clean-up "testing" log


frontend-specialisten:
Currently, some tests are outputting a lot of noise in the log due to missing translations:

![image.png](https://images.zenhubusercontent.com/612cbcaa9fd6401e03dda309/9927376a-beb2-461b-9b66-62bac6c6ce5f)

Steps to reproduce:
- [ ] Run all tests (unit and integration tests)

PR #682, opened 2022-06-30
test(dh): get rid of translation logs in tests

dzhavat:
### DataHub
- get rid of translation logs in tests

## References

- #495


frontend-specialisten (approved):
Looks good, however I'm wondering why these tests aren't executed by the CI :/ 
image

Issue #652, opened 2022-06-21
Generator for different types of libraries (part 1)


mknic:
**Description**
Create a generator for different libraries to make it easy to set up a specific library
So that we simplify how to configure each type individually.

The libraries to support is listed here: https://github.com/Energinet-DataHub/greenforce-frontend/blob/main/README.md#library-types

**AC**
- [ ] Create a generator to support each type of library
- [ ] Document how to use the generator
- [ ] Should update `test-setup.ts` file to correctly setup TestBed specific to the library type
- [ ] Should update README.md file to remove auto-generated text

**Note**
- `assets` and `styles` types are not in scope

PR #708, opened 2022-07-06
feat: add feature library generator

dzhavat:
### Workspace
- add a generator that can be used for generating new feature libraries in DataHub

image

## References

- #652


frontend-specialisten (dismissed):
Question: Do we want a command for each type of library or should we consider just prompting for the type instead? If we have a command for each type you need to have an overview of our different types beforehand, where if we prompt for it, the user can decide to use the generator. What do you think, each has of course its pros and cons?   

Question: Should we consider prompting for `product` as well, maybe ask @MadsThy if Fenris could use this generator as well so we're consistent?

dzhavat:
> Do we want a command for each type of library or should we consider just prompting for the type instead?

I didn't want to generalize too much from the start because each library type has its own specifics. So we could've looked at generalization after having 2-3 generators. But I agree that it's probably better with one command and letting the "user" select a library type.

So maybe that's how we should start and incrementally add new library types. I'm thinking of adding a third option for a library type which will be like a dropdown with predefined values. For now we'll only have "feature" in the list.

frontend-specialisten (approved):

PR #723, opened 2022-07-12
refactor(ci): remove unused affected check

frontend-specialisten:
## Description

Removed unused `affected` job from `frontend-ci.yml`


.github/workflows/frontend-ci.yml
```diff
@@ -51,64 +51,9 @@ on:
   workflow_dispatch: {}
 
 jobs:
-  dh-affected:
```
dzhavat, line 54: issue: the `dh-affected` job is needed for the `e2e` and `test` jobs (found later in the file)

frontend-specialisten: Should be fixed now, I just forgot to remove the `needs`. Because the dh-affected was actually not used since it's just running the tests for the affected projects.

dzhavat (approved):

Issue #698, opened 2022-07-01
Configurable top header


dzhavat:
**Description**
Add a top header to the drawer component that can take different content (e.g. breadcrumbs).

**AC**
- [x] Should be positioned at the top of the drawer component
- [x] Should always include a close button
- [x] Should have a bottom border
- [x] Should be "sticky"
- [x] Showcase in Storybook
- [x] Long text should wrap on a new line

![image.png](https://images.zenhubusercontent.com/612cb8944df0eb85278b93db/66c16eb0-8e70-4fb0-8dbb-842cccc0d233)

**Tech notes**
- The close button should be in the upper-right corner
- The header should be "sticky" i.e. not scroll with the main content but remain in place

PR #732, opened 2022-07-14
feat(watt): add support for top bar in drawer component

dzhavat:
### Watt Design System
- add support for top bar in drawer component

### Energy Origin
- bump project version

## References

- #698

### DIFF
--- a/libs/ui-watt/src/lib/components/drawer/watt-drawer.component.ts
+++ b/libs/ui-watt/src/lib/components/drawer/watt-drawer.component.ts
@@ -23,7 +23,6 @@ import {
   Output,
   ViewChild,
   ViewContainerRef,
-  ViewEncapsulation,
 } from '@angular/core';
 
 import { WattDrawerContentDirective } from './watt-drawer-content.directive';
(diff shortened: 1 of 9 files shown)


libs/ui-watt/src/lib/components/drawer/watt-drawer.component.ts
```diff
@@ -38,33 +38,27 @@ export class WattDrawerComponent {
...
+  @ViewChild('topBarVcr', { read: ViewContainerRef, static: false })
+  private topBarVcr?: ViewContainerRef;
+
   /** @ignore */
   @ViewChild('contentVcr', { read: ViewContainerRef, static: false })
   private contentVcr?: ViewContainerRef;
 
   /** @ignore */
-  private _isOpened = false;
+  private areViewsCreated = false;
```
frontend-specialisten, line 58: Question: why do we need this, when we have the `opened` property?

dzhavat: Good question. We can only use `opened` to prevent the drawer from opening again. Will make the changes.

frontend-specialisten (approved):

Issue #700, opened 2022-07-01
Support for action buttons


dzhavat:
**Description**
The drawer component should support action buttons 

**AC**
- [ ] Should only support action buttons
- [ ] Should be positioned neat the top, bellow the top header
- [ ] Content should be aligned to the right
- [ ] Showcase in Storybook

![image.png](https://images.zenhubusercontent.com/612cb8944df0eb85278b93db/2ae7b103-3d76-4400-b2f3-05cd2b8d68cd)

**Tech notes**
- Buttons should not be hardcoded
- The element should be "sticky" i.e. not scroll with the main content but remain in place

PR #734, opened 2022-07-15
feat(ui-watt): add support for drawer action buttons

frontend-specialisten:
## Description

Add support for drawer action buttons

## References

- #700


libs/ui-watt/src/lib/components/drawer/watt-drawer.component.ts
```diff
@@ -84,25 +87,7 @@ export class WattDrawerComponent {
    * Closes the drawer
    */
   close() {
-    if (this.opened) {
```
dzhavat, line 87: suggestion: the `if` should be still here because pressing the escape button will still call the method even though the drawer itself is not open because the event is global on the drawer component. That component is still in the DOM.

dzhavat (approved):

PR #742, opened 2022-07-22
feat: Add feature flags to Wholesale UI.

defectiveAi:
(Link to issue)[https://app.zenhub.com/workspaces/mandalorian-5fd22a1a59e4740018f5ab5a/issues/energinet-datahub/opengeh-wholesale/128]


defectiveAi:
Awaiting team confirmation.

libs/dh/core/shell/src/lib/dh-primary-navigation.component.html
```diff
@@ -25,11 +25,13 @@
     <watt-nav-list-item link="/market-participant">{{
       transloco("marketParticipant")
     }}</watt-nav-list-item>
-    <watt-nav-list-item link="/wholesale/start-process">{{
-      transloco("wholesaleStartProcess")
-    }}</watt-nav-list-item>
-    <watt-nav-list-item link="/wholesale/overview">{{
-      transloco("wholesaleOverview")
-    }}</watt-nav-list-item>
+    <ng-container *dhFeatureFlag="'start_wholesale_process_feature_flag'">
```
dzhavat, line 28: praise: good job. Looks right 👍

libs/dh/shared/feature-flags/src/lib/feature-flags.ts
```diff
@@ -40,6 +40,11 @@ const makeFeatureFlags = satisfies<FeatureFlagConfig>();
  *   disabledEnvironments: [DhAppEnvironment.prod],
  * },
  */
-export const dhFeatureFlagsConfig = makeFeatureFlags({});
+export const dhFeatureFlagsConfig = makeFeatureFlags({
+  start_wholesale_process_feature_flag: {
+    created: '15-08-2022',
```
dzhavat, line 45: nitpick: there's a test in the pipeline that will fail if a feature flag is older than 14 days. There's also a task in our backlog about evaluating whether 14 days is enough and it should be changed to something else. So the "right" thing to do for now if your feature flag "expires" is to set a new `created` value.

dzhavat (approved):

PR #745, opened 2022-07-27
feat(ui-watt): only allow one drawer at a time

### LINKED ISSUE #727: Single drawer open at a time
**Description**
Only one drawer component must be opened at a time.

**AC**
- [ ] Only one drawer open at a time (a modal can be opened on top of the drawer)

**CONSIDER IF THIS SHOULD BE DONE AS THIS CAN BE FIXED BY JUST USING THE DRAWER IN A CERTAIN WAY**

ManBearTM:
## Description
Trying to take the simplest approach here, since I agree that this is normally a non-issue if using the drawer "correctly" (as in reusing the same drawer component and then just updating the content instead of opening a new drawer).

However, in certain cases reusing the same drawer component might not be that feasible (for instance if on the same page two separate components are using a drawer to render completely different data).

I considered various solutions such as using a [Portal](https://material.angular.io/cdk/portal/overview) along with a service, but they were all too complex solutions for something that might not even be an issue. So I settled on just storing the current drawer instance in a static property, making it shared between multiple drawer instances. Then if you open a new drawer, the current drawer (if any) is automatically closed.

## References

Closes #727

### DIFF
--- a/libs/ui-watt/src/lib/components/drawer/+storybook/watt-drawer.stories.ts
+++ b/libs/ui-watt/src/lib/components/drawer/+storybook/watt-drawer.stories.ts
@@ -27,25 +27,18 @@ export default {
   title: 'Components/Drawer',
   component: WattDrawerComponent,
   argTypes: {
+    size: { control: false },
+    opened: { control: false },
     closed: {
-      table: {
-        category: 'Outputs',
-      },
+      table: { category: 'Outputs' },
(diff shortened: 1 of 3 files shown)


dzhavat:
praise: looks like a nice solution. Nice work 👍

nitpick: strange why it's not flagged by CI but EO probably must be bumped

Issue #761, opened 2022-08-09
Integrate Application Insights in Angular


dzhavat:
**AC**
- [x] Configure Application Insights
- [x] Create an Application Insights service
- [x] Expose methods for tracking custom events
- [x] Verify that it works by sending a custom event

Libraries
* https://github.com/microsoft/ApplicationInsights-JS
* https://github.com/microsoft/applicationinsights-angularplugin-js

Link for inspiration https://timdeschryver.dev/blog/configuring-azure-application-insights-in-an-angular-application

PR #775, opened 2022-08-10
feat(dh): integrate Application Insights

dzhavat:
### DataHub
- integrate Application Insights
- expose service that can be used to track custom events
- increase initial budget to 1.75 Mb

### Workspace
- allow `type:environments` to be imported in `type:util`

### Energy Origin
- bump EO project version

## References

- #761


package.json
```diff
@@ -63,6 +63,8 @@
     "@angular/router": "14.0.2",
     "@azure/msal-angular": "2.3.3",
     "@azure/msal-browser": "2.27.0",
+    "@microsoft/applicationinsights-angularplugin-js": "^3.0.0",
```
frontend-specialisten, line 66: Should we use exact version to be consistent with our other deps?
```suggestion
    "@microsoft/applicationinsights-angularplugin-js": "3.0.0",
```

package.json
```diff
@@ -63,6 +63,8 @@
...
+    "@microsoft/applicationinsights-web": "^2.8.6",
```
frontend-specialisten, line 67: Should we use exact version to be consistent with our other deps?
```suggestion
    "@microsoft/applicationinsights-web": "2.8.6",
```

libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
```diff
@@ -0,0 +1,67 @@
...
+   */
+  init(): void {
+    if (this.appInsights.appInsights.isInitialized()) {
+      return;
+    }
+
+    this.appInsights.loadAppInsights();
+  }
+
+  trackEvent(name: string): void {
```
frontend-specialisten, line 60: Consideration: We should properly consider which methods we want to proxy? You can find the list [here](https://github.com/microsoft/ApplicationInsights-JS#sending-telemetry-to-the-azure-portal). Ex. `trackException`and maybe `trackPageView` could be used for overlays?

frontend-specialisten (approved):

libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
dzhavat: Nice suggestion. We can also expose more methods if there's a specific need/requirement at a later point. Will add `trackExceptionand` and `trackPageView` for now.

libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
```diff
@@ -0,0 +1,95 @@
...
+  dhAppEnvironmentToken,
+} from '@energinet-datahub/dh/shared/environments';
+
+@Injectable({ providedIn: 'root' })
+export class DhApplicationInsights {
+  private angularPlugin = new AngularPlugin();
+  private appInsights = new ApplicationInsights({
+    config: {
+      instrumentationKey:
+        this.dhAppConfig.applicationInsights.instrumentationKey,
```
frontend-specialisten, line 37: Question: should we change this to use a `connectionString` instead or create another issue for that?

libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
```diff
@@ -0,0 +1,95 @@
...
+   */
+  trackTrace(message: string): void {
+    this.appInsights.trackTrace({ message });
+  }
+
+  /**
+   * Logs that a page, or similar container was displayed to the user.
+   * @param pageView
+   */
+  trackPageView(pageView: IPageViewTelemetry): void {
```
frontend-specialisten, line 84: Consideration: `the interface of `IPageViewTelemetry` seems only to have optional properties. For the name, the `title` will be used by default. I guess this will not be ideal if the primary usage of this method will be to track when overlays are displayed.

libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.appInsights.trackPageView(pageView);
+  }
+
+  /**
+   * Log an exception that you have caught.
+   * @param exception
+   */
+  trackException(exception: IExceptionTelemetry): void {
```
frontend-specialisten, line 92: Consideration: This interface could properly also be simplified in first version, maybe only keep `exception` as argument as example?

libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
dzhavat: If we can get the `connectionString` the same way we got the `instrumentationKey` in the pipeline, then let's change it. We have time for it in the current task.

libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
dzhavat: Fair point. Let's start with one param and expand if needed.

frontend-specialisten (dismissed):
Praise: Nice solution with the app initializer

frontend-specialisten (approved):

PR #776, opened 2022-08-12
Existing market roles can only be edited/removed if actor status is 'New'

michaelmajgaard:


libs/dh/market-participant/feature-edit-actor/src/lib/market-participant-edit-actor/market-roles/dh-market-participant-actor-market-roles.component.ts
```diff
@@ -172,6 +175,8 @@ export class DhMarketParticipantActorMarketRolesComponent implements OnChanges {
   };
 
   readonly onRowDelete = (row: EditableMarketRoleRow) => {
+    if (this.isReadonly(row)) return;
```
michaelmajgaard, line 178: @dzhavat This is a hack, but `click` on the button fires even though `disabled = true`. Am I doing something wrong?

dzhavat: @michaelmajgaard I think that's a bug. Will investigate it further today. Let's go with your solution for now and remove the line when the bug is fixed.

dzhavat (approved):

libs/dh/market-participant/feature-edit-actor/src/lib/market-participant-edit-actor/market-roles/dh-market-participant-actor-market-roles.component.ts
dzhavat: Ugh, it's a 10+ years old bug in WebKit. 😭 https://bugs.webkit.org/show_bug.cgi?id=89041
I don't think it will ever be fixed, so we gotta find our own solution. 🤦‍♂️

michaelmajgaard: Yikes 😆

PR #785, opened 2022-08-15
 fix(watt): prevent emitting click events when disabled button is clicked

dzhavat:
### Watt

- prevent click events from emitting when disabled button is clicked

### DataHub
- remove unnecessary readonly check in method
- fix failing test

## References

- N/A


frontend-specialisten (changes requested):
Looks good, could we add or adjust the spec of the disabled button to make sure we won't end up with the same issue in the future?

michaelmajgaard (approved):

frontend-specialisten (approved):

PR #792, opened 2022-08-17
fix: passing continuation token when navigating to next page in message archive search

michaelmajgaard:


libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.html
```diff
@@ -154,8 +154,9 @@ <h1>{{ transloco("title") }}</h1>
           >{{ transloco("page") }} {{ this.pageNumber }}</watt-label
         >
         <watt-button
+          *rxLet="continuationToken$ as continuationToken"
           [disabled]="(continuationToken$ | async) === null"
```
dzhavat, line 158:
```suggestion
          *rxLet="continuationToken$ as continuationToken"
          [disabled]="continuationToken === null"
```

dzhavat (approved):

PR #797, opened 2022-08-18
Datarange picker for consumption and production

MadsThy:
## Description

Date range picker is now being being soft-pushed to consumption and production pages.

They are going to be in production, but hidden behind the feature flag 'daterange' for now.

They can be seen at https://demo.energioprindelse.dk/consumption?enableFeature=daterange

image


C-Christiansen (approved):
LGTM

Issue #653, opened 2022-06-21
Generator to add a completely new feature


mknic:
**Description**
A generator that scaffolds a completely new feature for a domain.

**AC**
- [ ] Should generate a `feature-` library
- [ ] Should generate a `routing` library
- [ ] Should generate a `data-access-api` library
- [ ] Should generate a `shell` library
- [ ] Should update `test-setup.ts` file in each library
- [ ] Should update `README` file in each library

PR #805, opened 2022-08-19
feat(dh): domain generator

ManBearTM:
## Todo
- [x] Add section to README once finalized

## Description
This PR adds a new domain generator, that simply generates a new {domain}
with the following libraries:

- data-access-api
- feature-{name}
- routing
- shell

I would like to discuss the following:

- Is the name "dh-domain-generator" indicative of its purpose?
- Would we like to have some additional options to the generator,
  such as "generate domain library as well" or "skip routing library"?

## References

- #653


tools/generators/dh-domain-generator/schema.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "cli": "nx",
+  "$id": "dh-domain-generator",
```
dzhavat, line 4: suggestion: I think the name fits well with one minor adjustment - `dh-new-domain-generator` to be more indicative. What do you think?

tools/generators/dh-domain-generator/schema.json
```diff
@@ -0,0 +1,25 @@
...
+      "$default": {
+        "$source": "argv",
+        "index": 0
+      },
+      "x-prompt": "What is the name of the domain?"
+    },
+    "name": {
+      "type": "string",
+      "description": "What is the name of the feature?",
+      "x-prompt": "What is the name of the feature?"
```
dzhavat, line 20: suggestion: Maybe too explicit but to explain a bit why a feature name is needed. Open for other suggestions too or keeping the current value, if you think it's fine enough.

```suggestion
      "description": "What is the name of the first feature under the new domain?",
      "x-prompt": "What is the name of the first feature under the new domain?"
```

tools/generators/dh-domain-generator/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  await dhLibraryGenerator(tree, {
+    domain: schema.domain,
+    libraryType: LibraryType.dataAccess,
+    name: 'api',
+  });
+
+  await dhLibraryGenerator(tree, {
+    domain: schema.domain,
+    libraryType: LibraryType.shell,
+  });
```
dzhavat, line 46: praise: nice that we can do that 💪

tools/generators/dh-domain-generator/schema.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "cli": "nx",
+  "$id": "dh-domain-generator",
+  "description": "Generate an entire domain that's going to be used in DataHub.",
```
dzhavat, line 5:
```suggestion
  "description": "Generate an entire new domain that's going to be used in DataHub.",
```

dzhavat:
As for supporting additional options, I think it's fine as it is right now. A `domain` library would've been a good idea if we wrote DTOs and interfaces ourselves but that's done automatically. Skipping the routing library could be an option but I think it's better to have it from the start than to generate it at a later point.

tools/generators/dh-domain-generator/schema.json
ManBearTM: I like it!

dzhavat (approved):

Issue #593, opened 2022-06-03
Tooltip/popover


mknic:
**Description**
Implement a tooltip/popover similar to what is in the design system on the charge details page.

**Acceptance Criteria**
- [ ] Implement Watt version of angular materials tooltip
- [ ] Default positioning should be above 
- [ ] Positioning should be configurable 
- [ ] Add example / showcases to storybook
- [ ] Should be aligned with figma
- [ ] Inform feature teams to refactor existing usage of material tooltip when component is available in Watt

**Clarifications**
how much should the show delay be? @renehossle 

**Tech notes:**
https://www.figma.com/file/fwEDQkgEcpwfUojXl0EIC5/Watt-Design-System?node-id=244%3A51
https://material.angular.io/components/tooltip/overview

PR #830, opened 2022-08-25
feat(ui-watt): tooltip

ManBearTM:
## Description
This PR creates a new `[wattTooltip]` directive that extends `[matTooltip]`. The styles are applied globally, so they affect both directives, meaning the `[wattTooltip]` is basically just an alias for `[matTooltip]`.

**Disclaimer: I am not at all, whatsoever, responsible for any future breakage that will most assuredly happen when the Material team decides to change their internal implementation in even the slightest of ways.**

## References

- #593


libs/ui-watt/src/lib/components/tooltip/watt-tooltip.directive.ts
```diff
@@ -0,0 +1,103 @@
...
+  TooltipPosition,
+} from '@angular/material/tooltip';
+
+export type wattTooltipPosition = TooltipPosition;
+
+@Directive({
+  selector: '[wattTooltip]',
+  exportAs: 'wattTooltip',
+  host: {
+    class: 'mat-mdc-tooltip-trigger',
```
dzhavat, line 50: question: is this in order to be forwards compatible 😁

libs/ui-watt/src/lib/foundations/_tooltip.scss
```diff
@@ -0,0 +1,67 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@use "./typography/typography.import" as typography;
+@use "./spacing/spacing.import" as space;
+
+$arrow-size: 6px; // magic number by designer
```
dzhavat, line 20: 👍

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.directive.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    Story({}, {} as never)
+  );
+
+  return render(component, { imports: [ngModule] });
+}
+
+describe(WattTooltipDirective.name, () => {
+  it('always displays accessible tooltip', async () => {
+    await setup();
+    expect(screen.getByRole('tooltip', { hidden: true })).toBeInTheDocument();
```
dzhavat, line 40: praise: nice solution including the hidden element. I need something similar in my task so I'm "stealing" the idea :)

dzhavat (approved):

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.directive.ts
ManBearTM: Haha, I guess 😄 I basically just copied the entire decorator from the internal implementation. Do you think I should remove it?

dzhavat: If it should be here, then it must probably be just `mat-tooltip-trigger` because that's used in the [current version](https://github.com/angular/components/blob/main/src/material/legacy-tooltip/tooltip.ts#L48) (renamed to "legacy" in the upcoming v15) of the tooltip component. The `mat-mdc-tootip-trigger` class is used in the "new" version (based on MDC). 

But I think we can remove it and the right class will still be applied by the parent class (hope so).

dzhavat (approved):

Issue #807, opened 2022-08-19
Expandle sidenav items


frontend-specialisten:
**Description**
Implement expandable menu items. 

![image.png](https://images.zenhubusercontent.com/612cbcaa9fd6401e03dda309/7328d8f4-e306-4dcf-958a-387b5b3f9eb9)

**Acceptance Criteria**
- [x] Support expandable menu items in page template 
- [x] Expanding menu items should have wild transitions
- [x] Make sure active page works with child item (parent should be expanded)
- [x] Add example / showcases to storybook
- [x] Should be aligned with Figma
- [x] Refactor existing "Settlement" links to be expandable
- [x] Do not show border on menu items on hover

**Clarifications**
Should parent menu items be navigated (and/or highlighted)? @renehossle 

**Tech notes:**
https://www.figma.com/file/fwEDQkgEcpwfUojXl0EIC5/Watt-Design-System?node-id=244%3A51
https://material.angular.io/components/tooltip/overview

PR #831, opened 2022-08-26
feat(ui-watt): add support for expandable sidebar navigation

dzhavat:
### Watt
- add support for expandable sidebar navigation
- add expandable menu example in Storybook showcase

### Energy Origin
- bump project version
- increase build budgets

## References

- #807


libs/ui-watt/src/lib/components/shell/nav-list/watt-expand-on-active-link.directive.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      from(links$)
+        .pipe(
+          mergeAll(),
+          filter((isActive) => isActive)
+        )
+        .subscribe(() => {
+          this.panel.open();
+        });
+    }
```
ManBearTM, line 48: nitpick: perhaps combine the `map` and `mergeAll` into the `mergeMap` operator?
```ts
from(navListItems)
  .pipe(
    mergeMap(item => item.isActive$),
    filter((isActive) => isActive),
  )
```

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.ts
```diff
@@ -15,28 +15,82 @@
...
+          <mat-panel-title class="watt-text-m">{{ title }}</mat-panel-title>
+        </mat-expansion-panel-header>
+        <ng-container *ngTemplateOutlet="navListTemplate"></ng-container>
+      </mat-expansion-panel>
+    </ng-container>
+
+    <ng-template #navListTemplate>
+      <mat-nav-list><ng-content></ng-content></mat-nav-list>
+    </ng-template>
+  `,
```
ManBearTM, line 61: Question: Is it time to move the template to a separate file? What is our preference on this?

ManBearTM (approved):
Praise: Looks good and works like a charm!

libs/ui-watt/src/lib/components/shell/nav-list/watt-expand-on-active-link.directive.ts
dzhavat: Magic 🔮 Nice snippet 🚀

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.ts
dzhavat: I don't think there's a clear guideline for when an inline template should be moved to a separate file. Probably a good idea in this case. Have done it 👍

ManBearTM (dismissed):

libs/ui-watt/src/lib/components/shell/nav-list/watt-expand-on-active-link.directive.ts
```diff
@@ -0,0 +1,48 @@
...
+    const navListItems = this.wattNavListItemComponents?.toArray();
+
+    if (navListItems) {
+      from(navListItems)
+        .pipe(
+          mergeMap((item) => item.isActive$),
+          filter((isActive) => isActive)
+        )
+        .subscribe(() => {
+          this.panel.open();
```
MadsThy, line 44: Will this subscribe be unsubscribed again on destroy of item that uses this directive? :)

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.html
```diff
@@ -0,0 +1,32 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<ng-container *ngIf="expandable; else navListTemplate">
+  <mat-expansion-panel
+    wattExpandOnActiveLink
+    [wattNavListItemComponents]="navListItemComponents"
+    class="mat-elevation-z0"
```
MadsThy, line 21: Where is this class used? :)

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
```diff
@@ -203,4 +205,106 @@ describe(WattNavListModule.name, () => {
...
+        <watt-nav-list [title]="'${title}'" [expandable]="true">
+          <watt-nav-list-item link="/sub-page">Sub page</watt-nav-list-item>
+        </watt-nav-list>
+      `,
+        {
+          imports: [WattNavListModule],
+        }
+      );
+
+      expect(view.queryByText(title)).not.toBeNull();
```
MadsThy, line 224: Its nice to know that its not null, for sure, but can i suggest that you make an expect that checks that the title is actually set to what we input? :)

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
```diff
@@ -203,4 +205,106 @@ describe(WattNavListModule.name, () => {
...
+        <watt-nav-list [title]="'Title'" [expandable]="true">
+          <watt-nav-list-item link="/sub-page">${text}</watt-nav-list-item>
+        </watt-nav-list>
+      `,
+        {
+          imports: [WattNavListModule],
+        }
+      );
+
+      expect(view.queryByText(text)).not.toBeNull();
```
MadsThy, line 241: Its nice to know that its not null, for sure, but can i suggest that you make an expect that checks that the list item is actually set to what we input? :)

MadsThy (changes requested):
Left a few comments :)

libs/ui-watt/src/lib/components/shell/nav-list/watt-expand-on-active-link.directive.ts
dzhavat: No. There's no need to unsubscribe because the menu items are always there and the component is never destroyed. So ngOnDestroy will never execute.

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.html
dzhavat: This is part of Material. A so called "elevation helper". More info here https://material.angular.io/guide/elevation

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
dzhavat: Isn't that implicitly true by using `view.queryByText(title)`? If the title is not found in the DOM, the query will not find it.

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
dzhavat: Isn't that implicitly true by using `view.queryByText(title)`? If the title is not in the DOM, the query will not find it.

MadsThy: Isn't the convention that we want to make sure it shows the text that we input? :)

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
MadsThy: https://github.com/Energinet-DataHub/greenforce-frontend/pull/831/files#r959194250

libs/ui-watt/src/lib/components/shell/nav-list/watt-nav-list.component.spec.ts
dzhavat: Yes, it is for most tests. In this case the focus is to test that we can successfully project content in a component. Not so much about what the projected content is. The content is provided by other developers when they use the component. Other tests in the suite assert that `watt-nav-list-item` is outputting a link, finding it by text value, that it has the right class when active, etc.

MadsThy (approved):

PR #833, opened 2022-08-26
Bug/fix dialog picker

MadsThy:
## Description

- All dates picked, should return time 00:00:00 for the day picked
-  The end date should be saved as day+1 so it will always be bigger than start date
-  Picking for 1 day only, should work.
-  Even if the end date is saved as day+1, the calendar should still only show the day intended.
-  If a new date is picked while components are loading, it should cancel the load and do a new API call
-  When selecting 2020, it selects 1/1-2020 - 1/1-2021, it should be 1/1-2020 - 31/12-2020
-  When clicking a predefined range, the calendar should update instantly and move it to display the start date
-  The start or end date should be highlighted depending on which is being selected
-  When clicking the start date, it should be highlighted and it should be be possible to select the start date, the same with the end date
-  When selecting 7 days, it selects 8 days :-)


PeterAGY (approved):
Looks good to me :)

PR #834, opened 2022-08-26
Fix calender pick on same date

MadsThy:
## Description

Fixed a bug where we couldn't pick the same date for start and end


PeterAGY (approved):
Approved :)

Issue #820, opened 2022-08-23
Bug: Pull-requests only related to the BFF is blocked for merging


frontend-specialisten:
`Production Dependencies License Check` is blocking pull-request when changes only affects the BFF. 

**Acceptance criteria:**
- [ ] `Production Dependencies License Check` should still be required but not blocking Pull-requests only related to the BFF

**Technical notes:**

1. Move `Production Dependencies License Check` to it's own workflow with a filter only triggering when changes are made in `package.json` or `yarn.lock`.
2. Create workflow with same name ignoring changes to `package.json` or `yarn.lock` and returning success, like described [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks)

PR #870, opened 2022-09-07
chore: move production deps license check to it's own workflow

frontend-specialisten:
## Description

- Move `Production Dependencies License Check` to it's own workflow
- Check now only runs when changes are made in `package.json` and `yarn.lock` with custom action `dependencies-has-changed`
- `Production Dependencies License Check` has now been cleaned-up and is now more performant. 

## References

- #820


MadsThy (approved):
LGTM

Issue #816, opened 2022-08-22
Move title into top bar


ManBearTM:
**Description**
We need to move the title (not to be confused with `<title>`) from the page into the top bar on DataHub (and EO).

**AC:**

- [x] Remove logo from top bar on small screens
- [x] Use top bar for title on DataHub
- [x] Height of topbar should be 64px (according to Figma)
- [x] Height of language selector buttons in topbar should be 64px
- [ ] ~~Coordinate with Fenris for EO~~ [not relevant]
- [ ] ~~Add title to showcase in Storybook~~ [not relevant]

**Definition of Ready:**

- [x] The issue is correctly estimated
- [x] The issue is adequately described
- [x] Possible dependencies are defined and aligned
- [x] We have the neccesary skills to complete this issue
- [x] The issue can be completed withing 1 iteration
- [x] The issue has acceptance criteria defined
- [x] The issue has adequate Definition Of Done described

**Definition of Done:**

- [ ] Acceptance Criteria have been met
- [ ] The product has been demo’ed for relevant stakeholders
- [ ] Dependencies are handled
- [ ] The work has been documented
- [ ] The issue has been handed over and reviewed
- [ ] The PO has accepted the product

PR #871, opened 2022-09-07
feat(dh): add support for top bar title

dzhavat:
### DataHub
- generate `data-access-top-bar` library
- add support for top bar title by adding translation key to route config

## References

- #816


frontend-specialisten (approved):
LGTM, just have a few considerations: 
1. The flexibility of just having a translation key and not being able to have some dynamic maybe an issue? ex. a details page if the title should include something from an entity (however this could be handled later on if this is not a requirement atm.)
2. It feels a bit weird that the library is of type `util` when the only thing it's exporting is a component store. Shouldn't this  be a `data-access` library instead?

frontend-specialisten (approved):

Issue #294, opened 2022-01-18
Add link icon


MadsBloendAndersen:
### description
link icon is build into the design system, for linking to messages

![image](https://user-images.githubusercontent.com/72008712/149949141-353c1591-31b4-4afa-a55a-4ae241bec01c.png)

### Acceptance criteria
- [ ] link icon is added to design system with documentation and showcase

PR #873, opened 2022-09-07
feat: 234 add date range to start process

johevemi:
## Description

Wholesale backend counterpart https://github.com/Energinet-DataHub/opengeh-wholesale/pull/287

image

## References

- [#234](https://app.zenhub.com/workspaces/mandalorian-5fd22a1a59e4740018f5ab5a/issues/energinet-datahub/opengeh-wholesale/234) [#294](https://app.zenhub.com/workspaces/mandalorian-5fd22a1a59e4740018f5ab5a/issues/energinet-datahub/opengeh-wholesale/294)


libs/dh/wholesale/feature-start/src/lib/dh-wholesale-start.component.ts
```diff
@@ -14,29 +14,71 @@
...
 
+  optionsGridAreas: WattDropdownOption[] = Object.keys({
+    805: '805',
+    806: '806',
+  }).map((key) => ({
+    displayValue: key,
+    value: key,
+  }));
+
+  formControlGridArea = new FormControl<string[] | null>(null);
```
dzhavat, line 51: suggestion: instead of adding the `required` on the `watt-dropdown` component in the HTML template, you can set the FormControl as required. This is recommended when using Reactive driven forms.

```suggestion
  formControlGridArea = new FormControl<string[] | null>(null, Validators.required);
```

Note: `Validators` is imported from `@angular/forms`.

johevemi: I have implemented the suggested change

libs/dh/wholesale/feature-start/src/lib/dh-wholesale-start.component.html
```diff
@@ -16,8 +16,32 @@
...
+        <watt-error *ngIf="formControlRange?.errors?.rangeRequired">
+          {{ transloco("startProcess.dateRangeRequired") }}
+        </watt-error>
+      </watt-form-field>
+    </div>
+    <watt-button
+      (click)="createBatch()"
+      [disabled]="formControlRange.invalid || formControlGridArea.invalid"
       >{{ transloco("startProcess.startLabel") }}
     </watt-button>
```
dzhavat, line 46: question: the user can still click the button if a request is going. This will cancel the current request and start a new one. Should the button be disabled? Another thing we can do is let the current request complete and not start a new one even if button is clicked several times. This will be quite easy actually. Just replace `switchMap` with `exhaustMap` in the store. 

question: should something happen after the request is done? Like a notification or

johevemi: At this time it is ok that the start button can be pushed multiple times, at a later date we will address this. The same goes for notification when the request is done

johevemi: Although the one about switchMap, I will change it to exhaustMap, because it makes sense that the first click has priority

dzhavat (approved):

PR #879, opened 2022-09-08
fix(ui-watt): esc no longer closes both modal and drawer"

ManBearTM:
## Description
The global keydown handler for the drawer was being triggered by pressing ESC within a modal, meaning if you were to open a modal within a drawer and then press ESC, both modal and drawer would close. This aims to solve that by making the keydown event scoped to the host element. For that to work though, the drawer must have focus, which is why I had to add a focus trap.


frontend-specialisten (changes requested):
Looks good to me! however it seems a spec is broken, but besides that, it works perfectly :)

Issue #845, opened 2022-08-30
[Bug] Sticky table header overlaps drawer component


dzhavat:
**Description:**
When a table has a sticky header, it overlaps the drawer component. 

![MicrosoftTeams-image.png](https://images.zenhubusercontent.com/612cb8944df0eb85278b93db/e906b9b4-fc38-438b-af81-02d97092dcff)

**AC:**

- [ ] The drawer component should open above the page content

**Definition of Ready:**

- [ ] The issue is correctly estimated
- [ ] The issue is adequately described
- [ ] Possible dependencies are defined and aligned
- [ ] We have the necessary skills to complete this issue
- [ ] The issue can be completed withing 1 iteration
- [ ] The issue has acceptance criteria defined
- [ ] The issue has adequate Definition Of Done described

**Definition of Done:**

- [ ] Acceptance Criteria have been met
- [ ] The product has been demo’ed for relevant stakeholders
- [ ] Dependencies are handled
- [ ] The work has been documented
- [ ] The issue has been handed over and reviewed
- [ ] The PO has accepted the product

PR #886, opened 2022-09-09
fix(ui-watt): increase drawer z-index

ManBearTM:
## Description
Very simple PR, just increases the z-index of the drawer component to ensure it stays above tables with sticky headers (while still below modals).

## References

- #845

### DIFF
--- a/libs/ui-watt/src/lib/components/drawer/watt-drawer.component.scss
+++ b/libs/ui-watt/src/lib/components/drawer/watt-drawer.component.scss
@@ -25,6 +25,12 @@
   right: 0;
   top: 0;
   width: auto;
+
+  // Most material components does not have a z-index exceeding 100,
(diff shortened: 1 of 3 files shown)


frontend-specialisten (approved):

libs/ui-watt/src/lib/components/drawer/watt-drawer.component.scss
```diff
@@ -25,6 +25,12 @@
   right: 0;
   top: 0;
   width: auto;
+
+  // Most material components does not have a z-index exceeding 100,
+  // except for modals which has a value of 1000. Since the drawer
+  // needs to stay on top of all content except modals, the value
+  // of 500 was chosen here which hopefully will remain stable.
+  z-index: 500;
```
dzhavat, line 33: praise: nice findings 👍 The safe middle :)

PR #911, opened 2022-09-13
Feat/autogenerate deps licenses markdown (WIP)

frontend-specialisten:
## References

- #0000


PRODUCTION_DEPENDENCIES.md
```diff
@@ -0,0 +1,37 @@
+# Production Dependencies
+
+  | Name | Version | License |
+  | ---- | ------- | ------- |
+  | [angular-slider/ngx-slider](https://github.com/angular-slider/ngx-slider) | 2.0.4 | [MIT](/home/runner/work/greenforce-frontend/greenforce-frontend/node_modules/@angular-slider/ngx-slider/LICENSE) |
```
dzhavat, line 5: issue: don't think the license path would work locally (i.e. `[MIT](/home/runner/work/...`). Would it be possible to not include that path?

.github/actions/update-production-dependencies-list/action.yml
```diff
@@ -0,0 +1,9 @@
+name: 'Update PRODUCTION_DEPENDENCIES.md'
+description: 'Update list in PRODUCTION_DEPENDENCIES.md'
+inputs:
+  dependencies-from-license-checker:
+    description: 'Output from license-checker'
+    required: true
+runs:
+  using: 'node16'
```
dzhavat, line 8: question: can we use the `node-version-file` as we do in other pipelines?

PRODUCTION_DEPENDENCIES.md
frontend-specialisten: Link has been removed since I couldn't find a way to link to the license at github as the link to the license file requires a branch and we can't know if the dependency is using main, master or whatever.

.github/actions/update-production-dependencies-list/action.yml
frontend-specialisten: Have changed the action to be `composite` like the other actions now. For javascript actions you can't specify any node-version-file.

dzhavat (approved):

PR #931, opened 2022-09-19
fix: remove plugin:ngrx/component-store-strict from .eslintrc.json

frontend-specialisten:
## Description
replace `plugin:ngrx/component-store-strict` with `plugin:@ngrx/component-store-strict` in `.eslintrc.json`, to fix broken build on main.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -197,6 +197,10 @@
       "extends": ["plugin:@nrwl/nx/typescript"]
     },
     {
+      "files": ["*.ts"],
+      "extends": ["plugin:@ngrx/component-store-strict"]
+    },
+    {
       "files": ["*.js", "*.ts"],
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -196,10 +196,6 @@
       "files": ["*.ts", "*.tsx"],
       "extends": ["plugin:@nrwl/nx/typescript"]
     },
-    {
-      "files": ["*.ts"],
-      "extends": ["plugin:ngrx/component-store-strict"]
```
dzhavat, line 201: suggestion: looks like the `ngrx/component-store-strict` can be replaced with `@ngrx/component-store-strict` and things will just work ™ as before :) Can we try it?

frontend-specialisten: It's seems that it was that easy, good spotted :)

dzhavat (approved):

PR #933, opened 2022-09-19
feat: 1641 add charge prices tab

ShaBloze:
## Description

Add "Prices" tab in the sidemenu which navigates to an empty page.
This will be feature flagged.

## References

- #1641
https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/energinet-datahub/geh-charges/1641


libs/dh/charges/feature-prices/src/lib/dh-charges-prices.component.ts
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Component, OnInit, NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+@Component({
+  selector: 'dh-dh-charges-prices',
```
HenrikSommer, line 21: It the double prefix "dh-dh-" here on purpose?

HenrikSommer (commented):
LGTM, be aware of the two code smells identified by SonarCloud

libs/dh/charges/feature-prices/src/lib/dh-charges-prices.component.ts
ShaBloze: That is a mistake. I removed the double prefix.

ManBearTM (changes requested):
Code looks good, but you need to fix the linting errors. Consider installing the "eslint" extension in VS code if you haven't done so already (also check out the list of recommended extensions in VS code, just do `ctrl + shift + p` and type "Show Recommended Extensions").

ManBearTM (approved):

PR #939, opened 2022-09-20
feat: removed actor status deleted

michaelmajgaard:


libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -317,7 +317,9 @@
             "Inactive": "Inaktiv",
             "Passive": "Passiv",
             "Deleted": "Slettet"
-          }
+          },
+          "statusChangeWarningActionLabel": "Ja",
```
dzhavat, line 321: suggestion: there are "Yes"/"No" translations at the top of the file. If we know that `statusChangeWarningActionLabel` is going to always be "Yes", then we can use the global translation.

libs/dh/market-participant/feature-edit-actor/src/lib/market-participant-edit-actor/dh-market-participant-edit-actor.component.ts
```diff
@@ -95,8 +103,41 @@ export class DhMarketParticipantEditActorComponent {
...
+              'marketParticipant.actor.create.masterData.statuses.' +
+                initialStatus
+            ),
+          }
+        ),
+        action: (ref: WattToastRef) => {
+          this.store.save(this.backToOverview);
+          ref.dismiss();
+        },
+      });
```
dzhavat, line 136: suggestion: it works fine with a toast but the [modal](https://main--61765fc47451ff003afe62ff.chromatic.com/?path=/story/components-modal--small) is probably a better option here. Also because the modal wont auto close after some time. The `small` size should be fine.

libs/dh/market-participant/feature-edit-actor/src/lib/market-participant-edit-actor/dh-market-participant-edit-actor.component.html
```diff
@@ -67,13 +67,40 @@ <h1 *ngIf="!isEditing">{{ t("createHeader") }}</h1>
...
+                initialStatus: t("masterData.statuses." + status)
+              })
+            }}
+          </p>
+          <watt-modal-actions>
+            <ng-container *transloco="let t">
+              <watt-button (click)="save()">{{ t("yes") }}</watt-button>
+              <watt-button variant="secondary" (click)="closeModal()">{{
+                t("no")
+              }}</watt-button>
```
dzhavat, line 99: suggestion: another way of doing it which utilizes the `watt-modal` APIs is to use its `close` method. So for the "Yes" button we'd use `(click)="confirmationModal.close(true)"` and for the "No" button we'd use `(click)="confirmationModal.close(false)"`.

Then on the `<watt-modal ..>` itself we'd add `(closed)="onCloseModal($event)"` attribute, which calls the `onCloseModal` in the `dh-market-participant-edit-actor.component.ts` with a param (could be `isAccepted`).

In the component we'd then have:

```ts
onCloseModal(isAccepted: boolean): void {
    if (isAccepted) {
      this.save();
    }
  }
```

libs/dh/market-participant/feature-edit-actor/src/lib/market-participant-edit-actor/dh-market-participant-edit-actor.component.html
```diff
@@ -67,13 +67,40 @@ <h1 *ngIf="!isEditing">{{ t("createHeader") }}</h1>
...
+          <watt-button variant="primary" (click)="onSaved(status, changes)">{{
+            t("saveLabel")
+          }}</watt-button>
+        </div>
+        <watt-modal
+          #confirmationModal
+          [size]="'small'"
+          [title]="t('masterData.statusChangeTitle')"
+          [disableClose]="true"
+        >
```
dzhavat, line 85: praise: it looks a lot better with a modal 🌟

dzhavat (dismissed):

dzhavat (approved):

PR #952, opened 2022-09-23
feat(dh): adjust icons on charges result

ManBearTM:
## Description

Adjust the icons used in the charges result component to match the design.

## References

- #1655


ShaBloze (approved):
LGTM

PR #969, opened 2022-09-28
fix(watt): add startWith for date-range combinelatest

frontend-specialisten:
## Description

If date range had an initial value and ex. only the end date is changed the model won't be correct as `combineLatest` will never be triggered. This is fixed by adding a `startWith` with the initial values.

Issue can be reproduced at wholesale -> search


libs/ui-watt/src/lib/components/input/datepicker/watt-datepicker.spec.ts
```diff
@@ -344,6 +344,30 @@ describe('Datepicker', () => {
           )
         ).toBeInTheDocument();
       });
+
+      it('should update control value when only end date has changed', async () => {
+        await setup(withInitialValue);
+
+        const [, range] = screen.getAllByRole('button', {
```
johevemi, line 351: what does the comma mean in the const [, range] ?

johevemi (approved):

libs/ui-watt/src/lib/components/input/datepicker/watt-datepicker.spec.ts
frontend-specialisten: `screen.getAllByRole` will return two results. The first result will be the single datepicker and the second will be the datepicker with range. We could have written it like this `const [single, range] = screen.getAllByRole('button'` however since we don't use the singe datepicker for anything, we would then have an unused variable so by just writing the `,` tells that we just ignore the first "result" which is the single datepicker.

frontend-specialisten: You can find more examples here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#examples

johevemi (approved):

PR #1008, opened 2022-10-07
Mocked BFF data for market participant

mimse:
## Description

Have added mock data for market participant, so there is no need to have BFF running locally to see data in the UI


dzhavat (approved):
praise: nice PR 💯 
It will save a lot of time for everyone by not having to write our own mock data in order to test something.

Issue #922, opened 2022-09-16
Card component


renehossle:
**Description**
We need a card component as part of the Watt Design System available for feature development.

**Note**
Aligning colors in headers etc. is not in scope for this issue. It will be solved in a separate issue.

**AC**
- [ ] Should be possible to set a title (not required)
- [ ] Should be possible to set main content
- [ ] The header should support actions like pagination or tables
- [ ] Use default elevation and create a separate issue to implement the intended elevation

renehossle:
![Screenshot 2022-09-16 at 12.30.58.png](https://images.zenhubusercontent.com/62274ef6bb2c7c0c7fd46c8c/a051441b-02d3-42e2-bdbe-6b44a122f12c)

renehossle:
![Screenshot 2022-09-16 at 12.32.29.png](https://images.zenhubusercontent.com/62274ef6bb2c7c0c7fd46c8c/664cde42-44bf-40b2-9e0a-4ed6cf0f4a84)

PR #1019, opened 2022-10-11
feat(watt): add card component

dzhavat:
### Watt Design System
- add card component
- add elevation class

image

## References

- #922


libs/ui-watt/src/lib/components/card/watt-card.component.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'watt-card',
+  styles: [
+    `
+      :host {
+        border-radius: 4px;
+        display: block;
+        padding: calc(1.5 * var(--watt-space-m)); /* 24px */
```
ManBearTM, line 31: > It is not a magic number, I swear!

ManBearTM (approved):
LGTM!

PR #1023, opened 2022-10-11
Feature/configure mock service worker api base

mimse:
## Description

If apiBase value is changed, it would break the service worker mock. Because the URL is hardcoded. To fix this the service worker is now setup with the same apiBase value.


libs/dh/shared/data-access-msw/src/lib/mockedServiceWorker.ts
```diff
@@ -17,8 +17,10 @@
...
-  worker.start({ onUnhandledRequest });
-  // eslint-disable-next-line no-empty
-} catch (error) {}
+export function setupServiceWorker(apiBase: string) {
+  try {
+    const worker = setupWorker(...handlers(apiBase));
+    worker.start({ onUnhandledRequest });
+    // eslint-disable-next-line no-empty
+  } catch (error) {}
+}
```
dzhavat, line 26: nitpick: can we name the file `setup-service-worker.ts`. To reflect the function's name, but also for consistency with general file naming. Can see that some files don't follow that but let's handle that in separate PRs.

dzhavat (dismissed):
praise: nice changes. Removes a lot of manual work should we for some reason change the base url/port for local development 💯💯💯

libs/dh/shared/data-access-msw/src/lib/mockedServiceWorker.ts
mimse: Resolved in https://github.com/Energinet-DataHub/greenforce-frontend/pull/1023/commits/292a3d1e48691ab462bca94911c520779f4cbe91

ManBearTM (approved):

PR #1044, opened 2022-10-17
feat(ui-watt): intersection observer

ManBearTM:
## Description

This adds a new Intersection Observer service and then uses it in the input mask class. I had to defer the first call to `setInputColor` to fix the issue with date/time pickers in drawers.

## References

- #0000


frontend-specialisten (changes requested):
Praise: Nice work! Could you add a single integration test then I think it’s ready for merge?

frontend-specialisten (approved):

PR #1045, opened 2022-10-17
Bump wholesale webapi nuget version

johevemi:
## Description

Bumps version on wholesale webapi  nuget.
This changes batchNumber from the databricks runId long type to the batchId Guid
image

## References

- #0000


libs/dh/shared/feature-flags/src/lib/feature-flags.ts
```diff
@@ -42,7 +42,7 @@ const makeFeatureFlags = satisfies<FeatureFlagConfig>();
  */
 export const dhFeatureFlagsConfig = makeFeatureFlags({
   start_wholesale_process_feature_flag: {
-    created: '15-08-2022',
+    created: '18-10-2022',
```
BjarkeMeier, line 45: Nice :-)

BjarkeMeier (approved):

PR #1047, opened 2022-10-18
Update dh-wholesale-table.component.html

johevemi:
## References

- #0000


libs/dh/wholesale/feature-search/src/lib/table/dh-wholesale-table.component.html
```diff
@@ -33,7 +33,7 @@ <h3 class="watt-on-light--high-emphasis">{{ transloco("headline") }}</h3>
     [dataSource]="_data"
     aria-label="batches"
     matSort
-    matSortActive="batchNumber"
+    matSortActive="executionTimeStart"
```
BjarkeMeier, line 36: What if it is null?

BjarkeMeier (approved):

Issue #1032, opened 2022-10-13
Align description text on Pagination component


renehossle:
For simplicitys sake, instead of using 'XYZ per page' (different label depending on content), use 'Results per page' across the solution.

![Screenshot 2022-10-13 at 13.07.23.png](https://images.zenhubusercontent.com/62274ef6bb2c7c0c7fd46c8c/c3a5152e-25ce-424e-b889-9cd5f0d07606)
![Screenshot 2022-10-13 at 13.07.11.png](https://images.zenhubusercontent.com/62274ef6bb2c7c0c7fd46c8c/75f7c460-2878-4f53-b1e4-64870b52f937)![Screenshot 2022-10-13 at 13.06.49.png](https://images.zenhubusercontent.com/62274ef6bb2c7c0c7fd46c8c/1d6cb685-d2ae-4ad9-827c-93c5c4f32bc7)

Issue #941, opened 2022-09-21
Shared DataHub pagination component


frontend-specialisten:
**Description:**
Currently whenever the pagination component from material is used, each feature team has to provide different settings including translations etc. This could be handled by a shared DataHub component to make these more consistent and easier to add. Currently, the pagination and settings is copied around. 

_Note that current implementation of translations of pagination causes that no tooltips are shown hovering the buttons of the pagination._ 

**AC:**

- [ ] …..
- [ ] …..

**Definition of Ready:**

- [ ] The issue is correctly estimated
- [ ] The issue is adequately described
- [ ] Possible dependencies are defined and aligned
- [ ] We have the necessary skills to complete this issue
- [ ] The issue can be completed withing 1 iteration
- [ ] The issue has acceptance criteria defined
- [ ] The issue has adequate Definition Of Done described

**Definition of Done:**

- [ ] Acceptance Criteria have been met
- [ ] The product has been demo’ed for relevant stakeholders
- [ ] Dependencies are handled
- [ ] The work has been documented
- [ ] The issue has been handed over and reviewed
- [ ] The PO has accepted the product

PR #1049, opened 2022-10-19
Feat/dh shared pagination component

frontend-specialisten:
## Description

Shared pagination component.

## References

- #941 #1032


libs/dh/shared/ui-paginator/src/lib/dh-shared-ui-paginator.component.ts
```diff
@@ -0,0 +1,99 @@
...
+    >
+    </mat-paginator>
+  `,
+})
+export class DhSharedUiPaginatorComponent implements OnDestroy {
+  @ViewChild(MatPaginator, { static: true }) private paginator!: MatPaginator;
+
+  @Input() length?: number;
+  @Input() pageSizeOptions: number[] = [50, 100, 150, 200, 250];
+  @Input() pageSize?: number;
```
ManBearTM, line 46: Question: Should this have a default value as well?

libs/dh/shared/ui-paginator/src/lib/dh-shared-ui-paginator.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  ariaLabel?: string;
+
+  private matPaginatorIntl = inject(MatPaginatorIntl);
+  private translocoService = inject(TranslocoService);
+  private destroy$ = new Subject<void>();
+
+  getPaginator = () => this.paginator;
```
ManBearTM, line 54: Suggestion: Have you considered making this a getter function instead? The function call makes it seem like it has a non-trivial cost (fx. I would be hesitant to call this in ngOnChanges, even though it is basically free). Although `this.paginator.paginator` looks wrong, perhaps `this.paginator.instance`?

libs/dh/shared/ui-paginator/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+   ```ts
+   @ViewChild(DhSharedUiPaginatorComponent) paginator!: DhSharedUiPaginatorComponent;
+   ```
+
+4. Connect the data source with the paginator:
+
+   ```ts
+   ngAfterViewInit() {
+     this.dataSource.paginator = this.paginator.getPaginator();
```
ManBearTM, line 25: If you decide to change the function call as mentioned below, then don't forget to update the docs here 😄

libs/dh/shared/ui-paginator/README.md
```diff
@@ -0,0 +1,40 @@
...
+| --------------- | ------------------------------------------------------------------------------------------------------------------------------------------ | -------------------------- |
+| length          | The length of the total number of items that are being paginated. **Note: using `MatTableDataSource` will overwrite this value**. `number` | `0`                        |
+| pageSizeOptions | The set of provided page size options to display to the user. `number[]`                                                                   | `[50, 100, 150, 200, 250]` |
+| pageSize        | Number of items to display on a page. `number`                                                                                             | `50`                       |
+
+## MatTableDataSource
+
+The built-in `MatTableDataSource` is designed for filtering, sorting and pagination of a **client-side** data array.
+For server-side filtering, sorting and pagination a custom data source is necessary [Read more.](https://blog.angular-university.io/angular-material-data-table)
```
ManBearTM, line 40: Praise: Love the docs!

libs/dh/shared/ui-paginator/src/lib/dh-shared-ui-paginator.component.ts
```diff
@@ -0,0 +1,99 @@
...
+  @ViewChild(MatPaginator, { static: true }) private paginator!: MatPaginator;
+
+  @Input() length?: number;
+  @Input() pageSizeOptions: number[] = [50, 100, 150, 200, 250];
+  @Input() pageSize?: number;
+
+  ariaLabel?: string;
+
+  private matPaginatorIntl = inject(MatPaginatorIntl);
+  private translocoService = inject(TranslocoService);
```
ManBearTM, line 51: Noob question: Is there any difference between this and constructor DI?

ManBearTM (commented):
Stellar work, removes a lot of duplication! Only a few comments, just minor stuff.

libs/dh/shared/ui-paginator/src/lib/dh-shared-ui-paginator.component.ts
frontend-specialisten: Angular materials sets the `pagesize` to have a default value to 50, I will add it to make it explicit :) 
image

libs/dh/shared/ui-paginator/src/lib/dh-shared-ui-paginator.component.ts
frontend-specialisten: There shouldn't be any differences regarding the dependency injection itself. However, this syntax is not as "magic" as the constructor-based injection and actually, this also makes it possible to make reusable functions using dependency injections. Example:
image

ManBearTM: I love it, will definitely use that more!

ManBearTM (approved):
💯

Issue #1018, opened 2022-10-10
Scrollable table


ManBearTM:
**Description:**
The design in Figma (https://www.figma.com/proto/4tVkjKm5fkIj61nx4XD3pA/DataHub-3.0?node-id=2520%3A24104&scaling=contain&page-id=2419%3A16889&starting-point-node-id=2520%3A24104&show-proto-sidebar=1) for the charges search result table involves a scrollable table area, where the table height never exceeds the page height. Making tables scrollable is not an easy task, and most solutions are dirty hacks with harsh tradeoffs. After investigating a little, I see two solutions (there might be more):

* Make the entire "card" scrollable and try to have two sticky components (pagination AND column headers). This will look a little off with the scrollbar though. Might be tricky to pull off.
* Create our own watt-table component that uses CSS grid for layout (could still use mat-table underneath, but layout is completely replaced with CSS grid).

**AC:**

- [ ] …..
- [ ] …..

**Definition of Ready:**

- [ ] The issue is correctly estimated
- [ ] The issue is adequately described
- [ ] Possible dependencies are defined and aligned
- [ ] We have the necessary skills to complete this issue
- [ ] The issue can be completed withing 1 iteration
- [ ] The issue has acceptance criteria defined
- [ ] The issue has adequate Definition Of Done described

**Definition of Done:**

- [ ] Acceptance Criteria have been met
- [ ] The product has been demo’ed for relevant stakeholders
- [ ] Dependencies are handled
- [ ] The work has been documented
- [ ] The issue has been handed over and reviewed
- [ ] The PO has accepted the product

PR #1050, opened 2022-10-19
experiment(ui-watt): css grid table

ManBearTM:
## Description

Just a little attempt at making the material table use CSS grid, in order to have a "scrollable body".

## References

- #1018


ManBearTM:
> We should however remember to make sure the determined `max-height` is responsive and doesn't break mobile so EO can still make use of the tables.

For suresies, will have to play around with the responsiveness some more 🙂 The `300px` was just to force the scrollbars here, but good point!

Issue #924, opened 2022-09-16
Copy to Clipboard


renehossle:

renehossle:
![Screenshot 2022-09-16 at 12.40.44.png](https://images.zenhubusercontent.com/62274ef6bb2c7c0c7fd46c8c/a13fb168-11ec-47b9-8e79-76370a2f6af5)

renehossle:
![Screenshot 2022-09-16 at 12.40.52.png](https://images.zenhubusercontent.com/62274ef6bb2c7c0c7fd46c8c/b483b4a0-15e8-435a-8f43-63e8ce8ffc57)

PR #1052, opened 2022-10-19
feat(ui-watt): copy to clipboard directive

ManBearTM:
## Description
This is a relatively low-level directive for adding "copy to clipboard" functionality to any element. The reasoning for keeping it at this level is that it is going to be used in a variety of ways, so adding styles to it would not make much sense. It also requires translations (which cannot be provided by watt), so this encourages the consuming teams to create their own abstraction on top of this directive.

## References
- #924


frontend-specialisten (commented):
Praise: nice work! The only thing I will add as consideration is the status messages `wattCopyToClipboardSuccess` and `wattCopyToClipboardError.` Isn't it annoying to be forced to set these every time using the directive and should they be configurable at all? like the text for the pagination, this should properly be clarified with UX (if you haven't done this already, to ensure consistency).  If it should be configurable, I would advocate for some sensible defaults provided once by the DataHub application through a token(s).

libs/ui-watt/src/lib/components/clipboard/watt-copy-to-clipboard.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+        message: this.translation.translate('clipboard.success'),
+      });
+    } else {
+      this.toast.open({
+        type: 'danger',
+        message: this.translation.translate('clipboard.error'),
+      });
+    }
+  }
+}
```
dzhavat, line 68: praise: the whole thing works quite well - copying, translations and showing a toast.

suggestion: looking at the implementation as a whole, it seems like it's taken on a few responsibilities, which might be better off to be somewhere else. There's the copy functionality which is the core of the directive. Then there's a translation layer, and a notification/toast functionality.

I'm thinking of a possible split. What if we let the translation layer, and the notification/toast be handled by the host app? DH in our case. So from `WattCopyToClipboardDirective` we expose an `Output` that is called with the result of `this.clipboard.copy`. Then the host app will have to decide how to notify the user about the copy being successful/unsuccessful.

This leads to another issue that the developers should handle the notification/toast part. In this case we can make it really easy for them by creating another directive specific to DH that is applied to the same element the `wattCopyToClipboard` is applied (let's theoretically call it `dhNotifyOnCopy`). This way the DH directive can inject the `WattCopyToClipboardDirective` instance and subscribe to its `Output`.

So the end result for the developer will be:

```html
<span wattCopyToClipboard dhNotifyOnCopy`>Copy text</div>
```

If we feel adventurous enough, we can also use a `wattCopyToClipboard` selector in the DH directive. This way the `dhNotifyOnCopy` will not be needed at all on the element.

Advantages of this solution:
- no custom translation layer in the copy directive
- no extra translation files to maintain
- existing translation layer can be used
- each app decides how to notify the user on copy
- it's still easy to show a toast on copy by creating the extra directive

What do you think?

frontend-specialisten: I'm not a huge fan of having multiple directives for this case. If the problem is having multiple translation files or the fact that transloco is used inside of Watt my suggestion would be to configure the clipboard by just defining a token just like we set the locale of the date picker.

`
@NgModule({
  providers: [
    {provide: WATT_CLIPBOARD_SOMETHING_SOMETHING, useValue: ...},
  ],
})
export class MyApp {}
`

libs/ui-watt/src/lib/components/toast/index.ts
```diff
@@ -15,5 +15,6 @@
  * limitations under the License.
  */
 export { WattToastModule } from './watt-toast.module';
+export { WattToastComponent } from './watt-toast.component';
```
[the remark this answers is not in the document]
ManBearTM: ec08461d3480ab7403d58e9a2531bf8b431743d6

libs/ui-watt/src/lib/components/clipboard/watt-copy-to-clipboard.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const WATT_COPY_TO_CLIPBOARD_DEPS = [
+  WattToastComponent,
```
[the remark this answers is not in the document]
ManBearTM: 7152ee6c37063444c6156266497110c95e48680f

libs/ui-watt/src/lib/components/clipboard/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export {
+  WattCopyToClipboardDirective,
+  WATT_COPY_TO_CLIPBOARD_DEPS,
+} from './watt-copy-to-clipboard.directive';
```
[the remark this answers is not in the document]
ManBearTM: 7152ee6c37063444c6156266497110c95e48680f

dzhavat:
praise: looks great 🌟

question: do you think we should configure DH to replace `WattClipboardIntlService` and work with changing language, etc. so it's ready for other developers to just use? It can also be a follow up PR.

dzhavat:
question: what about tests? Should we add some as well?

libs/dh/globalization/ui-watt-translation/src/lib/dh-globalization-ui-watt-translation.module.ts
```diff
@@ -0,0 +1,70 @@
...
+    transloco
+      .selectTranslate('clipboard.error')
+      .subscribe((value) => (this.error = value));
+  }
+}
+
+/**
+ * Do not import directly. Use `DhConfigurationLocalization.forRoot`.
+ */
+@NgModule({})
```
dzhavat, line 45: nitpick: 

```suggestion
/**
 * Do not import directly. Use `DhGlobalizationUiWattTranslationModule.forRoot`.
 */
@NgModule()
```

ManBearTM: Nice catch!

dzhavat (dismissed):

dzhavat (approved):

PR #1066, opened 2022-10-25
feat: Support incremental scopes consent.

defectiveAi:
Add support for incrementally requesting consent for scopes, i.e. when changing actors.


libs/dh/shared/feature-authorization/src/lib/scope.service.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    // act
+    const actual = target.getActiveScope();
+
+    // assert
+    expect(actual).toEqual('actor1');
+    expect(storage.getItem(localAccountId + scopeKey)).toEqual('actor1');
+  });
+});
+
+class LocalStorageMock {
```
dzhavat, line 153: suggestion: Looks like [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) uses `Storage` interface to implement its methods. Let's use the same to make the mock similar to how `localStorage` works.

```suggestion
class LocalStorageMock implements Storage {
```

After adding the interface, TS will complain about missing `length` and `clear` properties.

In order to implement the `length` property, we can turn it into a getter that reads from another internal property. E.g.

```ts
class LocalStorageMock implements Storage {
  store: { [key: string]: string } = {};

  private internalLength = 0;

  public get length() {
    return this.internalLength;
  }

  public readonly setItem = (key: string, value: string) => {
    this.store[key] = value;
    this.internalLength = Object.keys(this.store).length;
  };

  // ...

  public readonly clear = () => {
    this.store = {};
    this.internalLength = 0;
  };
}
```

The `getLength` method should not be part of this mock, since it's not a property on the real `localStorage`.

libs/dh/shared/feature-authorization/src/lib/scope-storage.service.ts
```diff
@@ -0,0 +1,49 @@
+/**
```
dzhavat, line 1: nitpick: rename the file to `scope-storage.ts` since it's just a simple class and not an Angular service.

libs/dh/shared/feature-authorization/src/lib/scope.service.ts
```diff
@@ -16,18 +16,105 @@
...
+
+        if (!account) {
+          this.clearScopeRelatedItems();
+          return;
+        }
+
+        const userId = account.localAccountId;
+        const actors =
+          account.idTokenClaims &&
+          (account.idTokenClaims['extn.actors'] as string[]);
```
dzhavat, line 56: suggestion: consider moving `extn.actors` to a variable as well. It's used in the spec file too.

libs/dh/shared/feature-authorization/src/lib/scope.service.ts
```diff
@@ -16,18 +16,105 @@
...
 import {
   DhB2CEnvironment,
   dhB2CEnvironmentToken,
 } from '@energinet-datahub/dh/shared/environments';
+import { DhFeatureFlagsService } from '@energinet-datahub/dh/shared/feature-flags';
+import { filter } from 'rxjs';
+import { ScopeStorage, scopeStorageToken } from './scope-storage.service';
+
+export const scopesKey = 'actor-scopes';
+export const scopeKey = 'active-actor-scope';
```
dzhavat, line 30: suggestion: let's make this variable names a bit more explicit. Consider `actorScopesKey` and `activeActorScopeKey`.

libs/dh/shared/feature-authorization/src/lib/scope-storage.service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { InjectionToken } from '@angular/core';
+
+export class ScopeStorage {
+  private length = 0;
+
+  public readonly getLength = () => this.length;
+
+  public readonly setItem = (key: string, value: string) => {
+    localStorage.setItem(key, value);
```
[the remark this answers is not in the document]
defectiveAi: Yes, it is safe, because:
- Scopes are not secret, they are just an API ressource, `api://<guid>/actor.default`
- Scopes do not grant permission, tokens do. If you ask for a token using a scope you are not part of, you will get nothing.
- If I "steal" a scope from localstorage than belongs to another user, I would have a random guid representing an actor. I would still need to request an access_token using this scope - that would fail if I am not a member of that actor.

We do not want to clear the scopes at logout, as we store the users preferred scope/actor across sessions.

libs/dh/shared/feature-authorization/src/lib/scope.service.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    clientId: clientId,
+    knownAuthorities: [],
+  };
+
+  return new ScopeService(
+    config,
+    msalBroadcastService as MsalBroadcastService,
+    msalService as MsalService,
+    featureFlagService as DhFeatureFlagsService,
+    store as unknown as ScopeStorage
```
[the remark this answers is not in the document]
michaelmajgaard: fixed

libs/dh/shared/feature-authorization/src/lib/scope.service.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  public readonly removeItem = (key: string) => {
+    delete this.store[key];
+    this.length = Object.keys(this.store).length;
+  };
+
+  public readonly key = (index: number) => {
+    return Object.keys(this.store).at(index) ?? null;
+  };
+}
```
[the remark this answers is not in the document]
michaelmajgaard: fixed

libs/dh/shared/feature-authorization/src/lib/scope.service.spec.ts
michaelmajgaard: fixed

libs/dh/shared/feature-authorization/src/lib/scope-storage.service.ts
michaelmajgaard: fixed

libs/dh/shared/feature-authorization/src/lib/scope.service.ts
michaelmajgaard: fixed

libs/dh/shared/feature-authorization/src/lib/scope.service.ts
michaelmajgaard: fixed

libs/dh/shared/test-util-auth/src/lib/local-storage-fake.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    this[key] = value;
+    this.internalLength = Object.keys(this).length;
+  };
+
+  public readonly getItem = (key: string) => {
+    return this[key];
+  };
+
+  public readonly removeItem = (key: string) => {
+    delete this[key];
```
dzhavat, line 38: issue: this will also delete one of the methods (e.g. `getItem` or `setItem`) if sent by mistake.

libs/dh/shared/feature-authorization/src/lib/scope.service.ts
```diff
@@ -16,18 +16,114 @@
...
+        this.localAccountId = account.localAccountId;
+        this.authService.instance.setActiveAccount(account);
+
+        const userId = account.localAccountId;
+        const actors =
+          account.idTokenClaims &&
+          (account.idTokenClaims[actorScopesClaimsKey] as string[]);
+        const scopes = this.getActorClaims(actors);
+
+        this.scopeStorage.setItem(userId + actorScopesKey, scopes.join(' '));
```
dzhavat, line 73: suggestion: using the `userId + actorScopesKey` approach results in `353d36a1-c62f-4905-8318-842b009cd008actor-scopes` key. Since the `userId` also contains letters, the `actorScopesKey` value (or a part of it) can be confused as being part of the `userId` string. Consider separating the two. It could be by dot `.`. Can see that MSAL is using this approach (`msal.88e5d356-0c71-49e9-b260-d0629f3c0445.active-account`).

This comment is also related to all other places where `userId + actorScopesKey | actorScopesKey` is used. To make this easier, the logic can be moved to a method, so it can also be used in tests.

libs/dh/shared/feature-authorization/src/lib/scope-storage.ts
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InjectionToken } from '@angular/core';
+
+export class ScopeStorage {
+  constructor(private localStorage: Storage) {}
+
+  public length = 0;
```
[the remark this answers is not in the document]
michaelmajgaard: Fixed. Refactored service to be scope specific.

libs/dh/shared/test-util-auth/src/lib/local-storage-fake.service.ts
michaelmajgaard: Fixed. ScopeStorage no longer has a public removeItem method.

libs/dh/shared/feature-authorization/src/lib/scope.service.ts
michaelmajgaard: Fixed

libs/dh/shared/feature-authorization/src/lib/scope.service.ts
```diff
@@ -16,18 +16,114 @@
...
+
+    for (let i = 0, l = this.scopeStorage.length; i < l; ++i) {
+      const key = this.scopeStorage.key(i);
+      if (
+        key &&
+        (key.endsWith(activeActorScopeKey) || key.endsWith(actorScopesKey))
+      ) {
+        keys.push(key);
+      }
+    }
```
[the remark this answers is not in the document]
michaelmajgaard: Fixed. On ACQUIRE_TOKEN_FAILURE we clear scopes for the user.

libs/dh/shared/test-util-auth/src/lib/local-storage-fake.service.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export class LocalStorageFake implements Storage {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [name: string]: any;
```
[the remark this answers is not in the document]
michaelmajgaard: Not sure I understand? Is it the eslin-disable-next-line you're questioning? :)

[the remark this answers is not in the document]
dzhavat: At one point this class used a `store` property to save keys/values [reference](https://github.com/Energinet-DataHub/greenforce-frontend/pull/1066/commits/3ac883cc3623b8483c3046ecced82cc2f8d6f8f8#diff-21a0b8a929f0679b92b9740e289cf582a0d33c3dcb7a8f46ba2a361ae1744669L153-L178). With the current implementation, every saved item is added as a property on the class. Why not use a single property as before instead?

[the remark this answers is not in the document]
michaelmajgaard: Fixed

dzhavat (approved):

PR #1067, opened 2022-10-26
feat(ui-watt): Add output changes to tabs

ShaBloze:
## Description

This is needed for the charge drawer, which has tabs. 
We want to know whenever a tab has been selected, so we can load the data inside that tab. 
We would also like to know which tab is currently active when the drawer is open, to load the correct data.

## References

- #0000

### DIFF
--- a/libs/ui-watt/src/lib/components/tabs/tab/tab.component.ts
+++ b/libs/ui-watt/src/lib/components/tabs/tab/tab.component.ts
@@ -31,7 +31,7 @@ export class WattTabComponent {
   @Input() label = '';
   @ViewChild('templateRef') public templateRef: TemplateRef<unknown> | null =
     null;
-  @Output() changed = new EventEmitter();
+  @Output() changed = new EventEmitter<void>();
 
   emitChange() {
     this.changed.emit();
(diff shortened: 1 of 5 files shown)


libs/ui-watt/src/lib/components/tabs/tab/tab.component.ts
```diff
@@ -24,4 +31,9 @@ export class WattTabComponent {
   @Input() label = '';
   @ViewChild('templateRef') public templateRef: TemplateRef<unknown> | null =
     null;
+  @Output() changed = new EventEmitter();
```
frontend-specialisten, line 34: Suggestion: add type for the event emitter
```suggestion
  @Output() changed = new EventEmitter<void>();
```

libs/ui-watt/src/lib/components/tabs/tabs.component.ts
```diff
@@ -39,4 +39,17 @@ export class WattTabsComponent {
...
+
+  emitSelectedTabChange(selectedIndex: number) {
+    this.activeTabIndex = selectedIndex;
+    const currentTab = this.tabElements.find(
+      (tab, index) => index === selectedIndex
+    );
+    currentTab?.emitChange();
+  }
+
+  triggerChange() {
```
frontend-specialisten, line 52: Question: where is `triggerChange()` used or going to be used? :)

ShaBloze: In our case it is going to be used when we open a drawer with tabs inside. Then we would like to make the active tab emit an event, so we can get the correct data inside the tab content. 

I hope it makes sense :)

frontend-specialisten (approved):

PR #1068, opened 2022-10-26
feat:the download button is enabled and will generate a csv file with all charges from search result

x-platformcoder:
## Description

When pressing the download button on charge search page, a csv file is created with the search results

## References

story 1766 from the Volt board

- #1766


ShaBloze (approved):
LGTM

PR #1073, opened 2022-10-27
Certificates sites

MadsThy:
## Description

Start of the new certificates site

Can be shown on the page using ?enableFeature=certificates in the URL

image


libs/eo/certificates/src/lib/eo-certificates.store.ts
```diff
@@ -0,0 +1,80 @@
...
+import { HttpErrorResponse } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { ComponentStore } from '@ngrx/component-store';
+import {
+  EoCertificate,
+  EoCertificatesService,
+} from './eo-certificates.service';
+
+interface EoCertificatesState {
+  loadingDone: boolean;
```
NielsJorck, line 26: Consider using a naming convention for booleans indication a boolean value.
Eg. prefix with "is" or "has".
Examples: "hasLoaded" or "isLoading"

NielsJorck (dismissed):

libs/eo/certificates/src/lib/eo-certificates.store.ts
MadsThy: Fixed :)

NielsJorck (approved):

PR #1074, opened 2022-10-27
chore: update dependencies

dzhavat:
### Workspace
- update to latest Nx v15
- update to latest Angular v14

Note: all changes in files other than `package.json` are due to Nx migrations

## References

- N/A


nx.json
```diff
@@ -55,7 +48,33 @@
...
+    "lint": {
+      "inputs": ["default", "{workspaceRoot}/.eslintrc.json"]
+    },
+    "build-storybook": {
+      "inputs": ["default", "^production"]
     }
+  },
+  "namedInputs": {
+    "default": ["{projectRoot}/**/*", "sharedGlobals"],
+    "sharedGlobals": [],
```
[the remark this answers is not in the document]
dzhavat: No :D 

All changes except the ones in `package.json` (and `yarn.lock`) are coming from Nx as automatic migrations. 

Digging in their changelog, and looking at the description in this PR https://github.com/nrwl/nx/pull/12416, it looks like `sharedGlobals` is used for "Various plugins that use babel". So my guess is that it's empty because we don't use Babel.

apps/dh/api-dh/project.json
```diff
@@ -17,7 +17,7 @@
         ],
         "parallel": false
       },
-      "outputs": ["apps/dh/api-dh/source/DataHub.WebApi/bin/Release/net6.0"]
+      "outputs": ["{projectRoot}/source/DataHub.WebApi/bin/Release/net6.0"]
```
frontend-specialisten, line 20: Question: are we sure this is right?

apps/dh/api-dh/project.json
```diff
@@ -70,7 +70,7 @@
         "commands": ["npx ts-node tools/dh/api-dh/ci/stop-api-dh.ts"],
         "parallel": false
       },
-      "outputs": ["apps/dh/api-dh/source/DataHub.WebApi/bin/Release/net6.0"]
```
frontend-specialisten, line 73: Question: are we sure this is right?

frontend-specialisten (commented):
LGTM besides some of the path changes for the BFF Project, I think we should double check :)

apps/dh/api-dh/project.json
dzhavat: I think it is. You can see the migration [here](https://github.com/nrwl/nx/pull/12470/files#diff-c8817fc8a17002f12e5c8565ad91616142cd2ccdd965c47b18bb0bcf7d9f442e). It looks like if the path has something to do with the current project, then it's prefixed with `{projectRoot}`. On the other hand, if the path has something to do with another project in the workspace, it's prefixed with `{workspaceRoot}`.

apps/dh/api-dh/project.json
dzhavat: Resolving this comment since it's similar to the other one.

frontend-specialisten (approved):

PR #1075, opened 2022-10-27
feat: added gln, renamed columns and sorted by chargeid

x-platformcoder:
## Description

added gln no as column to csvfile
changed header "Id" to CHARGE ID/ PRIS ID
now sorts rows in csvfile from ChargeId

## References

story 1766 on volt board
- #1766


ShaBloze (approved):
LGTM

PR #1062, opened 2022-10-21
Chore/remove watt barrel

frontend-specialisten:
## References

- #0000

PR #1081, opened 2022-10-30
chore(ui-watt): update tsconfig.base.json

Lubbe91:
## Description

Seems after some of the changes done in the recent update to the repository, I've started to receive problems with incorrect pathing in various `config.lib` files.
 
Screenshot 2022-10-30 213834

I've traced these paths to be causing my IDE to complain and propose to update them accordingly[^1].

[^1]: Not sure about removing the reference to the index files as it seems to have a reference in the [app styles file](https://github.com/Energinet-DataHub/greenforce-frontend/blob/main/apps/dh/app-dh/src/styles.scss)

## References

- #1062

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -285,7 +285,6 @@
       "@energinet-datahub/gf/util-browser": [
         "libs/gf/util-browser/src/index.ts"
       ],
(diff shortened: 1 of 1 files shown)


dzhavat (approved):
praise: nice findings 🥇🥇

PR #1103, opened 2022-11-01
Added resolution picker on consumption and production pages

NielsJorck:
## Description

Added resolution picker to be used in conjunction with date picker and graph data.
image

## References

https://github.com/Energinet-DataHub/energy-origin-issues/issues/725


libs/eo/consumption-page/shell/src/lib/eo-consumption-page-shell.component.ts
```diff
@@ -56,11 +59,15 @@ import { EoConsumptionStore } from './eo-consumption.store';
...
         ></eo-consumption-line-chart>
         <eo-date-picker
+          class="watt-space-stack-l"
           [onFeatureFlag]="'daterange'"
           *rxLet="appSettingsDates$ as dates"
           [dateRangeInput]="dates"
           (newDates)="setNewAppDates($event)"
         ></eo-date-picker>
+        <eo-resolution-picker
+          [onFeatureFlag]="'resolution'"
```
MadsThy, line 69: I mean... im not complaining that you made a new feature flag, but the idea was the hide resolution and datepicker behind the datepicker flag 😆 But good! Thank you for using it 😄

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-resolution-picker/eo-resolution-picker.component.ts
```diff
@@ -0,0 +1,149 @@
...
+      }
+
+      mat-button-toggle {
+        border-color: var(--watt-color-primary) !important;
+      }
+
+      mat-button-toggle.mat-button-toggle-appearance-standard
+        .mat-button-toggle-label-content {
+        line-height: unset;
+        padding: 10px 14px;
```
MadsThy, line 64: Are these magical UX sizes? 😉

libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-resolution-picker/eo-resolution-picker.component.ts
```diff
@@ -0,0 +1,149 @@
...
+    `,
+  ],
+  template: `<label class="watt-space-stack-s title">Resolution</label>
+    <mat-button-toggle-group
+      [value]="appSettingsResolution$ | async"
+      name="resolution"
+    >
+      <mat-button-toggle
+        [disabled]="
+          !(((appSettingsCalenderDateRangeInDays$ | async) ?? 0) <= 7)
```
MadsThy, line 84: can i maybe get you to use rxLet in the mat-button-toggle-group, and then reference that variable instead of using these async calls? 😄 I think they can be more reliable that way. Remember how the !(whatever | async) negation calls made us unhappy in Eloverblik? :-)

Issue #979, opened 2022-09-30
Spike: Tooltips used within table sometimes empty


ManBearTM:
![tooltips.png](https://images.zenhubusercontent.com/626fbc6d95e6587091961d95/2378e744-15a6-425c-95f4-b96a2c7eba20)

**Description**

When hovering the icons in the charges result table, the tooltip sometimes appears empty. Inspecting the DOM shows that multiple empty tooltips are created on hover. This issue also happens with regular matTooltip, so it is not a specific to wattTooltip.

Other people have encountered this issue: https://github.com/angular/components/issues/24278

**AC:**

- [ ] Investigate what's the best solution to fix empty tooltips
- [ ] Decide if issue is fixed as part of this or a new issue is created to fix it

**Definition of Ready:**

- [ ] The issue is correctly estimated
- [ ] The issue is adequately described
- [ ] Possible dependencies are defined and aligned
- [ ] We have the necessary skills to complete this issue
- [ ] The issue can be completed withing 1 iteration
- [ ] The issue has acceptance criteria defined
- [ ] The issue has adequate Definition Of Done described

**Definition of Done:**

- [ ] Acceptance Criteria have been met
- [ ] The product has been demo’ed for relevant stakeholders
- [ ] Dependencies are handled
- [ ] The work has been documented
- [ ] The issue has been handed over and reviewed
- [ ] The PO has accepted the product

PR #1106, opened 2022-11-02
Fix/tooltips

frontend-specialisten:
## Description

- Replace `Angular Material - Tooltips` with own implementation of popperjs
- Adjust storybook examples / showcase
image

- Fixes issues with empty tooltips used in the DataHub application (see #979)

image

## References

- #979


libs/ui-watt/src/lib/components/tooltip/watt-tooltip.component.scss
```diff
@@ -0,0 +1,109 @@
...
+  z-index: 500;
+  display: inline-block;
+  background: var(--watt-color-primary-dark);
+  color: var(--watt-color-primary-dark-contrast);
+  line-height: 18px;
+  border-radius: 6px;
+  @include watt.space-inset-squish-s;
+  @include watt.typography-watt-text-s;
+  opacity: 0;
+  transition: opacity 0.3s ease-in-out;
```
dzhavat, line 33: issue: some of the arrows (e.g. `top-start` or `top-end`) are misaligned.

image

dzhavat (commented):
praise: looks good! Only one comment :)

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.component.scss
```diff
@@ -0,0 +1,109 @@
...
+ * limitations under the License.
+ */
+@use "sass:math";
+@use "@energinet-datahub/watt/utils" as watt;
+
+$arrow-size: 8px;
+$offset: 6px;
+
+:host {
+  z-index: 500;
```
ManBearTM, line 24: Is this high enough for use in modals? 😄

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.component.ts
```diff
@@ -0,0 +1,144 @@
...
+  private renderer: Renderer2 = inject(Renderer2);
+  private focusMonitor: FocusMonitor = inject(FocusMonitor);
+
+  /** Emits when the component is destroyed. */
+  private readonly destroyed = new Subject<void>();
+  private listeners: unlistenerFunction[] = [];
+  private showClass = 'show';
+
+  ngAfterViewInit(): void {
+    createPopper(this.target, this.element, {
```
ManBearTM, line 69: This returns an instance object with a destroy method - do we need to call this when the component is destroyed?

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.component.scss
frontend-specialisten: Should be fixed now :)

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.component.scss
frontend-specialisten: It is, I've added an example of the tooltip to the modal showcase, so it's easy to test in the future.

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.directive.ts
```diff
@@ -14,87 +14,52 @@
...
+  | 'bottom-start'
+  | 'bottom'
+  | 'bottom-end'
+  | 'left';
 
 @Directive({
   selector: '[wattTooltip]',
+  standalone: true,
   exportAs: 'wattTooltip',
 })
```
dzhavat, line 42: praise: nice work 💪 Approving the changes so far.

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+    this.renderer.addClass(this.element, this.showClass);
+  }
+
+  private hide(): void {
+    this.renderer.removeClass(this.element, this.showClass);
+  }
+
+  private platformSupportsMouseEvents() {
+    return !this.platform.IOS && !this.platform.ANDROID;
```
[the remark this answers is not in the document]
ManBearTM: I guess it might cause issues with touchscreen PC's like ours, which sort of supports both 🤔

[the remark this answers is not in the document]
frontend-specialisten: Yes this is exactly how material is doing it:
https://github.com/angular/components/blob/main/src/material/tooltip/tooltip.ts#L794

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.directive.spec.ts
```diff
@@ -35,14 +35,50 @@ async function setup() {
...
+  const getTooltipTarget = () => screen.getByText('Button');
+
+  const isTooltipVisible = (): boolean =>
+    getTooltip().classList.contains('show');
+
   it('always displays accessible tooltip', async () => {
     await setup();
-    expect(screen.getByRole('tooltip', { hidden: true })).toBeInTheDocument();
+    expect(screen.getAllByRole('tooltip')[0]).toBeInTheDocument();
   });
```
ManBearTM, line 47: FYI, I think this test was meant to ensure that tooltips are always present in the DOM, to allow for things like screen readers to see them without requiring hover. I don't think that is the case with with this implementation, since the tooltips are not created until `show()` is triggered. No worries though, as long as we don't have any strict requirements regarding accessibility it probably doesn't matter.

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.component.ts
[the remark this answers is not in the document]
ManBearTM: Then let's not pretend we know any better 😄

ManBearTM (approved):
Nice work!

libs/ui-watt/src/lib/components/tooltip/watt-tooltip.directive.spec.ts
frontend-specialisten: Actually this is not the case, the tooltips are still created initially. It's only `popper` which are initialized on show :)

ManBearTM (approved):

Issue #948, opened 2022-09-22
Date-range picker - end date should be included


frontend-specialisten:
Choosing a date range should include the whole end date with time. 

**Example:**
Currently: _2022-10-12T22:00:00.000Z (Thu Oct 13 2022 00:00:00 GMT+0200 (Central European Summer Time)_
What we want: _2022-10-13T21:59:59.999Z (Thu Oct 13 2022 23:59:59 GMT+0200 (Central European Summer Time)_

- [x] Apply changes to the date picker
- [x] Remove all temp fixes implemented eg. https://github.com/Energinet-DataHub/greenforce-frontend/pull/1024

**Technical notes:**
A temp fix has been created because it wasn't as easy to implement the change inside the date picker as you would think. We have some issues updating the model accordingly.

PR #1108, opened 2022-11-02
feat(watt): adjust to date in date range picker to always be end of selected day

dzhavat:
### Watt Design System
- set "to" date in data range picker to be end of the selected day (i.e. `2022-11-02T22:59:59.999Z` (UTC) which is `2022-11-02T23:59:59.999+01:00` Danish time)

Demo

demo

Note: The `end` date is always set to "end of selected day" internally in the datepicker component on initialization. So it doesn't matter what value is set to `end` initially in the component where the datepicker is used.

### Workspace
- remove temp fixes related to "to" date in date range picker

## References

- #948


libs/dh/wholesale/feature-search/src/lib/form/dh-wholesale-form.component.ts
```diff
@@ -62,7 +62,7 @@ export class DhWholesaleFormComponent implements OnInit, OnDestroy {
       {
         start: sub(new Date().setHours(0, 0, 0, 0), { days: 10 }).toISOString(),
         end: zonedTimeToUtc(
-          new Date().setHours(0, 0, 0, 0),
+          new Date().setHours(23, 59, 59, 999),
```
frontend-specialisten, line 65: Question: is ‘setHours’ still necessary?

dzhavat: Not really but it might be useful for mentally mapping the initial value with the value written back to the model.

It doesn't matter what value is set to "end", it will automatically be converted to end of day on initialization.

Say it was `new Date()` (`2022-11-02T13:05:37.038Z`) before initializing the component, afterward initialization the model will change to `2022-11-02T22:59:59.999Z`. Or something like `2022-11-01T23:00:00Z` came from the API, it will also be adjusted to `2022-11-02T22:59:59.999Z`.

So while technically it's not necessary, it might be desired for consistency.

If we decide to keep it, we can extract it to a util method (say `setToEndOfDay`) so developers don't have to use `setHours` directly.

What do you think?

dzhavat: To be consistent with how the datepicker in Material works, the initial value is not changed. So whatever is set to the model, that will be used in the datepicker. Changing the date afterwards (either by selecting it from the UI or manually typing it) will result in the "to" date always being set to end of the day.

dzhavat: After the latest changes, the "end" date is always set to "end of selected day" internally in the datepicker component on initialization. So it doesn't matter what value is set to `end` initially in the component where the datepicker is used.

frontend-specialisten (approved):
Praise: Good work!

PR #1120, opened 2022-11-08
fix: Charge message not getting message

ShaBloze:
## Description

Charge message not getting the message from MessageArchive. Most likely because the model was not filled.

## References

- #1693
https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/energinet-datahub/geh-charges/1693


libs/dh/charges/feature-prices/src/lib/drawer/dh-charges-prices-drawer.service.ts
```diff
@@ -15,26 +15,27 @@
...
 import { Injectable } from '@angular/core';
+import { ChargeMessageV1Dto } from '@energinet-datahub/dh/shared/domain';
 import { BehaviorSubject } from 'rxjs';
 
 @Injectable({
   providedIn: 'root',
 })
 export class DhChargesPricesDrawerService {
-  defaultMessageId?: string = undefined;
+  defaultMessageId?: ChargeMessageV1Dto = undefined;
```
HenrikSommer, line 25:
```suggestion
  defaultChargeMessageV1Dto?: ChargeMessageV1Dto = undefined;
```

libs/dh/charges/feature-prices/src/lib/drawer/dh-charges-prices-drawer.component.html
```diff
@@ -21,17 +21,17 @@
...
-      <ng-container *ngIf="(messageId$ | push) !== undefined">
+      <ng-container *ngIf="(message$ | push) !== undefined">
         <watt-icon name="right"></watt-icon>
-        <span>{{ messageId$ | push }}</span>
+        <span>{{ (message$ | push)?.messageId }}</span>
       </ng-container>
     </watt-drawer-topbar>
 
-    <watt-drawer-actions *ngIf="(messageId$ | push) === undefined">
+    <watt-drawer-actions *ngIf="(message$ | push) === undefined">
```
HenrikSommer, line 34: Why does the visibility of _stop_ and _edit_ rely on `message` rather than `charge`?

ShaBloze: Because we don't want to show the drawer Actions when showing a Message.

I wanted to do this in a prettier way, but I couldn't figure out how.

HenrikSommer: Maybe define a new boolean with a name that clearly tells whether the buttons should be visible, and then set the boolean in ts based on the `message`?

ShaBloze: Good idea, I will do that 👍

HenrikSommer (approved):

libs/dh/charges/feature-prices/src/lib/drawer/dh-charges-prices-drawer.component.html
HenrikSommer: I can see you where able to reuse an already existing variable 👍

PR #1123, opened 2022-11-08
Feat/breadcrumbs

frontend-specialisten:
## Description
Breadcrumbs component

image


libs/ui-watt/src/lib/components/breadcrumbs/watt-breadcrumbs.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  encapsulation: ViewEncapsulation.None,
+  template: `<ng-content></ng-content
+    ><watt-icon *ngIf="!isLast" name="right"></watt-icon>`,
+})
+export class WattBreadcrumbComponent {
+  // eslint-disable-next-line @angular-eslint/no-output-native
+  @Output() click: EventEmitter<unknown> = new EventEmitter<unknown>(); // Used to determine if the breadcrumb is interactive or not
+  @HostBinding('class.interactive') get isInteractive() {
+    return this.click.observed;
+  }
```
dzhavat, line 45: question: Interesting. Does this check whether the element has `click` handler attached to it?

libs/ui-watt/src/lib/components/breadcrumbs/watt-breadcrumbs.component.ts
```diff
@@ -0,0 +1,83 @@
...
+ * `import { WattBreadcrumbs } from '@energinet-datahub/watt/breadcrumbs';`
+ */
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+  standalone: true,
+  imports: [CommonModule, WattIconModule],
+  selector: 'watt-breadcrumbs',
+  styleUrls: ['./watt-breadcrumbs.component.scss'],
+  templateUrl: './watt-breadcrumbs.component.html',
```
dzhavat, line 63: nitpick: the template is quite small so it can be inlined as well.

libs/ui-watt/src/lib/components/breadcrumbs/watt-breadcrumbs.component.ts
frontend-specialisten: Yes 👍

libs/ui-watt/src/lib/components/breadcrumbs/watt-breadcrumbs.component.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * @ignore
+   */
+  ngAfterViewInit() {
+    this.breadcrumbs.last.isLast = true;
+  }
+}
+
+export const WattBreadcrumbs = [
```
ManBearTM, line 82: Should we decide on a convention for dependency exports like this? I have previously exported them using uppercase notation like WATT_BREADCRUMBS_DEPS. I don’t have a strong preference either way, would just like to be consistent.

libs/ui-watt/src/lib/components/breadcrumbs/watt-breadcrumbs.component.ts
ManBearTM: Pretty cool!

libs/ui-watt/src/lib/components/breadcrumbs/watt-breadcrumbs.component.ts
dzhavat: There's an example in Angular's docs page 👇 
https://angular.io/guide/standalone-components#standalone-components-for-library-authors

image

They are using all caps too. Let's try it but without the `_DEPS` part. The all caps should be enough to distinguish it from other types of exports. The `as const` part though I think we should add, as it's also a recommendation from Angular.

libs/ui-watt/src/lib/components/breadcrumbs/watt-breadcrumbs.component.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  const getSeperators = () => screen.getAllByRole('img'); // role of watt-icon is img
+  const getBreadcrumbWithRouterLink = () => screen.queryByText('Breadcrumbs');
+  const getBreadcrumbWithClick = () => screen.queryByText('Components');
+  const getNoninteractiveBreadcrumb = () => screen.queryByText('Overview');
+
+  it('renders', async () => {
+    await setup(overviewStory);
+  });
```
dzhavat, line 49: suggestion: this test doesn't add much value and can be removed :)

dzhavat (approved):

PR #1128, opened 2022-11-09
fix: ChargeMessage shows previous message

ShaBloze:
## Description

Problem:
If you select one message and afterwards another, the previous message will be shown until the new one is loaded.

Solution:
Don't show message when it is loading the new one.

## References

- #0000


libs/dh/charges/feature-prices/src/lib/drawer/charge-message/dh-charge-price-message.component.html
```diff
@@ -51,8 +51,18 @@ <h1>{{ message.messageId }}</h1>
...
+  <ng-container
+    *ngIf="!(blobIsDownloading$ | push) && !(messageIsSearching$ | push)"
+  >
+    <ng-container *rxLet="blobContent$ as blobContent">
+      <pre>{{ blobContent }}</pre>
+    </ng-container>
+
+    <div *transloco="let transloco; read: 'messageArchive'">
+      <watt-badge *ngIf="blobHasGeneralError$ | push" type="danger">{{
+        transloco("unknowenError")
```
x-platformcoder, line 63:
```suggestion
        transloco("unknownError")
```

x-platformcoder (dismissed):

x-platformcoder (approved):

PR #1146, opened 2022-11-16
feat: Handle summer/winter in charge price table

ShaBloze:
## Description

Right now it is showing the DateTime based on the Offset. However we want to show it differently.

When switching to normal time in March, then the time intervals shown in the search results must be:
00-01
01-02
02-03 this line is not shown (in danish time) because it doesn't exists.
03-04

When switching to DST in October, then the time intervals shown in the search results must be:
00-01
01-02
02-03
02-03
03-04

## References

- #1851
https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/energinet-datahub/geh-charges/1851


libs/dh/charges/feature-prices/src/lib/drawer/charge-content/price-tab/dh-charges-charge-prices-tab.component.ts
```diff
@@ -196,15 +196,43 @@ export class DhChargesChargePricesTabComponent
...
+
+      return this.getHours(toDateTime, 0);
+    }
+
+    return this.getTime(toDateTime);
+  }
+
+  getTime(date: Date) {
+    const minutes = this.addLeadingZeros(getMinutes(date), 2);
+    const hours = this.getHours(date, 0);
```
dzhavat, line 235: suggestion: I'm sorry for inviting myself here, but there's a better way to format time that is to be shown in the UI and that is by creating an Angular pipe. You can have a look at `dh-date.pipe.ts` or `dh-datetime.pipe.ts` to get a sense of how it works. This, I believe, will also simplify the logic quite a lot. Let me know if you need help with creating a pipe.

dzhavat (approved):

PR #1147, opened 2022-11-16
fix: Charges Drawer date picker: Set date range to include all 24 hours of today

HenrikSommer:
## Description

When drawer is opened to show Charge information, the date range picker dates are set to today at midnight and tonight right before midnight, to ensure all prices of today and all messages of today are included in the search results based on the date range.


libs/dh/charges/feature-prices/src/lib/drawer/charge-content/drawer-datepicker/drawer-datepicker.service.ts
```diff
@@ -26,9 +26,12 @@ export interface DatePickerData {
...
 })
 export class DrawerDatepickerService {
+  todayAtMidnight = new Date(new Date().setHours(0, 0, 0, 0));
+  tonightBeforeMidnight = new Date(new Date().setHours(23, 59, 59, 59));
+
   dateRangeDefault: DatePickerData = {
-    startDate: new Date().toISOString(),
-    endDate: new Date().toISOString(),
+    startDate: this.todayAtMidnight.toISOString(),
+    endDate: this.tonightBeforeMidnight.toISOString(),
```
dzhavat, line 34: Hey @HenrikSommer 
Just noticed the PR title and was tempted to write a comment :)

suggestion: as part of #1108, the datepicker automatically sets the end date to the end of the day. So there's no need to do that (also whatever the initial date, it will also be set to end of day. This can be seen in the PR description).

The start date, however, should be set explicitly to the desired time initially. Date selection afterwards will always set the "from" date to start of day.

HenrikSommer: Oh, that's nice @dzhavat,
I will revert the change for the endDate param :-)

ShaBloze (approved):
LGTM. 

Just a single comment.

ShaBloze (approved):

PR #1154, opened 2022-11-22
Added no result icon

mimse:
Added no result icon

image


libs/ui-watt/src/lib/foundations/icon/icon.service.ts
```diff
@@ -49,6 +49,7 @@ export class WattIconService {
       'custom-primary-info',
       '/assets/ui-watt/icons/primary-info.svg'
     );
+    this.addSvgIcon('custom-noResults', '/assets/ui-watt/icons/noResults.svg');
```
dzhavat, line 52: nitpick: consider using `custom-no-results` (in this and other places) for consistency with the other icons.

mimse: Solved in https://github.com/Energinet-DataHub/greenforce-frontend/pull/1154/commits/8e027aaf7c58d2cea31aa85494ca87e14b44d79f

dzhavat (approved):

Issue #590, opened 2022-06-03
Extend dropdown with "Select all" option


mknic:
**Description**
As an end user I want to be able to select all items in the dropdown list
So that I don't have to click each item one by one

As a platform team we want to extend the dropdown component with a select all property
So that the product teams easily can implement the select all possibilities and give to the end user

**Note**
Follow behaviour as in: https://stackblitz.com/github/bithost-gmbh/ngx-mat-select-search-example?file=src%2Fapp%2Fapp.component.html

**AC**
- [x] Dropdown should include a "Select all" option
- [x] "Select all" should always select all applicable to applied filter
- [x] Dropdown should include a "Deselect all" option (when all results are selected)
- [ ] Update dropdown look in Figma to follow design from link in `Note`

PR #1174, opened 2022-11-29
feat(watt): add support for selecting all options in dropdown

dzhavat:
### Watt Design System
- add support for selecting all options in dropdown
- adjust styling so search input is aligned with the dropdown

demo

## References

- #590


ManBearTM (dismissed):
Looks good, nice solution! Only issue I see is that the `determineToggleAllCheckboxState` function is only triggered after the dropdown is closed, meaning it won't update the state until closed and then reopened. But I'm not sure if that is even possible to fix.

dzhavat:
> Looks good, nice solution! Only issue I see is that the `determineToggleAllCheckboxState` function is only triggered after the dropdown is closed, meaning it won't update the state until closed and then reopened. But I'm not sure if that is even possible to fix.

That happens because the `FormControl` has `{ updateOn: 'blur' }`. Removing triggers `determineToggleAllCheckboxState` as the user selects individual options. I don't remember now why we set it to `blur`. 🤔

ManBearTM (approved):

PR #1181, opened 2022-11-30
feat(dh): use watt table in charges

ManBearTM:
## Description

This replaces MatTable with the new WattTable component in the charges domain.

In addition, this PR makes all components standalone instead of using SCAM modules.


dzhavat (dismissed):
👍 For the changes related to Watt

frontend-specialisten (approved):

PR #1185, opened 2022-11-30
feat(ui-watt): grid layout in shell

ManBearTM:
## Description

This PR makes the sidenav-content area use CSS grid for layout, enabling the outlet components to take up the remaining space of the viewport.

### DIFF
--- a/libs/ui-watt/src/lib/components/shell/shell.component.html
+++ b/libs/ui-watt/src/lib/components/shell/shell.component.html
@@ -37,7 +37,7 @@ limitations under the License.
     </div>
   </mat-sidenav>
 
-  <mat-sidenav-content>
+  <mat-sidenav-content class="watt-sidenav-content">
     <mat-toolbar class="watt-toolbar">
       <watt-button
         *ngIf="isHandset$ | async"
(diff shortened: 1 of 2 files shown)


mimse (approved):

frontend-specialisten (commented):
Suggestion: what about setting `max-height: 100%` and `overflow-y: auto`, at `.watt-main-content`. By doing this we can avoid setting height 100% on everything? 

This means however that we maybe should do something with the `.mat-header-row`, I've only done some minor experiments so far.
Also this change should ofcourse be ensured that it doesn't break anything for EO.

ManBearTM:
> Suggestion: what about setting `max-height: 100%` and `overflow-y: auto`, at `.watt-main-content`. By doing this we can avoid setting height 100% on everything?
> 
> This means however that we maybe should do something with the `.mat-header-row`, I've only done some minor experiments so far. Also this change should ofcourse be ensured that it doesn't break anything for EO.

Good suggestion, I can try to do some experimenting. I'm not sure it is sufficient in the cases where we want to "delegate" the scroll to inside the table component, but it could be a nice default behavior to have.

PR #1186, opened 2022-12-01
Remove rendering bug for paginator in Certificates page

MadsThy:
## Description

This removes the bug from paginator on certificates page. It does not fix the underlying issue though, but that is for another day :)


NielsJorck (approved):
LGTM

PR #1216, opened 2022-12-08
feat(ui-watt): no enforcing of boundaries for type imports

ManBearTM:
## Description

This PR makes it possible to import types from watt regardless of current package scope/tag, avoiding the need to disable the eslint rule.


frontend-specialisten (approved):
Nice!

PR #1219, opened 2022-12-09
Renamed for shorter names

mimse:
Renamed files to make them more precise with less noise


libs/dh/admin/user-management/jest.config.ts
```diff
@@ -16,7 +16,7 @@
  */
 /* eslint-disable */
 export default {
-  displayName: 'dh-admin-feature-user-management',
+  displayName: 'dh-admin-user-management',
```
dzhavat, line 19: suggestion: let's leave these as before. They are auto generated and follow a specific pattern where the `feature` part indicates that type of library. So it's going to be a lot of manual editing if this should be different.

mimse: Should the path be reverted as well, or just the properties value

dzhavat: Pretty much everything outside of a `lib` library is auto-generated. So it's better to stay as is. So I'd say that our focus should primarily be on the files inside `lib`.

PR #1220, opened 2022-12-09
fix(watt): fix change detection error

dzhavat:
### Watt Design System
- fix `ExpressionChangedAfterItHasBeenCheckedError` error that appears when tabs component is used
- rename files 

## References

- N/A


ManBearTM (approved):
🥇

PR #1228, opened 2022-12-13
feat(dh): add table to users tab

dzhavat:
### DataHub
- add table to users tab

Note: Content in the "Status" column will be added in a follow up PR.

image

## References


libs/dh/admin/feature-user-management/src/lib/user-management-tabs/users-tab/dh-users-tab-table.component.html
```diff
@@ -0,0 +1,41 @@
...
+limitations under the License.
+-->
+<watt-table
+  #table
+  [columns]="columns"
+  [dataSource]="dataSource"
+  sortBy="name"
+  sortDirection="asc"
+  [sortClear]="false"
+  [resolveHeader]="translateHeader"
```
mimse, line 24: Nikpick: It is possible to use transloco directly as a function to resolveHeader

mimse (approved):
Goods look, just one comment

libs/dh/admin/feature-user-management/src/lib/user-management-tabs/users-tab/dh-users-tab-table.component.html
dzhavat: Yes, I saw it used like that in another template. While possible, this means that the keys in the translation files must be the same as the column names. This works well sometimes, but in this case two of the column names have different values, so the custom function was necessary.

PR #1230, opened 2022-12-13
fix: No permission for multiple users.

defectiveAi:


libs/dh/shared/feature-authorization/src/lib/actor-token.service.ts
```diff
@@ -15,48 +15,118 @@
...
           return this.tokenHttp
-            .v1TokenPost({
-              actorId: r.actorIds[0],
-              externalToken: externalToken,
-            })
-            .pipe(
-              tap((r) => (this._internalToken = r.token)),
-              map((r) => r.token)
-            );
+            .v1TokenPost(r.actorIds[0])
```
dzhavat, line 85: question: is the first id always the right one? If yes, is there a reason for returning an array from the endpoint?

libs/dh/shared/feature-authorization/src/lib/actor-token.service.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    // assert
+    expect(actual).toEqual(false);
+  });
+
+  test('should return false if request is not for internal authentication flow', async () => {
+    // arrange
+    const request = new HttpRequest<string>(
+      'GET',
+      'https://dev002datahubb2c.b2clogin.com/dev002datahubb2c.onmicrosoft.com/b2c_some_policy/oauth2/v2.0/token'
```
dzhavat, line 99: nitpick: the `isPartOfAuthFlow` checks whether the request url ends with a specific value. The origin is not checked. So in theory if `dev002datahubb2c.b2clogin.com` had an endpoint that ends with `/v1/Token` (or the other endpoint), the `isPartOfAuthFlow` would return `true`. Though veeery unlikely to happen. :)

libs/dh/shared/feature-authorization/src/lib/actor-token.service.ts
defectiveAi: Yes, it is for now. Next feature (Feature 3) will add the drop-up to allow for changing this value.

libs/dh/shared/feature-authorization/src/lib/permission.service.ts
```diff
@@ -32,7 +30,6 @@ export class PermissionService {
   constructor(
     @Inject(dhB2CEnvironmentToken) private config: DhB2CEnvironment,
```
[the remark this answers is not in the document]
defectiveAi: Good catch, fixed.

libs/dh/shared/feature-authorization/src/lib/actor-token.service.spec.ts
defectiveAi: We have a test (should return false if request is not for internal authentication flow) checking that `b2clogin.com`-url is rejected. But the code is the second line of defense, as the endpoint is sent without Bearer header. So it is currently impossible to reach `isPartOfAuthFlow`.

dzhavat (approved):

PR #1236, opened 2022-12-14
If value is null, set it to the initial value

mimse:
## Description

When using reset on the FormGroup, it would try to set the range or single value to null. It now sets it to the initial value it was created with.

## References
https://github.com/Energinet-DataHub/geh-post-office/issues/828


libs/ui-watt/src/lib/components/input/shared/watt-picker-base.ts
```diff
@@ -140,13 +140,13 @@ export abstract class WattPickerBase
...
       this.setRangeValue(
-        value as WattRange,
+        (value ?? this.initialValue) as WattRange,
         this.startInput.nativeElement,
         this.endInput.nativeElement
       );
     } else {
       this.setSingleValue(
-        value as Exclude<WattPickerValue, WattRange>,
+        (value ?? this.initialValue) as Exclude<WattPickerValue, WattRange>,
```
dzhavat, line 149: suggestion: nice change. Can we add a unit test verifying the behaviour in case someone decides to poke with for whatever reason it at some point.

libs/dh/message-archive/feature-log-search/src/lib/dh-message-archive-log-search.component.ts
```diff
@@ -232,7 +235,6 @@ export class DhMessageArchiveLogSearchComponent {
 
   resetSearchCritera() {
     this.store.resetState();
-    //TODO: reset form not working
-    this.searchForm.reset();
+    this.searchForm = buildSearchForm();
```
dzhavat, line 238: issue: interesting solution. This doesn't really reset the `FormGroup` assigned to `searchForm` initially but rather creates a new `FormGroup` instance with initial values. It works but it feels more like a workaround.

I took a look at the documentation for [FormControl](https://angular.io/api/forms/FormControl#formcontrol) and found something interesting. When creating a new `FormControl`, an extra second param object can be passed that takes a few values. One of them is ` nonNullable`.

By looking at it, it seems like adding it will make the form control reset to its initial value on reset and not `null`, as we saw a few days ago.

image

They also have [an example in the docs](https://angular.io/api/forms/FormControl#reset-the-control-to-its-initial-value):

image

They also have an example for how to reset a FormControl to specific value:

image

So maybe we don't need the adjustments to our component. What we need to do instead is make sure we set `{nonNullable: true}` on those FormControls that have an initial value different than `null`.

mimse: image

Array gets set to null even though it is set to nonNullable: true

image

mimse:
message archive is not in scope 1, and going to be deleted

PR #1238, opened 2022-12-15
Feat/block failed batch

frontend-specialisten:
## Description

- Add disable state of the table (WATT)
- Disable the grid areas table when a batch has failed (DH)
- Ensure calculation steps won't be accessible when the batch has failed. (DH)
- Adjust wholesale mocks, so endpoints use the same dataset. (DH)

### DIFF
--- a/libs/ui-watt/src/lib/components/table/watt-table.component.html
+++ b/libs/ui-watt/src/lib/components/table/watt-table.component.html
@@ -80,10 +80,10 @@ limitations under the License.
     mat-row
     *matRowDef="let row; columns: _getColumns()"
     [attr.aria-selected]="row === activeRow"
-    (click)="rowClick.emit(row)"
+    (click)="disabled ? null : rowClick.emit(row)"
     [ngClass]="{
-      'watt-table-highlight-row': !suppressRowHoverHighlight,
-      'watt-table-clickable-row': rowClick.observed,
+      'watt-table-highlight-row': !disabled || !suppressRowHoverHighlight,
+      'watt-table-clickable-row': !disabled && rowClick.observed,
       'watt-table-active-row': _isActiveRow(row)
     }"
   ></tr>
--- a/libs/ui-watt/src/lib/components/table/watt-table.component.html
+++ b/libs/ui-watt/src/lib/components/table/watt-table.component.html
@@ -80,9 +80,9 @@ limitations under the License.
     mat-row
     *matRowDef="let row; columns: _getColumns()"
     [attr.aria-selected]="row === activeRow"
-    (click)="disabled ? null : rowClick.emit(row)"
+    (click)="_onRowClick(row)"
     [ngClass]="{
-      'watt-table-highlight-row': !disabled || !suppressRowHoverHighlight,
+      'watt-table-highlight-row': !disabled && !suppressRowHoverHighlight,
       'watt-table-clickable-row': !disabled && rowClick.observed,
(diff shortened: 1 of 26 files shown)


libs/ui-watt/src/lib/components/table/watt-table.component.html
```diff
@@ -80,10 +80,10 @@
     mat-row
     *matRowDef="let row; columns: _getColumns()"
     [attr.aria-selected]="row === activeRow"
-    (click)="rowClick.emit(row)"
+    (click)="disabled ? null : rowClick.emit(row)"
     [ngClass]="{
-      'watt-table-highlight-row': !suppressRowHoverHighlight,
-      'watt-table-clickable-row': rowClick.observed,
+      'watt-table-highlight-row': !disabled || !suppressRowHoverHighlight,
```
ManBearTM, line 85: question: shouldn't this be `&&`?

libs/ui-watt/src/lib/components/table/watt-table.component.html
```diff
@@ -80,10 +80,10 @@
     mat-row
     *matRowDef="let row; columns: _getColumns()"
     [attr.aria-selected]="row === activeRow"
-    (click)="rowClick.emit(row)"
+    (click)="disabled ? null : rowClick.emit(row)"
```
ManBearTM, line 83: suggestion: move this to class method, something like `_onRowClick`, seems a bit wrong to return null here IMO.

libs/ui-watt/src/lib/components/table/watt-table.component.ts
```diff
@@ -159,6 +160,26 @@ export class WattTableComponent<T>
...
+  get disabled(): boolean {
+    return this._disabled;
+  }
+
+  /**
+   * Used for disabling the table. This will disable all user interaction
+   */
+  set disabled(value: BooleanInput) {
+    this._disabled = coerceBooleanProperty(value);
+  }
```
ManBearTM, line 181: question: is it neccessary to coerce here as opposed to a simple boolean?

libs/ui-watt/src/lib/components/table/watt-table.component.html
```diff
@@ -80,10 +80,10 @@
     mat-row
     *matRowDef="let row; columns: _getColumns()"
     [attr.aria-selected]="row === activeRow"
-    (click)="rowClick.emit(row)"
+    (click)="_onRowClick(row)"
```
dzhavat, line 83: nitpick: probably something we can't do much about right now, but even though the click doesn't have a visual effect, change detection will still be triggered every time there's a click on a row. That's most likely not an issue when the visual cues, like mouse pointer or row highlight, are not there, but still something to be aware of.

The best, of course, is if it was possible to not set the click in the first place, but I'm not sure there's an easy solution for that. :)

dzhavat: What do you mean by adding it manually?

ManBearTM: I feel like this is an optimization we can look into **if** it becomes relevant. But it is a good observation!

ManBearTM (approved):

libs/ui-watt/src/lib/components/table/watt-table.component.html
dzhavat: Yes, for sure. Nothing

PR #1245, opened 2022-12-16
feat(dh): add page number and page size as params to request

dzhavat:
### DataHub
- add page number and page size as params to request
- rename column names in user management page

## References

### DIFF
--- a/libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
+++ b/libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 import { Injectable } from '@angular/core';
-import { filter, map, Observable, switchMap, tap } from 'rxjs';
+import { filter, map, Observable, switchMap, tap, withLatestFrom } from 'rxjs';
 import { ComponentStore, tapResponse } from '@ngrx/component-store';
 
 import {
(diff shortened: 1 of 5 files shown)


libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
```diff
@@ -60,26 +64,29 @@ export class DhAdminUserManagementDataAccessApiStore extends ComponentStore<DhUs
...
-              this.setLoading(LoadingState.LOADED);
-
-              this.updateUsers(users);
-            },
-            () => {
-              this.setLoading(LoadingState.LOADED);
-
-              this.handleError();
-            }
+      switchMap(([, state]) =>
```
mimse, line 73: Q: ever seen [ before for unused variable. Normally it is _ any history to this conversion?

mimse (approved):
Looks good, just one question :-)

libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
dzhavat: :) Well. the `[` is the opening square bracket of the array. It's the `,` part that is the magic. It allows us to skip the first element and not assign it to anything since it's not going to be used. Then the second element is assigned to `state`.

dzhavat: So it's array destructuring but ignoring elements at the same time.

image

PR #1261, opened 2022-12-21
Conditional add project reference if exists, to make a poor man monor…

mimse:
## Description

To make it a bit easier to work with both frontend and backend at the same time. I have added a backend directory to the solution with readme to explain how to add more projects.

All the projects clone to this backend directory will be ignored by git so they will be commit to the monorepo, but we will still be able to make a relative reference from the BFF to the projects.

This has been done with a condition, so if the project exists in the backend directory it will use a project reference, or else it will use the published nuget package as before instead.

I have been using this setup for a while now, and I think it hits a pretty good balance for a good DX. 

Hope this can help others as well :-)

Launch all dotnet project from one VSCode

billede


ManBearTM (approved):
💯

ManBearTM (approved):

PR #1262, opened 2022-12-21
feat(dh): add pagination to user management page

dzhavat:
### DataHub
- add pagination to user management page
- use Mat Table instead of Watt Table with because Watt Table doesn't support custom data source
- add `pageIndex` Input to paginator component

image

## References


libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
```diff
@@ -57,6 +57,11 @@ export class DhAdminUserManagementDataAccessApiStore extends ComponentStore<DhUs
   users$ = this.select((state) => state.users);
   totalUserCount$ = this.select((state) => state.totalUserCount);
 
+  // 1 needs to be substracted here because our endpoint's `pageNumber` param starts at `1`
+  // whereas the paginator's `pageIndex` property starts at `0`
+  paginatorPageIndex$ = this.select((state) => state.pageNumber - 1);
```
ManBearTM, line 62: question: seems to me like we should either make the endpoint 0-based or perhaps add a `pageNumber` input to the new `watt-paginator` to make the component 1-based instead. what do you think?

libs/dh/admin/feature-user-management/src/lib/user-management-tabs/users-tab/dh-users-tab-table.component.html
```diff
@@ -14,28 +14,36 @@
...
-
-  <ng-container *wattTableCell="columns['phone']; let element">{{
-    element.phoneNumber | dhEmDashFallback
-  }}</ng-container>
-
-  <ng-container *wattTableCell="columns['status']; let element">{{
-    undefined | dhEmDashFallback
-  }}</ng-container>
-</watt-table>
+  <mat-table [dataSource]="dataSource">
```
ManBearTM, line 20: This is unacceptable! 🙃

libs/dh/admin/data-access-api/src/lib/dh-custom-data-source.ts
```diff
@@ -0,0 +1,34 @@
...
+  private readonly store = inject(DhAdminUserManagementDataAccessApiStore);
+
+  connect() {
+    return this.store.users$;
+  }
+
+  disconnect(): void {
+    // Intentionally left empty.
+  }
+}
```
ManBearTM, line 34: suggestion: perhaps follow the convention from the material library by adding a paginator accessor and then control pagination from the data source instead?

libs/dh/admin/feature-user-management/src/lib/user-management-tabs/users-tab/dh-users-tab.component.scss
```diff
@@ -16,8 +16,16 @@
  */
 :host {
   display: block;
+  // TODO: Add spacing variable for 24px
```
ManBearTM, line 19: `--watt-spacing-xm` (extra medium)? Or perhaps `--watt-spacing-sl` (small large)? 😄

ManBearTM (commented):
I'm a little sad that this was not made for use with `watt-table`, feels a bit incomplete because of that. But if it is important to get something out now, then I think it is a good first step.

libs/dh/admin/feature-user-management/src/lib/user-management-tabs/users-tab/dh-users-tab.component.scss
dzhavat: Yes, something creative :)

libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
dzhavat: I've added a `pageIndex` input to the `DhSharedUiPaginatorComponent` component (as seen in the last file in this PR). We should definitely add something similar to `watt-paginator`. But since `watt-paginator` is still based on Mat Paginator, I doubt we can simply add a `pageNumber`/`pageIndex` property that is 1-based. The conversion from 1 to 0 should still happen somewhere. It can be internal to `watt-paginator` though.

Without thinking about it too deeply, I'd say 1-based pagination feels more natural. Mat Paginator is 0-based probably because of something technical. Hopefully no one says "Give me page 0" 😄

libs/dh/admin/feature-user-management/src/lib/user-management-tabs/users-tab/dh-users-tab-table.component.html
dzhavat: The minute `watt-table` supports custom data source, I'll convert this table to use our component. Promised 😄

libs/dh/admin/data-access-api/src/lib/dh-custom-data-source.ts
dzhavat: Do you have an example I can look at? I'm not sure I can follow the suggestion.

ManBearTM: https://github.com/angular/components/blob/main/src/material/table/table-data-source.ts#L131

ManBearTM: Yeah, the implementation obviously needs to be different. I was imagining it would work much the same as you are doing now, but instead of subscribing to the PageEvent in your component, you would subscribe to it in the DataSource - and then just update the page metadata from within the DataSource.

ManBearTM: Probably not much to gain from that right now, but it would make it easier to transition into a more generic and reusable DataSource that could be used in other areas as well. I imagine server side pagination is going to be needed in multiple places in the future.

dzhavat: As it is right now, the data source is destroyed and recreated every time a new page data has loaded. That is because the table and paginator are in a component controlled by a `isLoading` flag used to show a spinner. So the entire page content is removed, a spinner is shown, and the recreated again once a request completes.

So in order for this to work, the loading spinner probably needs to be all the way next to the table, so only the table is hidden when fetching data. This way the paginator will always be visible on the page. That's a detail I haven't worked out yet.

dzhavat: But you're right that server side pagination is going to be needed in other places as well. So that's a use case that must be taken into consideration.

ManBearTM: That makes it harder for sure 😄 Doesn’t sound right to rebuild everything when changing pages. Perhaps you should have two separate loading states, one for loading the table initially and another for when changing the page?

dzhavat: Two loading states is probably is going to be the best solution.

ManBearTM (approved):

libs/dh/admin/data-access-api/src/lib/dh-custom-data-source.ts
dzhavat: Looked quickly at a two spinner solution, and it doesn't seem as promising. Hiding the table behind an `ngIf` in any way will mean that the related data source will be destroyed and recreated again. Couldn't find a way to show a spinner from within a table and still keep the table's elements (like column names. etc.). So I'm back to the drawing board.

PR #1263, opened 2022-12-21
feat(ui-watt): watt paginator

ManBearTM:
## Description

This PR adds a new `watt-paginator` component for use below tables. Actually implementing it in DataHub will come in a later PR.

It should replace `dh-shared-ui-paginator`, since translations are now handled differently.


libs/ui-watt/src/lib/components/paginator/watt-paginator.component.ts
```diff
@@ -0,0 +1,112 @@
...
+  @Input() pageSize = 50;
+  @Output() changed = new EventEmitter<PageEvent>();
+
+  @ViewChild(MatPaginator, { static: true }) instance!: MatPaginator;
+
+  description?: string;
+
+  private intl = inject(WattPaginatorIntlService);
+  private matPaginatorIntl = inject(MatPaginatorIntl);
+  private subscription!: Subscription;
```
dzhavat, line 73: nitpick: instead of using non-null, let's make the property optional  (i.e. `subscription?: Subscription`) and use the `?` mark again when unsubscribing.

libs/ui-watt/src/lib/components/paginator/watt-paginator.component.ts
```diff
@@ -0,0 +1,112 @@
...
+      [pageSizeOptions]="pageSizeOptions"
+      [showFirstLastButtons]="true"
+      [attr.aria-label]="description"
+    ></mat-paginator>
+  `,
+})
+export class WattPaginatorComponent implements OnInit, OnDestroy {
+  @Input() length = 0;
+  @Input() pageSizeOptions = [50, 100, 150, 200, 250];
+  @Input() pageSize = 50;
```
dzhavat, line 64: suggestion: consider adding `pageIndex` as well in order to use the component with lazy-loaded table data.

libs/dh/globalization/ui-watt-translation/src/lib/dh-globalization-ui-watt-translation.module.ts
```diff
@@ -39,6 +40,26 @@ export class DhWattClipboardIntlService extends WattClipboardIntlService {
   }
 }
 
+@Injectable()
+export class DhWattPaginatorIntlService extends WattPaginatorIntlService {
```
dzhavat, line 44: nitpick: rename to `DhPaginatorIntlService`, also the clipboard service above.

libs/ui-watt/src/lib/components/table/watt-table-data-source.ts
```diff
@@ -14,9 +14,93 @@
...
+    filterPredicate: MatTableDataSource<T>['filterPredicate']
+  ) {
+    this.dataSource.filterPredicate;
+  }
+
+  get sortData() {
+    return this.dataSource.sortData;
+  }
+  set sortData(sortData: MatTableDataSource<T>['sortData']) {
+    this.dataSource.sortData;
```
dzhavat, line 87: question: shouldn't `sortData` be asigned to `this.dataSource.sortData`?

libs/ui-watt/src/lib/components/table/watt-table-data-source.ts
```diff
@@ -14,9 +14,93 @@
...
+    this.dataSource.sort = sort;
+  }
+
+  get filterPredicate() {
+    return this.dataSource.filterPredicate;
+  }
+  set filterPredicate(
+    filterPredicate: MatTableDataSource<T>['filterPredicate']
+  ) {
+    this.dataSource.filterPredicate;
```
dzhavat, line 80: question: shouldn't `filterPredicate` be asigned to `this.dataSource.filterPredicate`?

libs/ui-watt/src/lib/components/table/watt-table-data-source.ts
ManBearTM: Whoops, good catch!

libs/ui-watt/src/lib/components/table/watt-table-data-source.ts
ManBearTM: Whoops, good catch!

libs/ui-watt/src/lib/components/paginator/watt-paginator.component.ts
ManBearTM: Fair enough 🙂

libs/dh/globalization/ui-watt-translation/src/lib/dh-globalization-ui-watt-translation.module.ts
ManBearTM: Done!

libs/ui-watt/src/lib/components/paginator/watt-paginator.component.ts
ManBearTM: I think I'd rather not commit to a specific interface until we have decided if it should be 0-based or 1-based. Let's do it as part of integrating with your custom data source, then we might have a better idea of which solution to go with.

dzhavat: Good point. Let take it at that time.

libs/ui-watt/src/lib/components/table/watt-table-data-source.ts
```diff
@@ -14,9 +14,93 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { DataSource } from '@angular/cdk/collections';
+import { MatPaginator } from '@angular/material/paginator';
+import { MatSort } from '@angular/material/sort';
```
dzhavat, line 19: question: the linting error is due to this. Is it supposed it be used?

dzhavat (approved):

PR #1274, opened 2023-01-03
Update package and generate API for audit log.

defectiveAi:


apps/dh/api-dh/source/DataHub.WebApi/Controllers/MarketParticipant/Dto/UserAuditLogsDto.cs
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+
+namespace Energinet.DataHub.WebApi.Controllers.MarketParticipant.Dto
+{
+    public sealed record UserAuditLogsDto(IEnumerable<UserRoleAssignmentAuditLogDto> RoleAssignments);
```
FirestarJes, line 19: nitpick: variable name should probably include Audit somewhere

apps/dh/api-dh/source/DataHub.WebApi/Controllers/MarketParticipantUserController.cs
```diff
@@ -51,17 +55,46 @@ public Task<ActionResult<GetAssociatedUserActorsResponseDto>> GetUserActorsAsync
...
+            {
+                var auditLogs = await _marketParticipantClient
+                    .GetUserAuditLogsAsync(userId)
+                    .ConfigureAwait(false);
+
+                var roleAssignmentAuditLogs = new List<UserRoleAssignmentAuditLogDto>();
+
+                foreach (var auditLog in auditLogs.UserRoleAssignmentAuditLogs)
+                {
+                    var userDto = await _marketParticipantClient
```
FirestarJes, line 78: We should consider making a function that can fetch multiple user in a single call

apps/dh/api-dh/source/DataHub.WebApi/Controllers/MarketParticipantUserController.cs
```diff
@@ -51,17 +55,46 @@ public Task<ActionResult<GetAssociatedUserActorsResponseDto>> GetUserActorsAsync
...
+                        .GetUserAsync(auditLog.ChangedByUserId)
+                        .ConfigureAwait(false);
+
+                    var userRoleDto = await _marketParticipantUserRoleClient
```
FirestarJes, line 82: The same here, we should consider whether it is worth it to make a function to fetch multiple roles in a single call

FirestarJes (changes requested):

apps/dh/api-dh/source/DataHub.WebApi/Controllers/MarketParticipant/Dto/UserAuditLogsDto.cs
defectiveAi: Fixed.

apps/dh/api-dh/source/DataHub.WebApi/Controllers/MarketParticipantUserController.cs
defectiveAi: In case of perfromance issues, we should add query APIs for this, yes. But in this case, the list will always be short, so I see no reason to optimise at the moment.

apps/dh/api-dh/source/DataHub.WebApi/Controllers/MarketParticipantUserController.cs
defectiveAi: In case of perfromance issues, we should add query APIs for this, yes. But in this case, the list will always be short, so I see no reason to optimise at the moment.

apps/dh/api-dh/source/DataHub.WebApi/Controllers/MarketParticipantUserController.cs
FirestarJes: ok :)

apps/dh/api-dh/source/DataHub.WebApi/Controllers/MarketParticipantUserController.cs
FirestarJes: Ok :)

FirestarJes (approved):

PR #1296, opened 2023-01-10
Run batch for grid areas provided by market participants

frontend-specialisten:
## References

- [#627](https://github.com/Energinet-DataHub/opengeh-wholesale/issues/627)


libs/dh/wholesale/domain/src/grid-areas.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+import { DateRange, GridAreaDto } from '@energinet-datahub/dh/shared/domain';
+import { isAfter, isBefore, isEqual } from 'date-fns';
+
+export function filterValidGridAreas(
+  gridAreas: GridAreaDto[],
+  dateRange: DateRange | null
+): GridAreaDto[] {
+  if (dateRange === null) return gridAreas;
+  return gridAreas?.filter((gridArea) => {
```
ManBearTM, line 25: nitpick: the function definition says `gridAreas` cannot be empty, so this shouldn't be neccessary:
```suggestion
  return gridAreas.filter((gridArea) => {
```

libs/dh/wholesale/domain/src/grid-areas.ts
```diff
@@ -0,0 +1,38 @@
...
+    const { validTo, validFrom } = gridArea;
+    const isValidTo =
+      isBefore(new Date(validFrom), new Date(dateRange.end)) ||
+      isEqual(new Date(validFrom), new Date(dateRange.end));
```
ManBearTM, line 29: question: nothing wrong with this, but doesn´t `new Date(validFrom) <= new Date(dateRange.end)` do the same thing?

libs/dh/wholesale/domain/src/grid-areas.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    const isValidFrom = !validTo
```
ManBearTM, line 31: question: am I just confused or is the naming backwards? 😄 To me it seems like this variable should be named `isValidTo`.

libs/dh/wholesale/feature-start/src/lib/dh-wholesale-start.component.ts
```diff
@@ -142,6 +148,46 @@ export class DhWholesaleStartComponent implements OnInit, OnDestroy {
...
+  private mapGridAreasToDropdownOptions(
+    gridAreas: GridAreaDto[]
+  ): WattDropdownOption[] {
+    return (
+      gridAreas?.map((gridArea) => {
+        return {
+          displayValue: `${gridArea?.name} (${gridArea?.code})`,
+          value: gridArea?.code,
+        };
+      }) || []
```
ManBearTM, line 160: nitpick: again, function definition says it is never undefined
```suggestion
      gridAreas.map((gridArea) => {
        return {
          displayValue: `${gridArea?.name} (${gridArea?.code})`,
          value: gridArea?.code,
        };
      })
```

libs/dh/wholesale/domain/src/grid-areas.ts
frontend-specialisten: `new Date(validFrom).getTime() <= new Date(dateRange.end).getTime()` would do the same thing, you can see the implementation [here](https://github.com/date-fns/date-fns/blob/main/src/isBefore/index.ts). If you prefer I can easily change it to that. I was just thinking that it provided more readability since we're using date-fns. Pro is of course that we can do it in a one liner instead of both having to use `isBefore` and `isEqual`.

frontend-specialisten: I've removed the usage of date-fns now :)

ManBearTM (approved):

PR #1297, opened 2023-01-10
feat(dh): use watt-paginator for wholesale

ManBearTM:


libs/ui-watt/src/lib/components/table/watt-table.component.scss
```diff
@@ -23,6 +23,7 @@ watt-table .mat-table {
   grid-auto-rows: max-content;
   height: 100%;
   overflow: auto;
+  isolation: isolate;
```
frontend-specialisten, line 26: Praise: Cool, I didn't know this property

libs/dh/wholesale/feature-calculation-steps/src/lib/time-series-points/dh-wholesale-time-series-points.component.scss
```diff
@@ -19,8 +19,18 @@
   width: 100%;
 }
 
+watt-card {
```
frontend-specialisten, line 22: Question: could we do some context-specific styling with ex. `:host-context('something something')` for the card, so we don't have to do this in multiple places?

frontend-specialisten (approved):

libs/ui-watt/src/lib/components/table/watt-table.component.scss
ManBearTM: Neither did I 😄 Much better than the old hacks to force a new stacking context!

libs/dh/wholesale/feature-calculation-steps/src/lib/time-series-points/dh-wholesale-time-series-points.component.scss
ManBearTM: Oh cool, wasn't aware of `:host-context`, good idea! Will check it out 🙂

PR #1300, opened 2023-01-11
feat(watt): make it possible to disable the dropdown

frontend-specialisten:
## Description

Make it possible to disable the dropdown

image


libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.ts
```diff
@@ -122,6 +122,11 @@ export class WattDropdownComponent
    */
   @ViewChild('matSelect', { static: true }) matSelect?: MatSelect;
 
+  /**
+   * Wether the dropdown is disabled
```
ManBearTM, line 126:
```suggestion
   * Whether the dropdown is disabled
```

ManBearTM (dismissed):
👏

ManBearTM (approved):

PR #1304, opened 2023-01-11
user roles list filter and refactor

xnetsrak:
## Description

user roles list filter and refactor
## References

- #0000


mimse:
image
Getting this error

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab-list-filter.component.ts
```diff
@@ -0,0 +1,92 @@
...
+  ).map((key) => ({
+    displayValue: key,
+    value: key,
+  }));
+
+  constructor() {
+    this.init();
+  }
+
+  init = () => {
```
mimse, line 83: Suggestion: maybe use OnInit interface?

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab.component.scss
```diff
@@ -16,10 +16,21 @@
  */
 :host {
   display: block;
-  // TODO: Add spacing variable for 24px
   margin: 24px var(--watt-space-s);
 }
 
+.user-roles {
+  &__spinner {
```
mimse, line 23: Question: Haven't seen the conversion with __ in class names. Do we use this elsewhere?

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab.component.ts
```diff
@@ -14,30 +14,65 @@
...
 export class DhUserRolesTabComponent {
-  @Input() roles: UserRoleInfoDto[] = [];
+  private readonly store = inject(DhAdminUserRolesManagementDataAccessApiStore);
+  roles$ = this.store.rolesFiltered$;
+
+  isLoading$ = this.store.isLoading$;
+  hasGeneralError$ = this.store.hasGeneralError$;
+
+  updateFilterStatus(status: string | null) {
+    const statusParsed = status as UserRoleStatus;
```
mimse, line 66: Question: Are they save casts?

mimse (commented):
Looks good, have some suggestions and questions :-)

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab.component.scss
xnetsrak: yeah two other places, it is copy paste

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab.component.ts
xnetsrak: what do you mean ? yes it works with values and null.

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab.component.scss
dzhavat: For context, this class name is following [BEM](https://css-tricks.com/bem-101/) methodology. We are not really consistent with BEM, but it's being used in quite a few places.

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab.component.ts
dzhavat: An issue here is that we can get `null` from the dropdown that is also cast to `UserRoleStatus`. Maybe we should send the values as-is and have another way for handling `null` values in the store.

dzhavat (approved):

mimse (approved):

PR #1335, opened 2023-01-19
feat(dh): wholesale consumption per energy supplier

ManBearTM:
## Description

This PR adds a new "step" to the calculation-steps screen, where the user can view "Aggregated non-profiled consumption per energy supplier".

## References

- https://github.com/Energinet-DataHub/opengeh-wholesale/issues/199
- https://github.com/Energinet-DataHub/opengeh-wholesale/pull/727


apps/dh/api-dh/source/DataHub.WebApi/Controllers/WholesaleBatchController.cs
```diff
@@ -120,10 +120,20 @@ public async Task<ActionResult<BatchDto>> GetBatchAsync(Guid batchId)
...
-            var dto = await _client.GetProcessStepResultAsync(processStepResultRequestDto).ConfigureAwait(false);
+            var dto = await _client.GetProcessStepResultAsync(processResultRequestDto).ConfigureAwait(false);
+            return Ok(dto);
+        }
+
+        /// <summary>
+        /// Get a list of actors.
+        /// </summary>
+        [HttpPost("Actors")]
+        public async Task<ActionResult<WholesaleActorDto[]>> GetAsync(ProcessStepActorsRequest batchActorsRequestDto)
```
ManBearTM, line 134: nitpick: rename `batchActorsRequestDto` to `processStepActorsRequest`

libs/dh/shared/domain/src/lib/generated/v1/model/process-step-type.ts
```diff
@@ -11,9 +11,9 @@
  */
 
 
-export type ProcessStepType = 'AggregateProductionPerGridArea';
+export type ProcessStepType = 'AggregateConsumptionPerEnergySupplier';
 
 export const ProcessStepType = {
-    AggregateProductionPerGridArea: 'AggregateProductionPerGridArea' as ProcessStepType
+    AggregateConsumptionPerEnergySupplier: 'AggregateConsumptionPerEnergySupplier' as ProcessStepType
 };
```
ManBearTM, line 19: issue: this file seems out of sync. it shouldn't be needed anymore, but we probably should generate the types again to see if it updates/dissappears.

libs/dh/wholesale/feature-calculation-steps/src/lib/dh-wholesale-calculation-steps.component.ts
```diff
@@ -89,14 +96,61 @@ export class DhWholesaleCalculationStepsComponent {
...
-    this.router.navigate([step], { relativeTo: this.route });
-    this.drawer.open();
+  private getTimeSeriesType(step?: string) {
+    switch (step) {
+      case '1':
+        return TimeSeriesType.Production;
+      case '2':
+        return TimeSeriesType.NonProfiledConsumption;
+      default:
+        return TimeSeriesType.Production;
```
ManBearTM, line 152: question: should we add FlexConsumption to the list, just to be forwards-compatible?

libs/dh/wholesale/feature-calculation-steps/src/lib/energy-suppliers/dh-wholesale-energy-suppliers.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  getSelectedEnergySupplier(energySuppliers?: WholesaleActorDto[]) {
+    if (!energySuppliers) return;
+    return energySuppliers.find(
+      (e) => e.gln === this.route.firstChild?.snapshot.url?.[1]?.path
+    );
+  }
+
+  isSelectedEnerygySupplier(
```
ManBearTM, line 88: typo: should be `isSelectedEnergySupplier`

libs/dh/wholesale/feature-calculation-steps/src/lib/energy-suppliers/dh-wholesale-energy-suppliers.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.store.getEnergySuppliersForConsumption({
+      batchId: this.batchId,
+      gridAreaCode: this.gridAreaCode,
+    });
+  }
+
+  getSelectedEnergySupplier(energySuppliers?: WholesaleActorDto[]) {
+    if (!energySuppliers) return;
+    return energySuppliers.find(
+      (e) => e.gln === this.route.firstChild?.snapshot.url?.[1]?.path
```
ManBearTM, line 84: question: is it possible to retrieve the gln from the route using `parameters` instead of "segments"? E.g:
```ts
this.route.firstChild?.snapshot.params['gln'];
```

libs/dh/wholesale/feature-calculation-steps/src/lib/steps/consumption-per-energy-supplier.component.scss
```diff
@@ -14,3 +14,19 @@
...
+  // TODO: We don't have any spacing variable for 24px!
+  margin-bottom: 24px;
+  padding-bottom: 24px;
+  border-bottom: 1px solid var(--watt-color-neutral-grey-300);
+}
+
+watt-spinner {
+  display: flex;
+  justify-content: center;
+}
```
ManBearTM, line 32: question: there is some duplication in these two steps, which could advocate for creating some shared components to help with this. but i figured this would be easier to do once we have a few more steps, as otherwise we are probably not going to get the abstraction right in the first place. do you agree? 🙂

frontend-specialisten: Yea, I agree, but this is something we should look into for the next step to avoid inconsistency across the drawers

libs/dh/wholesale/feature-calculation-steps/src/lib/energy-suppliers/dh-wholesale-energy-suppliers.component.ts
frontend-specialisten: Good point; that's much more readable!

apps/dh/api-dh/source/DataHub.WebApi/Controllers/WholesaleBatchController.cs
```diff
@@ -120,10 +120,20 @@ public async Task<ActionResult<BatchDto>> GetBatchAsync(Guid batchId)
         /// <summary>
         /// Get a processStepResult.
         /// </summary>
-        [HttpPost("ProcessStepResult")]
-        public async Task<ActionResult<ProcessStepResultDto>> GetAsync(ProcessStepResultRequestDto processStepResultRequestDto)
+        [HttpPost("ProcessResult")]
+        public async Task<ActionResult<ProcessStepResultDto>> GetAsync(ProcessStepResultRequestDtoV2 processResultRequestDto)
```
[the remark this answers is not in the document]
ManBearTM: I'll take this one

libs/dh/wholesale/feature-calculation-steps/src/lib/dh-wholesale-calculation-steps.component.ts
frontend-specialisten: If we know it will be on step 3? otherwise, I think we should add it when the step which needs it is added.

libs/dh/shared/domain/src/lib/generated/v1/model/process-step-type.ts
frontend-specialisten: I think we have a clean-up problem with our generated files, so to solve this I guess we should clean up the whole `/generated/v1/*` and create the client again.

frontend-specialisten: Generated files should be cleaned-up now :)

frontend-specialisten (approved):

PR #1345, opened 2023-01-24
ref(style): Align dropdown

Lubbe91:
## Description

- change the color of text from primary to natural black
- add a chevron instead of the arrow.

The arrow in select component from the material UI is drawn with CSS, so I had to get creative and try to match the icon version.

If anybody has a better idea of how to implement the chevron I'm open to suggestions

image

image

## References

- #1191


dzhavat (approved):
praise: really like the chevron. Good job.

PR #1355, opened 2023-01-26
Feat/wholesale restrict datepicker

frontend-specialisten:
## Description

- Add support for setting min/max date for `watt-datepicker`.
- Set min date for selecting period at the start batch to the oldest grid area `validFrom`.
- Set max date for selecting period at the start batch to the current date. 

image


ManBearTM (approved):
👍

PR #1362, opened 2023-01-27
Fixes after UX review

FirestarJes:


libs/dh/admin/feature-user-management/src/lib/drawer/tabs/content/du-user-roles.component.spec.ts
```diff
@@ -53,7 +53,7 @@ describe('UserRolesComponent', () => {
     await setup();
     await waitFor(() => {
       expect(
-        screen.getByRole('heading', { name: /3 roles/i })
+        screen.getByRole('heading', { name: /3 user roles/i })
```
dzhavat, line 56: nitpick: not related to the change but noticed that the file name starts with `du-`. Can you change it to `dh-` instead? Nice change btw :)

dzhavat (approved):

PR #1397, opened 2023-02-07
Small cleanup and API optimization

MadsThy:
## Description

The store providers caused the stores to initialize twice, so cleaning those up.


Exobitt (approved):
Lgtm

PR #1403, opened 2023-02-08
feat(watt): add description list component

frontend-specialisten:
## Description

image


libs/ui-watt/src/lib/components/watt-description-list/watt-description-list.component.ts
```diff
@@ -0,0 +1,48 @@
...
+  imports: [CommonModule],
+})
+export class WattDescriptionListComponent {
+  @Input() groups: WattDescriptionListGroups = [];
+  @Input() set groupsPerRow(value: number) {
+    this.elementRef.nativeElement.style.setProperty(
+      '--watt-description-list-groups-per-row',
+      value
+    );
+  }
```
dzhavat, line 45: suggestion: what do you think about using `HostListener` to set the variable? So `ElementRef` wouldn't be needed. E.g.

```ts
@HostBinding('style.--watt-description-list-groups-per-row')
  get groupsPerRowCss() {
    return this.groupsPerRow;
  }
```

frontend-specialisten: Love it! I change that

libs/ui-watt/src/lib/components/watt-description-list/watt-description-list.component.html
```diff
@@ -0,0 +1,29 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<dl>
+  <div
+    *ngFor="let term of groups"
+    [ngClass]="{ 'force-new-row': term.forceNewRow }"
+  >
```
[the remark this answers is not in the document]
frontend-specialisten: The `div` annoys me as well. However, it's also used to wrap the `dt` & `dd` elements used for the layout. 
The good thing though is that it's semantically allowed: [Wrapping name-value groups](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#wrapping_name-value_groups_in_div_elements)

dzhavat (approved):

PR #1413, opened 2023-02-10
refactor(dh): send component data as Input

dzhavat:
### DataHub
- send component data as Input
- set `OnPush` on component
- organize imports

## References

- N/A

### DIFF
--- a/libs/dh/admin/feature-create-user-role/src/lib/tabs/create-role-permissions-tab/dh-create-userrole-permissions-tab-table.component.ts
+++ b/libs/dh/admin/feature-create-user-role/src/lib/tabs/create-role-permissions-tab/dh-create-userrole-permissions-tab-table.component.ts
@@ -24,6 +24,7 @@ import {
   ViewChild,
 } from '@angular/core';
 import { translate } from '@ngneat/transloco';
+
 import {
   WattTableDataSource,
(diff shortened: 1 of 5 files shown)


libs/dh/admin/feature-create-user-role/src/lib/tabs/create-role-permissions-tab/dh-create-userrole-permissions-tab-table.component.ts
```diff
@@ -24,6 +24,7 @@ import {
   ViewChild,
 } from '@angular/core';
 import { translate } from '@ngneat/transloco';
+
```
dzhavat, line 27: suggestion: @FirestarJes I completely forgot about the `this.permissionsTable._selectionModel.clear()` workaround on line 73. Removing it does not reset the selection when new data arrives in the table, so any previously selected rows are still returned. The reset is still needed but I extended the `watt-table` component to expose a method, instead of using an internal property directly.

Please use the method exposed by #1414

FirestarJes: Will do :)

mimse (approved):

PR #1419, opened 2023-02-13
add dh orchestration

Renetnielsen:
## References

- #0000


.github/workflows/detect-changes.yml
```diff
@@ -12,33 +12,28 @@
...
   changes:
     name: Determine relevant jobs
     runs-on: ubuntu-latest
     # Set job outputs to values from filter step
     outputs:
-      dotnet: ${{ steps.filter.outputs.dotnet }}
-      frontend: ${{ steps.affected.outputs.is-affected }}
+      dh_dotnet: ${{ steps.filter.outputs.dh_dotnet }}
+      dh_frontend: ${{ steps.dh_affected.outputs.is-affected }}
+      watt_frontend: ${{ steps.watt_affected.outputs.is-affected }}
```
djorgensendk, line 35: won't these two not always evaluate with the same result? If that's the case, should it be one output instead?

.github/workflows/detect-changes.yml
```diff
@@ -82,9 +77,16 @@ jobs:
       - name: Install package dependencies
         run: yarn install --frozen-lockfile --non-interactive
```
djorgensendk, line 78: this step and 'PR: Fetch base branch Git history' could that be moved into the ./.github/actions/affected-action?

.github/workflows/dh-cd.yml
```diff
@@ -26,7 +26,7 @@ jobs:
   #
 
   changes:
-    uses: ./.github/workflows/dh-detect-changes.yml
```
djorgensendk, line 29: the outputs from the workflow - used in several steps below have new names

.github/workflows/detect-changes.yml
```diff
@@ -12,33 +12,28 @@
...
 # limitations under the License.
 
-# DataHub UI Frontend and BFF
-name: DH Detect changes
-
-# DESCRIPTION:
-# This workflow will determine which categories of jobs should be executed,
-# based on detected changes.
-# It should be used by the 'dh-ci.yml' and 'dh-cd.yml' to ensure they both use
-# the same path's to detect changes.
```
djorgensendk, line 22: the description could be changed instead of removed

.github/workflows/dh-ci-frontend.yml
```diff
@@ -13,11 +13,10 @@
 # limitations under the License.
 
 # DataHub UI Frontend and BFF
-name: DH CI
+name: DH CI frontend
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
```
djorgensendk, line 20: so concurrency isn't needed anymore? Should the new orchestrator have this instead?

.github/workflows/dh-ci-dotnet.yml
```diff
@@ -0,0 +1,78 @@
...
+
+env:
+  APP_NAME: app-dh
+  IS_MAIN_BRANCH: ${{ github.ref == 'refs/heads/main' }}
+  IS_PULL_REQUEST: ${{ github.event_name == 'pull_request' }}
+  NODE_OPTIONS: --max-old-space-size=6144
+  NODE_VERSION_FILE: .nvmrc
+  # keep in sync with "engines.yarn" and "volta.yarn" in "package.json"
+  YARN_VERSION: 1.22.19
+  NX_MAX_PARALLEL: 2
```
djorgensendk, line 29: why are these needed

.github/workflows/watt-cd.yml
```diff
@@ -31,7 +31,7 @@ jobs:
   #
 
   changes:
-    uses: ./.github/workflows/watt-detect-changes.yml
+    uses: ./.github/workflows/detect-changes.yml
```
djorgensendk, line 34: name of outputs are changed

djorgensendk (changes requested):

.github/workflows/detect-changes.yml
Renetnielsen:
```
- name: Is DH affected?
     id: dh_affected
     uses: ./.github/actions/affected
     with:
       base: ${{ github.ref == 'refs/heads/main' && 'HEAD~1' || format('origin/{0}', env.BASE_BRANCH) }}
       project: app-dh

- name: Is Watt affected?
  id: watt_affected
  uses: ./.github/actions/affected
  with:
    base: ${{ github.ref == 'refs/heads/main' && 'HEAD~1' || format('origin/{0}', env.BASE_BRANCH) }}
    project: ui-watt
```

.github/workflows/detect-changes.yml
Renetnielsen: Other projects is using this `is-affected` action, and i have no idea what it does, so i would rather let the frontend guys refactor this when they think its needed

.github/workflows/detect-changes.yml
```diff
@@ -12,33 +12,33 @@
...
-# DataHub UI Frontend and BFF
-name: DH Detect changes
-
 # DESCRIPTION:
 # This workflow will determine which categories of jobs should be executed,
 # based on detected changes.
-# It should be used by the 'dh-ci.yml' and 'dh-cd.yml' to ensure they both use
+# It should be used by the 'ci-orchestrator.yml', 'watt-cd.yml' and 'dh-cd.yml' to ensure they both use
 # the same path's to detect changes.
+name: Detect changes
```
djorgensendk, line 20:
```suggestion
# It should be used by the 'ci-orchestrator.yml', 'watt-cd.yml' and 'dh-cd.yml' to ensure they both use the same path's to detect changes.
name: Detect changes
```

Issue #1359, opened 2023-01-27
Input ranges are cut off


renehossle:
How to replicate

On the page "Fremsøg forretningsbeskeder" input mask width somehow cuts off the values.

https://images.zenhubusercontent.com/399470875/2378dbaa-f275-450a-8e60-ded32bce69fd/screen_recording_2023_01_27_at_10_22_09.mov

PR #1424, opened 2023-02-17
refactor(style): adjust icon datepicker to not clip text

Lubbe91:
## Description

Seems the added suffix class on the button for the time-/datepicker`(TD-component)` had some additional margin, which is causing issues in the context of the `(TD-component)`.

clipping

I've removed the suffix class and added some button style specific to the icon version of the watt button inside the `TD-component`.

not-clipping

image

This means that there are still unhandled cases if you use other variants of the button inside the `TD-component`.

At the time of writing, I'm unsure if want to have other versions of the button variant inside the `TD-component`, so I've contain this "fix" to only address the icon variant

Here is a `primary` variant use-case

primary

primary-focus

## References

- #1359


dzhavat (approved):
praise: thanks for looking into this 🙂

Lubbe91:
> praise: thanks for looking into this 🙂

I broke it, so I fix it 👍, (at least the part of it I broke 😄)

dzhavat (approved):

PR #1431, opened 2023-02-20
feat(watt): add support for setting initially selected rows on table

dzhavat:
### Watt

- add support for setting initially selected rows on table

image

## References

- N/A


libs/ui-watt/src/lib/components/table/watt-table.component.ts
```diff
@@ -210,6 +210,12 @@ export class WattTableComponent<T>
   @Input()
   selectable = false;
 
+  /**
+   * Sets the initially selected rows. Only works when selectable is `true`.
+   */
+  @Input()
+  initiallySelectedRows: T[] = [];
```
ManBearTM, line 217: suggestion: how about `initialSelection` instead? seems more in line with `selectionChange` and how variables for initial values usually are named.

libs/ui-watt/src/lib/components/table/watt-table.component.ts
```diff
@@ -301,8 +307,12 @@ export class WattTableComponent<T>
         .map((key) => this.columns[key].size)
         .map((size) => size ?? 'auto');
 
-      // Add space for extra checkbox column
-      if (this.selectable) sizing.unshift('var(--watt-space-xl)');
+      if (this.selectable) {
+        // Add space for extra checkbox column
+        sizing.unshift('var(--watt-space-xl)');
+
+        this._selectionModel.setSelection(...this.initiallySelectedRows);
```
ManBearTM, line 314: suggestion: this will reset the selection when `columns` or `displayedColumns` changes, which I don't think we want. perhaps move it out of the `if`, so it only happens when `changes.selectable` is truthy.

ManBearTM (changes requested):

libs/ui-watt/src/lib/components/table/watt-table.component.ts
dzhavat: Good point. Will do some more testing with difference scenarios to see what happens.

dzhavat: Tried a few cases and pushed a fix. It becomes a bit tricky when `selectable` is toggled. For now when `selectable` is truthy, the initial selection will be set. Toggling `selectable` afterwards will not cause any selection to be reset.

ManBearTM: That sounds fine IMO, I can't imagine any scenarios where you would want to toggle `selectable` anyway.

ManBearTM (dismissed):

libs/ui-watt/src/lib/components/table/watt-table.component.ts
ManBearTM: Another solution that might cover the use case you mentioned, is to make the checkbox column configurable in the `columns` object. Something like this:

```ts
columns = {
  checkbox: {
    accessor: row => row.enabled
  },
  id: {
    accessor: `id`
  }
}
```

Then if the row data from the server has `enabled` set to `true`, the checkbox would be selected initially. If the data is updated at a later time, the new rows should also be checked according to the `enabled` state.

Just an idea, haven’t thought it all the way through 🙂

dzhavat: Sounds interesting. Might be something we can explore in the future :)

ManBearTM (approved):

mimse (approved):
Looks good to me :cake:

PR #1435, opened 2023-02-21
Added variant, with solid and elevation

mimse:
Added variant to card, so the border style can be changed

### DIFF
--- a/libs/ui-watt/src/lib/components/card/watt-card.component.ts
+++ b/libs/ui-watt/src/lib/components/card/watt-card.component.ts
@@ -44,9 +44,9 @@ import { ChangeDetectionStrategy, Component, HostBinding, Input } from '@angular
   `,
 })
 export class WattCardComponent {
-  @Input() variant: 'watt-solid' | 'watt-elevation' = 'watt-elevation';
+  @Input() variant: 'solid' | 'elevation' = 'elevation';
   @HostBinding('class')
   get cssClass() {
-    return `watt-card ${this.variant}`;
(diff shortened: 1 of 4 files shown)


frontend-specialisten (commented):
Looks good to me. Please add a showcase to the storybook where the solid variant is used. :)

libs/ui-watt/src/lib/components/card/watt-card.component.ts
```diff
@@ -40,8 +44,9 @@ import { ChangeDetectionStrategy, Component, HostBinding } from '@angular/core';
   `,
 })
 export class WattCardComponent {
+  @Input() variant: 'watt-solid' | 'watt-elevation' = 'watt-elevation';
```
frontend-specialisten, line 47: Nitpick: could we remove the `watt` prefix from the input? It seems unnecessary to prefix the input.

mimse: Did it because the style was called watt-elevation, when i should use it directly without doing variant === elevation use watt-elevation class
Should i rename watt-elevation to elevation? Or leave as is?

frontend-specialisten: I still think the classes should be prefixed, however this prefix should just be added in the ‘@hostbinding’, so the consumers of the card component dont have to type the prefix themselves.

frontend-specialisten (approved):

dzhavat (approved):

PR #1437, opened 2023-02-21
Feature/1027 invite user part 7

mimse:
Choosing user roles when inviting a user 

image

## References

- https://github.com/Energinet-DataHub/ARCHIVED-geh-post-office/issues/1027


libs/dh/admin/data-access-api/src/lib/dh-admin-assignable-user-roles.store.ts
```diff
@@ -0,0 +1,91 @@
...
+  readonly requestState: LoadingState | ErrorState;
+  readonly assignableUserRoles: UserRoleDto[];
+}
+
+const initialState: State = {
+  requestState: LoadingState.INIT,
+  assignableUserRoles: [],
+};
+
+@Injectable({ providedIn: 'root' })
```
dzhavat, line 34: question: is this going to be used in other places than `DhAssignableUserRolesComponent`? If not, the `providedIn: 'root'` part can be removed. Providing a service in `root` means that the service will be long-lived once instantiated, even if it's not needed anymore.

Alternatively, it can be provided in the component where it's needed and Angular will destroy it automatically when the component is destroyed.

It seems that it's used both in `DhInviteUserModalComponent` (indirectly) and `DhAssignableUserRolesComponent` (directly). Since `DhAssignableUserRolesComponent` is used inside `DhInviteUserModalComponent`, the store can be provided in `DhInviteUserModalComponent` and both components will have access to it. 

So it can be injected in `DhInviteUserModalComponent` without the need to find a component and call one of its methods.

libs/dh/admin/feature-invite-user-modal/src/lib/dh-assignable-user-roles/dh-assignable-user-roles.component.html
```diff
@@ -0,0 +1,78 @@
...
+            context: {
+              $implicit: assignableUserRoles.length
+            }
+          "
+        ></ng-container>
+      </watt-card-title>
+      <div class="userRoles">
+        <ul class="watt-text-s">
+          <li *ngFor="let userRole of assignableUserRoles">
+            <ng-container>
```
dzhavat, line 41: nitpick: the `ng-container` element can be removed since it's not used for anything.

libs/dh/admin/data-access-api/src/lib/dh-admin-assignable-user-roles.store.ts
mimse: Copy / paste error :-)

dzhavat (approved):

dzhavat (approved):

PR #1448, opened 2023-02-23
Update wholesale package to 5.0.9

johevemi:
## References

- [#782](https://github.com/Energinet-DataHub/opengeh-wholesale/issues/782)


apps/dh/api-dh/openapitools.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "./node_modules/@openapitools/openapi-generator-cli/config.schema.json",
+  "spaces": 2,
+  "generator-cli": {
+    "version": "6.4.0"
+  }
+}
```
dzhavat, line 7: suggestion: we already have two (not sure why two, though) `openapitools.json` files (one on root level, another one in `DataHub.WebApi`). Maybe these can be used instead of adding a third one.

note: this is just a quick stop-by comment, not a full PR review 🙂

johevemi: idk what this is, it must be auto generated. Is this some thing that needs to changed?

ManBearTM (approved):

Issue #787, opened 2022-08-16
UX: Move in


mknic:
**Description**

As an energy supplier
I want to be able to register a new customer on a metering point 
So that I can move in new customers  

**Clarification**

- What information is satisfactory for MVP?
- Can we enter all necessary info up front?

**AC**
- [ ] Clarify with stakeholders what is needed in order to perform move in on a metering point
- [ ] Automate steps to minimize friction
- [ ] Allow user to automatically fetch info from known data to minimize data entry 
- [ ] Build prototype based on these requirements with components from WATT
- [ ] Demonstrate flows from start to finish
- [ ] Verify with stakeholders

**Definition of Ready**

- [ ] The issue is correctly estimated
- [ ] The issue is adequately described
- [ ] Possible dependencies are defined and aligned
- [ ] We have the necessary skills to complete this issue
- [ ] The issue can be completed within 1 iteration
- [ ] The issue has acceptance criteria defined
- [ ] The issue has adequate Definition Of Done described

**Definition of Done**

- [ ] Acceptance Criteria have been met
- [ ] The product has been demo’ed for relevant stakeholders
- [ ] Dependencies are handled
- [ ] The work has been documented
- [ ] The issue has been handed over and reviewed
- [ ] The PO has accepted the product

PR #1455, opened 2023-02-24
Feat/wholesale 787

frontend-specialisten:
## Description

**Make the frontend ready for balance responsible.**
Currently, the backend only supports energy suppliers as `MarketRole`, this PR makes the frontend ready for the balance responsible to be added to the `MarketRole`

A pipe for creating a `WattTableDataSource` has also been created, to avoid creating functions in the template (these functions are less performant since they're called a lot of times due to cd).

@Output for the pagination has also been added to make the consumers aware when it's initialized, so it can be applied to the data source.  

The batch id column has been removed from the search batches table. 

## References

- [#787](https://github.com/Energinet-DataHub/opengeh-wholesale/issues/787)


libs/dh/wholesale/feature-calculation-steps/src/lib/actors/dh-wholesale-actors.component.html
```diff
@@ -16,27 +16,27 @@
...
-      [dataSource]="_dataSource(energySuppliers)"
+      [activeRow]="getSelectedActor(actors)"
+      [activeRowComparator]="isSelectedActor"
+      [dataSource]="actors | wattDataSource: _dataSource"
       [columns]="_columns"
       (rowClick)="rowClick.emit($event)"
     >
     </watt-table>
-    <watt-paginator></watt-paginator>
+    <watt-paginator (initialized)="setPaginator($event)"></watt-paginator>
```
ManBearTM, line 33: Question: This `initialized` event seems very specific to this use case (rendering paginator async), so I fear it might be a little confusing for other devs whether they should always use this event or not. Have you considered using the `changes` property with `ViewChildren` instead?

I do like the convenience here though, so alternatively to make it even more convenient, would it be possible to do something like this (simplified):

```html
<watt-table #actors [dataSource]="_dataSource" [columns]="_columns"></watt-table>
<watt-paginator [for]="actors"></watt-paginator>
```

Then in the `watt-paginator` the `for` input would be a setter that sets the paginator instance on the tables data source. Could be pretty cool if (at all possible) and would eliminate the need for any logic in the component itself 🙂

libs/dh/wholesale/feature-calculation-steps/src/lib/actors/dh-wholesale-actors.component.html
```diff
@@ -15,28 +15,22 @@
...
+      {{ t("amountOfResults", { length: (actors || []).length }) }}
     </h5>
     <watt-table
       [resolveHeader]="t"
-      [activeRow]="getSelectedEnergySupplier(energySuppliers)"
-      [activeRowComparator]="isSelectedEnergySupplier"
-      [dataSource]="_dataSource(energySuppliers)"
+      [activeRow]="getSelectedActor(actors)"
+      [activeRowComparator]="isSelectedActor"
+      [dataSource]="actors | wattDataSource: _dataSource"
```
ManBearTM, line 28: Question: I am not against this pipe, but couldn't the problem be solved in the component doing something like below?

```ts
_dataSource = new WattTableDataSource<WholesaleActorDto>([]);
actors$ = this.store.actors$.pipe(exists(), tap(actors => this._dataSource.data = actors);
```

libs/dh/wholesale/feature-calculation-steps/src/lib/actors/dh-wholesale-actors.component.html
frontend-specialisten: Agreed, that would be an excellent solution! I will try to implement that.

libs/dh/wholesale/feature-calculation-steps/src/lib/actors/dh-wholesale-actors.component.html
frontend-specialisten: Good point, I changed it and removed the pipe

libs/dh/wholesale/feature-search/src/lib/grid-areas/dh-wholesale-grid-areas.component.html
```diff
@@ -33,6 +32,7 @@ <h3 class="watt-on-light--high-emphasis">
     (rowClick)="this.selected.emit($event)"
     [disabled]="disabled"
   ></watt-table>
+  <watt-paginator></watt-paginator>
```
ManBearTM, line 35: Did you forget `[for]` here? 🙂

frontend-specialisten: Should be added now :D

ManBearTM (dismissed):

ManBearTM (approved):

PR #1480, opened 2023-03-02
add graphql schema

Renetnielsen:
## References

- #0000


.github/workflows/dh-ci-frontend.yml
```diff
@@ -68,7 +68,9 @@ jobs:
 
       - name: Setup Release
         shell: bash
-        run: cp -R ${{ github.workspace }}/dist/apps/dh/${{ env.APP_NAME }} ${{ env.RELEASE_FOLDER_PATH }}
+        run: |
+          cp -R ${{ github.workspace }}/dist/apps/dh/${{ env.APP_NAME }} ${{ env.RELEASE_FOLDER_PATH }}
+          cp -R ${{ github.workspace }}/graphql.schema ${{ env.RELEASE_FOLDER_PATH }}
```
frontend-specialisten, line 73: Shouldn't this be done in the `dh-ci-dotnet.yml` instead? The file is always at the repository root, so this should work. However, the file is more related to the API and not the front end itself, and it's copied when the BFF is built to the agreed destinations.

Renetnielsen: Im guessing you are right! 
I will look into moving it

ManBearTM (approved):

PR #1498, opened 2023-03-07
feat(dh): settlement reports graphql

ManBearTM:


libs/dh/wholesale/feature-settlement-reports/src/lib/dh-wholesale-settlement-reports.component.html
```diff
@@ -27,6 +27,7 @@ <h3 class="watt-on-light--high-emphasis">
         <dh-wholesale-form
           (filterChange)="onFilterChange($event)"
           [loading]="loading"
+          [executionTime]="executionTime"
```
frontend-specialisten, line 30: Suggestion: do you think we should do the input as all the filters instead of only the `execution time` for future-proofing?

libs/dh/wholesale/feature-settlement-reports/src/lib/form/dh-wholesale-form.component.ts
```diff
@@ -95,30 +97,35 @@ export class DhWholesaleFormComponent implements AfterViewInit, OnDestroy {
...
+      Boolean(filters.executionTime?.start) === Boolean(filters.executionTime?.end)
+    );
+  }
 
-    this.filterChange.emit(this.filters.value as SettlementReportsProcessFilters);
+  ngAfterViewInit() {
+    this.filters.valueChanges
+      .pipe(takeUntil(this.destroy$), debounceTime(200), filter(this.isComplete))
+      .subscribe((filters: SettlementReportFilters) => {
+        console.log(filters);
```
frontend-specialisten, line 126: nitpick: remove `console.log`
```suggestion
```

libs/dh/wholesale/feature-settlement-reports/src/lib/form/dh-wholesale-form.component.ts
```diff
@@ -95,30 +97,35 @@ export class DhWholesaleFormComponent implements AfterViewInit, OnDestroy {
...
     ],
   });
 
   constructor(private fb: FormBuilder) {}
 
-  ngAfterViewInit() {
-    this.filters.valueChanges.pipe(takeUntil(this.destroy$)).subscribe((filters: unknown) => {
-      this.filterChange.emit(filters as SettlementReportsProcessFilters);
-    });
+  private isComplete(filters: SettlementReportFilters) {
```
frontend-specialisten, line 115: Consideration: We may need a validator for our range inputs to handle these scenarios.

frontend-specialisten (approved):

libs/dh/wholesale/feature-settlement-reports/src/lib/dh-wholesale-settlement-reports.component.html
ManBearTM: Good suggestion, let's revisit it in the next PR 👍 Would especially be nice if it could avoid the `patchValue` in the input setter, as that felt a little hacky.

libs/dh/wholesale/feature-settlement-reports/src/lib/form/dh-wholesale-form.component.ts
ManBearTM: Good idea, that sounds like the solution I was looking for 😄 Let's try and implement that in the next PR!

frontend-specialisten (approved):

Issue #1506, opened 2023-03-09
Fine tune watt-description-list


ManBearTM:
* Declarative API (sub components ala watt-tabs)
* Add new table-like variant (stacked)

PR #1519, opened 2023-03-13
More declarative, and new variant

mimse:
## Description

Made the description list more declartive to use. Also added variant. So we now have the noarmal flow, and a stack version.

## References
https://github.com/Energinet-DataHub/greenforce-frontend/issues/1506


FirestarJes (commented):
Looks fine to me, but a more experienced frontender should probably also take a look at it.

libs/dh/wholesale/feature-search/src/lib/batch-details/dh-wholesale-batch-details.component.ts
```diff
@@ -75,29 +73,31 @@ export class DhWholesaleBatchDetailsComponent {
...
+  //         : '-',
+  //     },
+  //     {
+  //       term: translate('wholesale.batchDetails.executionTime'),
+  //       description: this.batch?.executionTimeStart
+  //         ? (dateTimePipe.transform(this.batch?.executionTimeStart) as string)
+  //         : '-',
+  //     },
+  //   ];
+  // }
```
dzhavat, line 100: question: any reason for keeping this code?

libs/dh/wholesale/feature-calculation-steps/src/lib/results/dh-wholesale-results.component.html
```diff
@@ -30,10 +30,23 @@
...
+    <watt-description-list-item
+      [label]="t('meteringPointType')"
+      [value]="t('timeSeriesType.' + processStepResults.timeSeriesType)"
+    ></watt-description-list-item>
+    <watt-description-list-item
+      [label]="t('calculationPeriod')"
+      [value]="batch?.period | dhDateRange"
+    >
+    </watt-description-list-item>
+    <watt-description-list-item [label]="t('sum')" [value]="processStepResults.sum + ' kWh'">
```
frontend-specialisten, line 43: Question: In the old implementation, we can force a new row. Do you know how we can do this here? Should this be done by having two lists or a property on the `watt-description-list-item` component?

frontend-specialisten (commented):
Praise: nice changes! I only have a single question and also the build needs to be fixed :)

libs/dh/wholesale/feature-calculation-steps/src/lib/results/dh-wholesale-results.component.html
mimse: You can set forceNewRow=true on a list item

frontend-specialisten: Perfect!

dzhavat (approved):

PR #1522, opened 2023-03-13
Update button for creating contract

MadsThy:
## Description

- Changed "Active" state into "Activated", and now show it as a badge from WATT
- Changed "Enable" to "Activate", and implemented a spinner inside it, and disable all buttons when one is clicked, until status returns from backend.

image


Exobitt (dismissed):

libs/eo/metering-points/shell/src/lib/eo-metering-point-table.component.ts
```diff
@@ -84,16 +98,22 @@ import { EoMeteringPoint, EoMeteringPointsStore } from './eo-metering-points.sto
...
-        >
+            <watt-badge *ngIf="element.contract" type="success">Activated</watt-badge>
+            <button
+              *ngIf="!element.contract"
+              [disabled]="(loading$ | async) === true"
+              class="link"
+              (click)="createContract(element.gsrn)"
+            >
+              Activate
+              <watt-spinner *ngIf="(loading$ | async) === true" [diameter]="16"></watt-spinner>
```
duizer, line 109: Using this approach means that when the user clicks "Activate" for a single metering point, then all "Activate" buttons for all other metering points will start spinning and the buttons will be disabled.
Is it possible to change this, so it is only the "Activate" button for that row/metering point that is spinning and disabled? That would also mean, that the other "Activiate" buttons can be clicked; this can result in multiple concurrent calls, but that is not a problem for the API.

duizer (changes requested):
Please change so only a single "Activate" button is spinning and not all in the table

libs/eo/metering-points/shell/src/lib/eo-metering-point-table.component.ts
MadsThy: @duizer yeah sure :) I was not sure if the API could handle it or not, but if it's fine, I will go ahead and make it possible to click more buttons at once

MadsThy: Changed it so they can now load seperately

libs/eo/metering-points/shell/src/lib/eo-metering-points.store.ts
```diff
@@ -65,15 +57,21 @@ export class EoMeteringPointsStore extends ComponentStore<EoMeteringPointsState>
...
-        }
-        return mp;
-      }),
+      meteringPoints: state.meteringPoints.map((mp) =>
+        mp.gsrn === contract.gsrn ? { ...mp, contract } : mp
+      ),
+    })
+  );
+
+  private readonly setContractLoading = this.updater(
```
duizer, line 69: Just a comment about naming of this function: I think this should be named `toggleContractLoading` as it is toggling if the contract is loading or not

duizer (approved):

libs/eo/metering-points/shell/src/lib/eo-metering-points.store.ts
MadsThy: Right you are @duizer! :)

duizer (approved):

PR #1549, opened 2023-03-22
feat(dh): refresh data in drawer when permission edit is successful

dzhavat:
### DataHub
- refresh data in drawer when permission edit is successful

## References

- https://github.com/Energinet-DataHub/Titans/issues/13


mimse (approved):
:100:

PR #1571, opened 2023-03-27
feat(dh): add query param for graphQL request with the operation name

frontend-specialisten:
## Description

Add query param for graphQL request with the operation name. 
Before, all requests were `/graphql` in the network tab. Now, the operation name is added `/graphql?some-awesome-operation` to make it easier to identify the requests.


mimse (approved):
Nice!

PR #1577, opened 2023-03-28
fix(dh): refresh permissions log data in drawer

dzhavat:
### DataHub
- refresh permissions log data in drawer
- rename Output property
- rename class property
- set initial values to properties

## References

- https://github.com/Energinet-DataHub/Titans/issues/4


mimse (approved):
:100:

PR #1582, opened 2023-03-28
feat(dh): remove status field from user role creation flow

dzhavat:
### DataHub
- remove status field from user role creation flow

image

## References

- https://github.com/Energinet-DataHub/Titans/issues/38


mimse (approved):
👍🏻

PR #1586, opened 2023-03-29
Update NX to latest version (15.8.9)

frontend-specialisten:
## Description

Update NX to latest version (15.8.9)

- NX migrations has been implemented
- Some `tsconfig.spec.json` has now a target of `ES2016` due to some issues caused by Material harnesses: 
ex. `Expected to be running in 'ProxyZone', but it was not found.`


mimse (approved):

migrations.json
dzhavat, line 1: question: do we need this file?

jest.preset.js
```diff
@@ -4,10 +4,16 @@ const path = require('path');
...
+  /* TODO: Update to latest Jest snapshotFormat
+   * By default Nx has kept the older style of Jest Snapshot formats
+   * to prevent breaking of any existing tests with snapshots.
+   * It's recommend you update to the latest format.
+   * You can do this by removing snapshotFormat property
+   * and running tests with --update-snapshot flag.
+   * Example: "nx affected --targets=test --update-snapshot"
+   * More info: https://jestjs.io/docs/upgrading-to-jest29#snapshot-format
+   */
+  snapshotFormat: { escapeString: true, printBasicPrototype: true },
```
dzhavat, line 18: question: did you try running the example command in the comment above to migrate to the latest format?

libs/ui-watt/tsconfig.spec.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "outDir": "../../dist/out-tsc",
     "module": "commonjs",
-    "types": ["jest", "node"]
+    "types": ["jest", "node"],
+    "target": "ES2015"
```
dzhavat, line 7: question: is there an Issue in Material's repo that tracks the "ProxyZone" error? We should definitely aim for bumping the "target" ES version, so maybe we can create an Issue for tracking it.

dzhavat (commented):
praise: nice that you've managed to upgrade to latest Nx (and Angular) 👌 
Have a few questions :)

ManBearTM (approved):

jest.preset.js
```diff
@@ -4,10 +4,16 @@ const path = require('path');
 
 module.exports = {
   ...nxPreset,
-  testEnvironment: path.join(
-    __dirname,
-    'scripts/gf/test-util',
-    'jsdom-lax-ssl-environment.ts'
-  ),
+  testEnvironment: path.join(__dirname, 'scripts/gf/test-util', 'jsdom-lax-ssl-environment.ts'),
   testURL: 'https://localhost:5001',
```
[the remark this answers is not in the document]
frontend-specialisten: If we change this, it will properly cause the need to manually migrate in later versions since it's not NX standard, so I wouldn't recommend changing this.

migrations.json
frontend-specialisten: No we don't, the file has now been deleted

jest.preset.js
frontend-specialisten: Yes and it didn't change anything, I guess we aren't using snapshot testing anywhere.

libs/ui-watt/tsconfig.spec.json
frontend-specialisten: There aren't any active issue on this, but found the solution here: https://github.com/thymikee/jest-preset-angular/issues/1058

The problem seems to be caused by Zone.JS cannot intercept native async/await which are present in ES2017+

dzhavat (approved):

PR #1607, opened 2023-03-31
feat: search permissions

michaelmajgaard:


libs/dh/admin/feature-permissions/src/lib/overview/dh-admin-permission-overview.component.ts
```diff
@@ -97,7 +91,26 @@ export class DhAdminPermissionOverviewComponent implements OnInit, OnDestroy {
   }
 
   refreshData(): void {
-    this.getPermissionsQuery.refetch();
```
dzhavat, line 100: suggestion: instead of unsubscribing and subscribing again, the new search term can be sent to `refetch` as a param. Have you tried that? See `dh-wholesale-search.component.ts#105` as an example. This way the original `getPermissionsWatchQuery` function will stay mostly the same (except the addition of `variables`).

libs/dh/shared/ui-search/jest.config.ts
```diff
@@ -0,0 +1,38 @@
...
+  transform: {
+    '^.+\\.(ts|mjs|js|html)$': 'jest-preset-angular',
+  },
+  transformIgnorePatterns: ['node_modules/(?!.*\\.mjs$)'],
+  snapshotSerializers: [
+    'jest-preset-angular/build/serializers/no-ng-attributes',
+    'jest-preset-angular/build/serializers/ng-snapshot',
+    'jest-preset-angular/build/serializers/html-comment',
+  ],
+};
```
dzhavat, line 38: suggestion: the config in this file has changed slightly since the update of some dependencies. The new format can be seen in existing `jest.config.ts` files e.g. "\libs\dh\shared\ui-paginator\jest.config.ts". Let's align the config here to that file.

Note: if you used the custom library generator to generate this library, the "old" config syntax is still used. The generator will need to be updated.

libs/dh/shared/ui-search/tsconfig.spec.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../../../dist/out-tsc",
+    "module": "commonjs",
+    "types": ["jest", "node"]
```
dzhavat, line 6: suggestion: let's add `"target": "es2016"` here similar to other `tsconfig.spec.json` files, e.g. "\libs\dh\shared\ui-paginator\tsconfig.spec.json". This is required after updating some dependencies a couple of weeks ago.

libs/dh/admin/feature-permissions/src/lib/overview/dh-admin-permission-overview.component.ts
michaelmajgaard: Fixed

libs/dh/shared/ui-search/jest.config.ts
michaelmajgaard: Fixed

libs/dh/shared/ui-search/tsconfig.spec.json
michaelmajgaard: Fixed

libs/dh/admin/feature-permissions/src/lib/overview/dh-admin-permission-overview.component.ts
```diff
@@ -48,15 +50,19 @@ import { getPermissionsWatchQuery } from '../shared/dh-get-permissions-watch-que
...
+    DhSharedUiSearchComponent,
   ],
 })
 export class DhAdminPermissionOverviewComponent implements OnInit, OnDestroy {
-  private getPermissionsQuery = getPermissionsWatchQuery();
-  private subscription!: Subscription;
+  private apollo = inject(Apollo);
+  private destroy$ = new Subject<void>();
+
+  query = getPermissionsWatchQuery(this.apollo, undefined);
```
dzhavat, line 60: suggestion: the `query` is initialized together with the class, so the `inject(Apollo)` will also be available inside the `getPermissionsWatchQuery` function. Let's move it there instead. The `searchTerm` param is also optional, so it can be skipped as it will be `undefined` by default.

libs/dh/admin/feature-permissions/src/lib/overview/dh-admin-permission-overview.component.ts
```diff
@@ -81,10 +87,12 @@ export class DhAdminPermissionOverviewComponent implements OnInit, OnDestroy {
         this.error = error;
       },
     });
+    this.refresh();
```
dzhavat, line 90: question: is this call necessary? I probably necessary on refresh but not on initialization. 🤔

michaelmajgaard: Fixed

dzhavat (approved):

PR #1613, opened 2023-04-04
refactor(dh): switch to using wizard version for create user role

dzhavat:
### DataHub
- change "create user role" to a stepper in a modal instead of going to a separate page
- add missing property to project files

image

## References

- https://github.com/Energinet-DataHub/Titans/issues/38


libs/dh/admin/feature-create-user-role-modal/src/lib/dh-create-user-role-modal.component.ts
```diff
@@ -0,0 +1,223 @@
...
+  private readonly onErrorFn = () => {
+    const message = this.transloco.translate(
+      'admin.userManagement.createrole.createRoleRequest.error'
+    );
+
+    this.toastService.open({ message, type: 'danger' });
+  };
+
+  private buildEicFunctionOptions() {
+    this.transloco
```
frontend-specialisten, line 207: Consideration: Maybe we should create a shared function for this. We are doing the same in wholesale.

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab.component.ts
```diff
@@ -111,12 +118,6 @@ export class DhUserRolesTabComponent {
...
 
   readonly createUserRole = () => {
-    const url = this.router.createUrlTree([
-      dhAdminPath,
-      dhAdminUserManagementPath,
-      dhAdminUserRoleManagementCreatePath,
-    ]);
-
-    this.router.navigateByUrl(url);
+    console.log('create user role');
```
frontend-specialisten, line 121: Question: is this console.log intentional to be deployed?

tsconfig.base.json
```diff
@@ -27,8 +27,8 @@
       "@energinet-datahub/dh/admin/data-access-graphql": [
         "libs/dh/admin/data-access-graphql/src/index.ts"
       ],
-      "@energinet-datahub/dh/admin/feature-create-user-role": [
-        "libs/dh/admin/feature-create-user-role/src/index.ts"
+      "@energinet-datahub/dh/admin/feature-create-user-role-modal": [
```
frontend-specialisten, line 30: Consideration: Maybe not name the library after the layout but functionality/business logic, otherwise it will need a rename again if the UX changes back to a page or uses a drawer or what so ever.

frontend-specialisten (approved):
Looks good to me, with only some minor comments :)

libs/dh/admin/feature-user-management/src/lib/tabs/roles-tab/dh-roles-tab.component.ts
dzhavat: No, the method is not used anymore. Should be deleted now.

tsconfig.base.json
dzhavat: Good point. Will rename the library.

libs/dh/admin/feature-create-user-role-modal/src/lib/dh-create-user-role-modal.component.ts
dzhavat: I've noticed the repetition as well. I was thinking of creating a directive that can be added to the dropdown (e.g. https://timdeschryver.dev/blog/use-angular-directives-to-extend-components-that-you-dont-own#directives-that-load-data). If we go with a function instead, then we'd probably need to return an Observable and subscribe to it in the template. Which is also fine. In the function above, the subscription is in the component.

I'll make the refactor in a separate PR.

frontend-specialisten (approved):

Issue #1636, opened 2023-04-14
Deprecated License checker GitHub action


dzhavat:
https://github.com/kt3k/license_checker_github_action

We use this action in `license-check-ci.yml` pipeline. They recommend using https://github.com/kt3k/deno_license_checker instead.
https://github.com/kt3k/license_checker_github_action

PR #1643, opened 2023-04-14
chore: replace deprecated action

dzhavat:
### Workspace
- replace deprecated action

## References

- #1636


dzhavat:
Doesn't look like the solution is a simple replacement of the deprecated action with an `npx` command. For some reason the license header in HTML files was not detected in existing files. Needs to be investigated more.

PR #1649, opened 2023-04-18
Removed message archive from BFF and frontend

mimse:
Message archive backend, is no more. So have removed the reference for it. 
I removed the message archive UI from the being loaded in the shell.
But were not able to only remove the reference for the component in charges. It still broke the build.


kft (changes requested):
Message archive is "owned" by EDI. And are using this frontend functionality, so please do not remove :-)

mimse:
> Message archive is "owned" by EDI. And are using this frontend functionality, so please do not remove :-)

I was under the impression that the message archive backend has been removed?

PR #1664, opened 2023-04-24
feat(dh): make it possible to debug app-dh from vscode

frontend-specialisten:
## Description

- Make it possible to debug `app-dh` from vscode

image

- Generated a self-signed certificate, so it doesn't change occasionally. 
- Add documentation on how to trust the certificate on Windows easily.


mimse (approved):
🥇

PR #1670, opened 2023-04-26
Fixed email validation

mimse:
Fixed e-mail validation

image


libs/dh/admin/feature-invite-user-modal/src/lib/dh-invite-user-modal.component.ts
```diff
@@ -96,7 +96,10 @@ export class DhInviteUserModalComponent implements AfterViewInit, OnDestroy {
     actorId: ['', Validators.required],
     firstname: ['', Validators.required],
     lastname: ['', Validators.required],
-    email: [{ value: '', disabled: true }, Validators.required],
+    email: [
+      { value: '', disabled: true },
+      [Validators.required, Validators.pattern('^[a-zA-Z0-9._%+-]+')],
```
dzhavat, line 101: suggestion: the pattern and translations are out of sync. There are more symbols in the pattern than the translation mentions. Should we synchronize them?

mimse: I just wrote what @PernilleChristiansen suggested, and allowed for all valid chars

dzhavat: Can it be something like 'Only write letters, numbers, dots, and "_, %, +, -"'?

mimse: Changed to "Type a valid email"

dzhavat (approved):

Issue #932, opened 2022-09-19
Replace ngx-slider with material slider


dzhavat:
**Description:**
The slider package is not maintained and will stop working with future versions of Angular at some point. We have agreed to replace the current slider with a slider from  Angular Material v15 has added support for range slider. 

**AC:**

- [ ] Replace `ngx-slider` with slider from material
- [ ] Adjust styling to use Watt colors

**Definition of Ready:**

- [ ] The issue is correctly estimated
- [ ] The issue is adequately described
- [ ] Possible dependencies are defined and aligned
- [ ] We have the necessary skills to complete this issue
- [ ] The issue can be completed withing 1 iteration
- [ ] The issue has acceptance criteria defined
- [ ] The issue has adequate Definition Of Done described

**Definition of Done:**

- [ ] Acceptance Criteria have been met
- [ ] The product has been demo’ed for relevant stakeholders
- [ ] Dependencies are handled
- [ ] The work has been documented
- [ ] The issue has been handed over and reviewed
- [ ] The PO has accepted the product

PR #1673, opened 2023-04-26
feat(dh): warning for start batch

ManBearTM:
## References

- #932


frontend-specialisten (approved):
We should do a follow-up PR making our mocks more intelligent making it possible to test more scenarios.

frontend-specialisten (approved):

PR #1675, opened 2023-04-27
Feature/dx git

mimse:
Prerequisite: Github CLI https://cli.github.com/

Added two scripts. One for creating a branch from an github issue, the other creates a PR.

To create a branch from an issue you write yarn take

Choose a team

image

Now you can write the issue number or leave it blank to select from a list

image

image

After selecting an issue press enter, and now the branch is created.

To create a PR, simply write yarn pr, and it open the PR web view on github


dx/create-branch-from-issue.js
```diff
@@ -0,0 +1,92 @@
...
+          });
+      });
+    } else {
+      createBracnhFromIssue(answers.issue, selectedRepo);
+    }
+  });
+
+function createBracnhFromIssue(issue, selectedRepo) {
+  exec(
+    `gh issue develop ${issue} --issue-repo ${selectedRepo} --checkout`,
```
ManBearTM, line 77: suggestion: this branch naming goes against our current naming strategy (branches prepended with `feature/`, `chore/`, `fix/` etc. I don't have strong opinions on the matter, but it should probably be cleared with the guild members if this is how it is going to be 🙂 Alternatively, you could prompt for which type of "branch" it is.

dx/create-pr.js
```diff
@@ -0,0 +1,52 @@
...
+          new inquirer.Separator(),
+          ...stdout
+            .split('\n')
+            .map((branch) => branch.trim())
+            .filter((branch) => branch !== '' && branch !== 'main'),
+        ],
+      },
+    ])
+    .then((answers) => {
+      exec(`gh pr create --base "${answers.branch}" --web`, (err, stdout, stderr) => {
```
ManBearTM, line 43: question: i use gh pr create a lot and it is usually pretty good at figuring out the base branch by itself, so I don't love it if I have to choose it manually every time. maybe we can have an additional choice which is to "not set" the base branch (as in use whatever gh cli thinks is correct)?

package.json
```diff
@@ -52,7 +52,9 @@
     "start": "ng serve",
     "test": "nx run-many --target=test --all --parallel",
     "update": "nx migrate latest",
-    "workspace-generator": "nx workspace-generator"
+    "workspace-generator": "nx workspace-generator",
+    "take": "node ./dx/create-branch-from-issue.js",
+    "pr": "node ./dx/create-pr.js"
```
ManBearTM, line 57: suggestion: consider using [`ts-node`](https://github.com/TypeStrong/ts-node) so we can use TypeScript.

ManBearTM (changes requested):
Great initiative!

I don't think this warrants a new top level folder named `dx`,  but should instead be placed under `tools` IMO. Also the name `dx` is too general I think, perhaps name it `github` since it contains helpers for automating github flows?

mimse:
@ManBearTM nice feedback, have implemented must of your ideas :-)

ManBearTM (approved):

Issue #346, opened 2022-02-07
Add notifications / alerts for smoke test


frontend-specialisten:
**Description**
Currently, we have smoke tests running every hour and we need to alert when a test is failing.

- health checks 

**AC**
- [x] Alert to the Teams channel `Frontend`.

**Note**
See https://github.com/Energinet-DataHub/greenforce-frontend/blob/main/.github/workflows/dh-cd.yml
```#
  # Send notification to teams channel if deployment dispatch failed
  #

  dispatch_failed:
    needs:
      [
        frontend_promote_prerelease,
        dotnet_promote_prerelease,
        dispatch_deploment_event,
      ]
    if: |
      always() &&
      contains(needs.*.result, 'failure')
    uses: Energinet-DataHub/.github/.github/workflows/notify-team.yml@v10
    with:
      team_name: TexasRangers
      subject: "Deployment dispatch failed: UI + BFF"
    secrets: inherit

PR #1689, opened 2023-05-04
Send notification to teams channel if healthchecks failed

frontend-specialisten:
## Description

Send notification to teams channel if healthchecks failed

## References

- #346


.github/workflows/app-dh-healthchecks.yml
```diff
@@ -42,3 +42,20 @@ jobs:
...
+    needs:
+      [
+        healthchecks
+      ]
+    if: |
+      always() &&
+      contains(needs.*.result, 'failure')
+    uses: Energinet-DataHub/.github/.github/workflows/notify-team.yml@v10
+    with:
+      team_name: TexasRangers
```
mimse, line 59: Should this not be a different team name?

dzhavat: It's a "legacy" name from the time we were part of The Outlaws. For now it needs to be TexasRangers. See 
https://github.com/Energinet-DataHub/.github/blob/main/.github/workflows/notify-team.yml#L56C51-L57

frontend-specialisten: What @dzhavat said :)

dzhavat (approved):

PR #1701, opened 2023-05-10
Fix wrong translation for edit userrole

FirestarJes:


libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -914,6 +914,7 @@
       "download": "Download",
       "drawer": {
         "edit": "Rediger rettighed",
```
dzhavat, line 916: suggestion: the edit button in user drawer also shows incorrect text value (in Danish). Can we fix that too :)

FirestarJes: Sure :)

dzhavat (approved):

PR #1715, opened 2023-05-15
Add actor picker component to new settlement reports page.

FirestarJes:


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/GraphQLQuery.cs
```diff
@@ -134,8 +138,8 @@ public GraphQLQuery()
 
             Field<NonNullGraphType<ListGraphType<NonNullGraphType<BatchType>>>>("batches")
                 .Argument<DateRangeType>("executionTime")
-                .Argument<BatchState>("executionState", nullable: true)
-                .Argument<ProcessType>("processType", nullable: true)
+                .Argument<BatchState>("executionState", true)
```
mimse, line 141: Suggestion: keep named param, so it is easier to see what is set to true

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/GraphQLQuery.cs
```diff
@@ -163,7 +167,7 @@ public GraphQLQuery()
 
             Field<NonNullGraphType<ListGraphType<NonNullGraphType<SettlementReportType>>>>("settlementReports")
                 .Argument<ProcessTypeEnum>("processType")
-                .Argument<string[]>("gridAreaCodes", nullable: true)
+                .Argument<string[]>("gridAreaCodes", true)
```
mimse, line 170: Suggestion: keep named param, so it is easier to see what is set to true

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/GraphQLQuery.cs
```diff
@@ -36,7 +36,8 @@ public GraphQLQuery()
                 .Resolve()
                 .WithScope()
                 .WithService<IMarketParticipantPermissionsClient>()
-                .ResolveAsync(async (context, client) => await client.GetPermissionAsync(context.GetArgument<int>("id")));
+                .ResolveAsync(
```
mimse, line 39: Question: Looks like your formatter is different that the standard one.

mimse (changes requested):

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/GraphQLQuery.cs
FirestarJes: hmm, have to double check, but it should just be the default Rider one

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/GraphQLQuery.cs
FirestarJes: Will do

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/GraphQLQuery.cs
FirestarJes: fixed

mimse (dismissed):

PR #1719, opened 2023-05-16
feat: batch created by name

michaelmajgaard:


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/BatchType.cs
```diff
@@ -32,7 +34,22 @@ public BatchType(IDataLoaderContextAccessor accessor)
...
             Field(x => x.ProcessType).Description("The process type.");
-            Field(x => x.CreatedByUserId).Description("Id of the user who started the batch.");
+
+            Field<NonNullGraphType<StringGraphType>>("CreatedByUserName")
+                .Resolve()
+                .WithScope()
+                .WithService<IMarketParticipantClient>()
+                .Resolve((context, markpart) =>
+                {
+                    var loader = accessor.Context!.GetOrAddLoader("SearchUsersAsync", () => markpart.SearchUsersAsync(
```
mimse, line 44: Is it not possible to search for the id?

michaelmajgaard: Idea was to fetch and cache all users and do lookup after, so as to not make a markpart call per batch

mimse (approved):

PR #1724, opened 2023-05-17
Chip mode for Dropdown 

frontend-specialisten:
## Description
Add chip mode for the dropdown component

## Tasks
- [x] Add chip mode for dropdown (both single and multi-select)
- [x] Handle validation errors for dropdown in chip mode 
- [x] Handle disable state
- [x] Handle "label" state
- [x] Fix build (maximum build size)
- [x] Fix tests

## Technical notes:
The chip component has been used to give us more flexibility (badge and trailing icon), and we don't have to duplicate the chip component behavior for the dropdown.

## References 

- Energinet-DataHub/team-titans/issues/116


libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.scss
```diff
@@ -49,6 +53,16 @@ watt-dropdown {
...
     color: currentColor;
   }
+
+  .watt-chip-label {
+    margin-left: var(--watt-space-s);
+    background: var(--watt-color-neutral-white);
+  }
+
+  .chip {
+    opacity: 0;
```
mimse, line 63: Question: if should be hidden, why not use display:none ?

libs/ui-watt/src/lib/components/dropdown/+storybook/storybook-dropdown-reactive-forms.stories.ts
```diff
@@ -101,6 +104,36 @@ SingleSelect.parameters = {
   },
 };
 
+export const SingleSelectChipMode: StoryFn<WattDropdownComponent> = (
```
mimse, line 107: Suggestion: Could be reuse the WattDropDownComponent stories we already have. But pass chipMode to them?

libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.scss
frontend-specialisten: Because `display: none` would make the dropdown not clickable, and then we would need to open the dropdown programmatically.

libs/ui-watt/src/lib/components/dropdown/+storybook/storybook-dropdown-reactive-forms.stories.ts
frontend-specialisten: Good point. I will look into that 👍

libs/ui-watt/src/lib/components/dropdown/watt-dropdown.component.scss
mimse: And visibility:hidden is not an option i guess? :D

frontend-specialisten: It has now been changed for `visibility: hidden` :D

Issue #1710, opened 2023-05-11
Chip design for datepicker


mimse:
Make a variant of the datepicker that looks like 

![image.png](https://images.zenhubusercontent.com/633fbb34b49ff882a80234ab/e1b8da12-096f-470f-be0c-3751cd595505)

PR #1731, opened 2023-05-22
Feature/chip design for datepicker

mimse:
Implementation of a chip readonly datepicker

billede

## References

- https://github.com/Energinet-DataHub/greenforce-frontend/issues/1710


dzhavat:
praise: looks quite nice so far! Good job 🚀

suggestion: the "normal" datepicker is set to an empty string as initial value. Do you think it's possible to do the same here (for consistency)?

image

suggestion: the input format in the "normal" datepicker is set to `dd-MM-yyyy`. Do you think we can do the same here? It can also be as an enhancement at a later point. 

image

frontend-specialisten:
Closing this in favor of [#1770](https://github.com/Energinet-DataHub/greenforce-frontend/pull/1770) using the new implementation of chips.

PR #1742, opened 2023-05-24
Move settlement reports tabs into component

FirestarJes:


libs/dh/wholesale/feature-settlement-reports/src/lib/tabs/dh-wholesale-settlements-reports-tabs-balance.component.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Apollo } from 'apollo-angular';
+import { WattDropdownModule, WattDropdownOption } from '@energinet-datahub/watt/dropdown';
+import { ActorFilter } from '@energinet-datahub/dh/wholesale/domain';
+
+@Component({
+  standalone: true,
+  selector: 'dh-wholesale-settlements-reports-tabs-balance',
+  templateUrl: './dh-wholesale-settlements-reports-tabs-balance.component.html',
+  styleUrls: [
+    '../dh-wholesale-settlements-reports-tab.component.scss',
```
dzhavat, line 36: suggestion: nice approach but with the twist that styling in this file *could* lead to side effects in this component. Can see that the common styling is related to `watt-card-title`. Let's copy the styling for now in this component's SCSS file. Later we can move it to a shared file. Or we can move it it a shared file already now. Could be in a `shared` sub-folder.

dzhavat (approved):

dzhavat (approved):

PR #1743, opened 2023-05-24
chore: add new self-signed certificate

frontend-specialisten:
## Description

Add a new self-signed certificate, since the current one is expired, for local development used by `app-dh.` 
The new certificate expires `30. april 2123` so it shouldn't cause any problems again in the nearest future. 

**REMEMBER!!!**
You will need to trust the certificate once again `certutil -addstore -f "Root" localhost.crt` (windows)

image


dzhavat (approved):
💯

Issue #1744, opened 2023-05-25
Remove modules from codebase


mimse:
Convert all ng modules to standalone components

PR #1755, opened 2023-05-26
Watt standalone

mimse:
Updated watt components to be standalone

## References

- https://github.com/Energinet-DataHub/greenforce-frontend/issues/1744


libs/ui-watt/src/lib/components/clipboard/watt-copy-to-clipboard.directive.spec.ts
```diff
@@ -17,13 +17,13 @@
...
 import userEvent from '@testing-library/user-event';
 
-import { WattToastModule } from '../toast';
+import { WattToastComponent } from '../toast';
 import { WattCopyToClipboardDirective } from './watt-copy-to-clipboard.directive';
 
 describe(WattCopyToClipboardDirective.name, () => {
   it('shows toast on click', async () => {
     await render(`<span wattCopyToClipboard>Text</span>`, {
-      imports: [WattToastModule.forRoot(), WattCopyToClipboardDirective],
```
dzhavat, line 26: suggestion: a bit weird that this works. Probably it's because `WattToastComponent` imports `MatLegacySnackBarModule`, which in turn provides the necessary providers. Instead of importing `WattToastComponent`, which looks unrelated to this directive anyway, let's use the `importProvidersFrom(MatLegacySnackBarModule)` pattern.

dzhavat (commented):
praise: fantastic job. Looks much better without NgModules. 🚀

Did you find it easy to convert SCAMs to Standalone?

mimse:
Most of them were easy to convert, the ones with forRoot were a bit harder :-)

dzhavat (approved):

PR #1770, opened 2023-05-31
feat(ui-watt): chip datepicker

ManBearTM:
## Description

Implementation of chip datepicker using new chip component.

## References

- https://github.com/Energinet-DataHub/greenforce-frontend/issues/1710


frontend-specialisten (dismissed):
We should do a follow-up PR consolidating `DhDatePipe` and `WattDatePipe`, if this isn't done as the next thing please create an issue for it at the bounty hunters board (so we don't forget).

frontend-specialisten (approved):

PR #1786, opened 2023-06-06
Chore/remove module part 1

mimse:
Removed modules in market participant

## References

- https://github.com/Energinet-DataHub/greenforce-frontend/issues/1744


libs/dh/shared/data-access-msw/src/lib/mocks/marketParticipant.ts
```diff
@@ -45,6 +47,15 @@ function getOrganizations(apiBase: string) {
   );
 }
 
+function getAllOrganizationsWithActors(apiBase: string) {
+  return rest.get(
+    `${apiBase}/v1/MarketParticipant/Organization/GetAllOrganizationsWithActors`,
+    (req, res, ctx) => {
+      return res(ctx.json(organizationWithActorsData));
```
dzhavat, line 54: suggestion: can we convert `organizationWithActorsData` to a TS variable? That way it can be typed, so changes to the endpoint will not result in out of sync mock.

mimse: Does that work with none graphql as well?

dzhavat (approved):

PR #1787, opened 2023-06-06
refac: use watt date pipe util

ManBearTM:
## Description

The WattDatePipe and WattDateTimePipe was introduced along with the new watt chip datepickers.
This removes the DhSharedUiDateTime library in favor of the new watt util.


libs/dh/wholesale/feature-calculation-steps/src/lib/results/dh-wholesale-results.component.html
```diff
@@ -37,7 +37,7 @@
     ></watt-description-list-item>
     <watt-description-list-item
       [label]="t('calculationPeriod')"
-      [value]="batch?.period | dhDateRange"
+      value="{{ batch?.period?.start | wattDate }} - {{ batch?.period?.end | wattDate }}"
```
frontend-specialisten, line 40: Question: Maybe we should keep a `DateRange` pipe to make our HTML cleaner? (since our date ranges always have the same format/type)?

libs/ui-watt/src/lib/utils/date/watt-datetime.pipe.spec.ts
```diff
@@ -14,15 +14,16 @@
...
-})
-export class DhSharedUiDateTimeModule {}
+  it('transforms "2015-01-24T03:14:15Z" to "24-01-2015 04:14"', () => {
+    expect(pipe.transform('2015-01-24T03:14:15Z')).toBe('24-01-2015 04:14');
+  });
+
+  it('transforms "2015-09-21T03:14:15Z" to "21-09-2015 05:14"', () => {
+    expect(pipe.transform('2015-09-21T03:14:15Z')).toBe('21-09-2015 05:14');
+  });
+});
```
dzhavat, line 29: suggestion: consider adding tests for falsy values (i.e. `undefined`, `null`, `''`)

PR #1792, opened 2023-06-07
refac(ui-watt): one pipe to rule them all

ManBearTM:
## Description

It seemed a little silly to have 4 different pipes for handling date transforms, so I consolidated the logic into a single WattDatePipe that handles both ranges and singular dates. It also adds an extra parameter for determining if the time should be included or not.


dzhavat (approved):
praise: it's perfect 🤗😄

frontend-specialisten (approved):
Praise: looks good, if we add a minimum of documentation to storybook it’s perfect 💪

ManBearTM:
> Praise: looks good, if we add a minimum of documentation to storybook it’s perfect 💪

I'll do it in a new PR 🙂

PR #1801, opened 2023-06-09
Create transfer agreement

MadsThy:
## Description
This PR does a few things:

- Adds danish as the locale of the app, provided by the GF shared component, which DH also uses. It is needed to get the watt components to behave in danish when it comes to locale.
- It enables the "New transfer agreement" button on the transfers page
- It implements a modal that shows when you click the "New transfer agreement" button on the transfers page
  - This modal for now only enables the user to enter an 8-digit CVR/TIN number of a receiver, and a to/from date
- It fixes the naming of the transfers components, so they are now all called "transfers" instead of "transfer"

Skærmbillede 2023-06-09 094158
Skærmbillede 2023-06-09 094149
Skærmbillede 2023-06-09 094132


libs/eo/transfers/src/lib/eo-transfers-modal.component.ts
```diff
@@ -0,0 +1,136 @@
...
+    private store: EoTransfersStore,
+    private cd: ChangeDetectorRef
+  ) {}
+
+  open() {
+    this.modal.open();
+  }
+
+  createAgreement() {
+    if (!this.form.controls['dateRange'].value || !this.form.controls['tin'].value) return;
```
duizer, line 114: Why not use `form.valid` ?

libs/eo/certificates/src/lib/certificate-details/eo-certificate-details.component.ts
```diff
@@ -76,11 +76,11 @@ import { EoCertificatesStore } from '../eo-certificates.store';
               <div>{{ cert?.quantity?.toLocaleString() }} Wh</div>
               <b>Start time</b>
               <div>
-                {{ cert?.dateFrom | date : 'HH:mm dd-MMM-y z' | uppercase }}
+                {{ cert?.dateFrom | date : 'HH:mm dd-MM-y z' | uppercase }}
               </div>
               <b>Start time</b>
               <div>
-                {{ cert?.dateTo | date : 'HH:mm dd-MMM-y z' | uppercase }}
+                {{ cert?.dateTo | date : 'HH:mm dd-MM-y z' | uppercase }}
```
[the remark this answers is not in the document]
MadsThy: That was a design choice by Kenneth back in the days :) I also don't think we should uppercase, but that is not within the scope of this PR.

libs/eo/transfers/src/lib/eo-transfers-table.component.ts
```diff
@@ -153,27 +155,25 @@ interface EoTransferTableElement extends EoTransfer {
     </watt-paginator>
     <ng-template #notActive><watt-badge type="neutral">Inactive</watt-badge></ng-template>
 
-    <eo-transfer-drawer></eo-transfer-drawer>
+    <eo-transfers-modal title="New Transfer Agreement"></eo-transfers-modal>
```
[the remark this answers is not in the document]
MadsThy: You are completely correct :) Changed this.

libs/eo/transfers/src/lib/eo-transfers-table.component.ts
```diff
@@ -201,12 +201,12 @@ export class EoTransferTableComponent implements AfterViewInit {
     this.dataSource.data = this.transfers;
   }
 
-  isDateActive(date: number): boolean {
+  isDateActive(date: string): boolean {
     return new Date(date).getTime() >= new Date().getTime();
```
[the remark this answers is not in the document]
MadsThy: Good point :) I have moved this functionality to a utility file in shared.

libs/eo/transfers/src/lib/eo-transfers-modal.component.ts
```diff
@@ -0,0 +1,136 @@
...
+      <watt-form-field>
+        <watt-label>Period</watt-label>
+        <watt-datepicker required="true" formControlName="dateRange" [range]="true" />
+        <watt-error *ngIf="form.controls.dateRange.errors"> A date range is required </watt-error>
+      </watt-form-field>
+    </form>
+    <watt-modal-actions>
+      <watt-button variant="secondary" (click)="modal.close(false)">Close</watt-button>
+      <watt-button [disabled]="!form.valid" (click)="createAgreement()"
+        >Create transfer agreement</watt-button
```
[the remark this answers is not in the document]
MadsThy: Corrected the other place it was writting in upper case :)

libs/eo/transfers/src/lib/eo-transfers-modal.component.ts
```diff
@@ -0,0 +1,136 @@
...
+  open() {
+    this.modal.open();
+  }
+
+  createAgreement() {
+    if (!this.form.controls['dateRange'].value || !this.form.controls['tin'].value) return;
+
+    const transfer = {
+      startDate: Math.round(new Date(this.form.controls['dateRange'].value.start).getTime() / 1000),
+      endDate: Math.round(new Date(this.form.controls['dateRange'].value.end).getTime() / 1000 - 1), // -1 to offset that endDate is next day at 00:00, when we need chosen day at 23:59:59
```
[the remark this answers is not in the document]
MadsThy: I completely agree. We usually have talked about that "end dates" should be until (but not including) the last hour/time that was sent forward. So if i send a time that says it should end at 12 feb, i will send that it ends at 12 feb 00:00 UTC. Unfortunately the current implementation of the API then sends back to me that the end date is also at this time. That means when i ask for "what day is the end date", it will says 12 feb., but in reality it ends on 11 feb 23:59:59.

So either i subtract the 1 second there, or i then always format dates to show "the date - 1 day". Which do you prefer? :)

libs/eo/transfers/src/lib/eo-transfers-modal.component.ts
MadsThy: Basically "because typescript" :) The fields can be null, and it does not like passing null values into "new Date"

Here is a picture of what it looks like. I avoid the errors by nullchecking in that if statement

image

MadsThy: I have put form.valid in now, just for good measure. But the issue here is mainly that typescript doesn't know i already checked for null outside the function :)

libs/eo/transfers/src/lib/test-data.ts
```diff
@@ -17,85 +17,85 @@
 export const testData = [
   {
     id: '99f96d44-494b-439f-83fd-8846b13d22ad',
-    startDate: new Date().setDate(new Date().getDate() - 60),
-    endDate: new Date().setDate(new Date().getDate() + 60),
+    startDate: '2023-06-08T22:00:00+00:00',
+    endDate: '2023-06-08T23:00:00+00:00',
```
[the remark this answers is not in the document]
MadsThy: I hope they change it too :)

libs/eo/transfers/src/lib/eo-transfers-modal.component.ts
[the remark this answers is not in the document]
duizer: Do not subtract 1 from these times. We will get into a world of problems. Here you are subtracting a single second - somewhere else it is a millisecond, somewhere else it is a minute; it ends up giving a lot of issues. 
Think of the time for a certificate this way: Each certificate is hourly-based, so a certificate that starts at 23:00:00 lasts 60 minutes; 23:00:00 + 60 minutes means that it will end at 24:00:00 (or 0:00:00). 

If there is a bug in the API, then you have to take that up with Saiuan or Christoffer.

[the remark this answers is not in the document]
MadsThy: Since the API and design currently don't support setting hours, and it seems like the datepicker wants to use start of day on startDate and end of day on endDate, i opted for just compensating when showing the table data :) It just goes back one day from whatever date it has set as endDate.

libs/eo/transfers/src/lib/eo-transfers-table.component.ts
```diff
@@ -201,12 +204,13 @@ export class EoTransferTableComponent implements AfterViewInit {
     this.dataSource.data = this.transfers;
   }
 
-  isDateActive(date: number): boolean {
-    return new Date(date).getTime() >= new Date().getTime();
+  compensateForEndOfDay(inputDate: string): number {
+    const date = new Date(inputDate);
+    return date.setDate(date.getDate() - 1);
```
duizer, line 209: Since the endDate can be at any point on the day, then I think you should check here, that you change the endDate if it is on midnight.

libs/eo/transfers/src/lib/test-data.ts
```diff
@@ -17,85 +17,85 @@
 export const testData = [
   {
     id: '99f96d44-494b-439f-83fd-8846b13d22ad',
-    startDate: new Date().setDate(new Date().getDate() - 60),
-    endDate: new Date().setDate(new Date().getDate() + 60),
+    startDate: '2023-06-08T22:00:00+00:00',
```
duizer, line 20: The API should return all timestamps in Unix Time Seconds. Is this a mock of what is returned from the API? If it is, then it should have timestamps in Unix Time Seconds

libs/eo/transfers/src/lib/eo-transfers-modal.component.ts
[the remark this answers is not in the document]
duizer: The API allows any timestamp you give it, so it can accept a period of `2023-06-12 01:23:45` to `2023-06-14 12:34:56`.

UI wise this is simplified to using a DatePicker. I have talked with Søren PO and what he wants is the start date and end date selected in the date picker is inclusive. This mean, that if the user picks `2023-06-12` as start and `2023-06-14`, then it should result in an agreement running from `2023-06-12 00:00` in the browser's timezone to `2023-06-15 00:00` in the browser's timezone. Remember that when communicating with the API, then the timestamps will always be in UTC (UnixTime Seconds) and the UI have to do the mapping between the timestamps in browser's timezone and UTC.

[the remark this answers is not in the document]
MadsThy: We found a solution to this :) FE should show the date from the day before, if the end date is midnight of the day after

libs/eo/transfers/src/lib/test-data.ts
MadsThy: Removed this file, and confirmed working with timestamps being returned instead of strings

libs/eo/transfers/src/lib/eo-transfers-table.component.ts
MadsThy: Removed this implementation

apps/eo/e2e-eo/src/e2e/transfers/transfers.feature
```diff
@@ -11,3 +11,19 @@ Feature: Transfers Page
     And I can see a download button
     And I can see a button to create a new agreement
 
+  Scenario: Charlotte CSR can access a modal for creating new transfer agreement on the transfers page
```
frontend-specialisten, line 14: Question: Aren't this scenario redundant as it's necessary to be able to access the modal to create the transfer agreement. Suggestion from my side would be to remove this scenario and instead add a scenario checking that we handle errors correctly.

libs/eo/transfers/src/lib/eo-transfers-modal.component.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.service.createAgreement(transfer).subscribe({
+      next: (transfer) => {
+        this.store.addTransfer(transfer);
+        this.requestLoading = false;
+        this.resetFormValues();
+        this.cd.detectChanges();
+        this.modal.close(true);
+      },
+      error: () => {
+        this.requestLoading = false;
```
frontend-specialisten, line 151: Question: should error handling be part of this PR, or are we doing a follow-up PR for that?

libs/eo/transfers/src/lib/eo-transfers.component.ts
```diff
@@ -18,13 +18,13 @@ import { AsyncPipe, NgIf } from '@angular/common';
...
-import { EoTransferStore } from './eo-transfers.store';
+import { EoTransfersTableComponent } from './eo-transfers-table.component';
+import { EoTransfersStore } from './eo-transfers.store';
 
 @Component({
   changeDetection: ChangeDetectionStrategy.OnPush,
-  selector: 'eo-transfer',
-  imports: [MatCardModule, EoTransferTableComponent, EoPopupMessageComponent, AsyncPipe, NgIf],
+  selector: 'eo-transfers',
+  imports: [MatCardModule, EoTransfersTableComponent, EoPopupMessageComponent, AsyncPipe, NgIf],
```
frontend-specialisten, line 27: Suggestion: it would be nice if we just use a `watt-card` instead `mat-card`

frontend-specialisten (dismissed):
Praise: only a few comments to consider :)

libs/eo/transfers/src/lib/eo-transfers-modal.component.ts
MadsThy: We haven't had a look at error handling yet, to be honest, so it's never been a priority :) Mainly because this is a beta product that does not have any users yet.

Maybe you and I should sit down and talk about scenarios? :)

apps/eo/e2e-eo/src/e2e/transfers/transfers.feature
MadsThy: Removed :)

frontend-specialisten (approved):

PR #1829, opened 2023-06-19
Edit end date of transfer agreements

frontend-specialisten:
## Description

Add support for editing the end date of a transfer agreement

image

## References

- [#1466](https://github.com/Energinet-DataHub/energy-origin-issues/issues/1466)


libs/eo/transfers/src/lib/eo-transfers.component.ts
```diff
@@ -39,11 +47,22 @@ import { EoTransfersStore } from './eo-transfers.store';
...
+export class EoTransfersComponent implements OnInit {
   error$ = this.store.error$;
-  constructor(private store: EoTransfersStore) {}
+  transfers$ = this.store.transfers$;
+  selectedTransfer$ = this.store.selectedTransfer$;
+
+  constructor(protected store: EoTransfersStore) {}
+
+  ngOnInit(): void {
+    this.store.getTransfers();
```
MadsThy, line 66: Question: why use the OnInit instead of placing it in the constructor? :)

MadsThy (approved):

libs/eo/transfers/src/lib/eo-transfers.component.ts
frontend-specialisten: The reason was to use the "inject" pattern instead, which should be done now. Besides that, it won't make any difference in the current state of the application, but in the future ex. Implementing transitions between the pages could have, and also, it's a common practice not to have logic inside the constructor.

libs/eo/transfers/src/lib/eo-transfers-edit-modal.component.ts
```diff
@@ -0,0 +1,164 @@
...
+      #modal
+      [title]="title"
+      size="small"
+      closeLabel="Close modal"
+      [loading]="patchingTransfer$ | push"
+      (closed)="whenClosed()"
+    >
+      <form [formGroup]="form">
+        <watt-form-field>
+          <watt-label>End date of the period!!</watt-label>
```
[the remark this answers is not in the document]
frontend-specialisten: A frustrating moment, which has now been remove :D

libs/eo/transfers/src/lib/eo-transfers-table.component.ts
```diff
@@ -164,32 +177,45 @@ interface EoTransferTableElement extends EoListedTransfer {
     </watt-paginator>
     <ng-template #notActive><watt-badge type="neutral">Inactive</watt-badge></ng-template>
 
-    <eo-transfers-modal title="New transfer agreement"></eo-transfers-modal>
-    <eo-transfers-drawer></eo-transfers-drawer>
+    <eo-transfers-create-modal title="New transfer agreement"></eo-transfers-create-modal>
+    <eo-transfers-drawer
+      [transfer]="selectedTransfer"
+      (closed)="transferSelected.emit(undefined)"
```
MadsThy, line 183: Question: Do we need send an undefined with this emit? :) I haven't used that before myself, so didn't know it was something you did

MadsThy (approved):

PR #1832, opened 2023-06-20
fix(dh): remove date-fns module from providers

### LINKED ISSUE #1827: Calendar has dots again
We deleted the dots before (17. instead of 17), now they are back.

They should be hidden or deleted.

dzhavat:
### DataHub
- remove `MatDateFnsModule` from providers. The reason is that our own `danishDatetimeProviders` provide the same providers that are coming from `MatDateFnsModule`. If we want to keep both `MatDateFnsModule` and `danishDatetimeProviders` then **the order is important** and `danishDatetimeProviders` **must** come last. Otherwise `MatDateFnsModule` will overwrite our providers.
- remove `MatDateFnsModule` from spec files
- update documentation

## References

- Closes #1827


mimse (dismissed):

mimse (approved):
💯 nice find

Issue #1796, opened 2023-06-08
Move start calculation functionality into search batch page


ManBearTM:
### Synopsis
As a user
I want to be able to start a batch from the batch overview page
So that I don't have to navigate between two pages

### AC
- [ ] A new batch can be created from a modal on the batch (calculations?) page
- [ ] UX 
- [x] #1880
- [x] #1879

MadsBloendAndersen:
@renehossle could you maybe paste the UX for this issue, I cant find it.

PR #1866, opened 2023-06-27
feat(dh): move start batch to search page

ManBearTM:
## Description

This PR moves the start batch functionality into the search batches page.

## References

- #1796


libs/ui-watt/src/lib/components/chip/watt-filter-chip.component.ts
```diff
@@ -19,13 +19,14 @@ import { Component, EventEmitter, Input, Output } from '@angular/core';
...
 import { WattChipComponent } from './watt-chip.component';
+import { useIsFirstRender } from '../../utils/lifecycle/use-is-first-render';
 
 @Component({
   standalone: true,
   imports: [WattChipComponent, WattIconComponent],
   selector: 'watt-filter-chip',
   template: `
-    <watt-chip [disabled]="disabled" [selected]="input.checked">
+    <watt-chip [disabled]="disabled" [selected]="isFirstRender() ? selected : input.checked">
```
frontend-specialisten, line 29: Gotcha! :D ... Suggestion: even though I praise the creativity, I will recommend you take another approach in Angular due to performance. You can try creating a console.log inside the function, and I will guess it goes crazy. Maybe you can use `ngOnInit` or consider creating a pure pipe for this. Also, I can recommend you this presentation by Minko where he shows some of the challenges of performance optimization in Angular.

https://www.youtube.com/watch?v=f8sA-i6gkGQ

libs/dh/wholesale/feature-search/src/lib/start/dh-wholesale-start.component.ts
```diff
@@ -94,15 +93,19 @@ export class DhWholesaleStartComponent implements OnInit, AfterViewInit, OnDestr
   private executionTypeChanged$ = new Subject<void>();
   private destroy$ = new Subject<void>();
 
-  @ViewChild('modal') modal!: WattModalComponent;
+  @ViewChild('modal') modal?: WattModalComponent;
+  @ViewChild('form') form?: NgForm;
```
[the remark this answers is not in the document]
ManBearTM: Good catch, it was leftover from when I tried to reset the form. Removed in 0b67760ef6611e29d4a00da5c016217d16b14c6e

libs/ui-watt/src/lib/components/chip/watt-filter-chip.component.ts
ManBearTM: I would totally agree with you if the function being called was computationally heavy, but this is an [extremely simple function](https://github.com/Energinet-DataHub/greenforce-frontend/blob/64ee482a7f218997427ec3221c8712e3793e4e15/libs/ui-watt/src/lib/utils/lifecycle/use-is-first-render.ts) that checks a single variable and then immediately returns. I doubt this is any different from the way pipes work, since they also have to run a function with a comparison to check if the input value has changed. But I was curious, so I made a performance test profiling setting a variable in ngOnInit vs useIsFirstRender. At 10.000.000 iterations of `cdr.detectChanges()` there were no noticeable difference, both methods taking between 1600-1700 ms for 10.000.000 change detections.

frontend-specialisten (approved):

PR #1887, opened 2023-07-03
refac(dh): remove wholesale store

ManBearTM:
## Description

Less is more.


libs/dh/wholesale/domain/src/streamToFile.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/** Streams data to a file for downloading. */
+export const streamToFile =
+  ({ name, type }: Options) =>
+  (data: unknown): Observable<void> =>
+    new Observable((observer) => {
+      const blobPart = data as BlobPart;
+      const blob = new Blob([blobPart], { type });
+      const basisData = window.URL.createObjectURL(blob);
```
mimse, line 33: Question: is URL, not top level?

libs/dh/wholesale/feature-settlement-reports/src/lib/tabs/dh-wholesale-settlements-reports-tabs-balance.component.ts
```diff
@@ -208,28 +211,30 @@ export class DhWholesaleSettlementsReportsTabsBalanceComponent
...
-      () => {
-        this.toastService.dismiss();
-      }
-    );
+
+    this.httpClient
+      .v1WholesaleSettlementReportDownloadGet(
+        gridAreas.map((g) => g.id),
+        WholesaleProcessType.BalanceFixing,
+        this.searchForm.controls.period.value?.start ?? '',
```
mimse, line 225: Question: Should we keep the named parameters?

mimse (approved):
Approved, with some questions :-)

libs/dh/wholesale/feature-settlement-reports/src/lib/tabs/dh-wholesale-settlements-reports-tabs-balance.component.ts
ManBearTM: Would love to, but I'm using the generated `httpClient` directly here. We could consider using the `useObjectParameters` setting for the openapi generator (https://openapi-generator.tech/docs/generators/typescript/#config-options). Perhaps in a separate PR, since this will require a fair amount of refactoring? 🙂

libs/dh/wholesale/domain/src/streamToFile.ts
ManBearTM: Good point, I just copied the code from the store. I have removed the `window` prefix 👍

mimse (approved):

PR #1892, opened 2023-07-04
Feature/new actors

dzhavat:
### DataHub
- generate `market-participant/actors/shell` library
- add feature flag for `market-participant/actors`
- add link to sidebar navigation to `market-participant/actors` page
- add redirect config to `market-participant` route

### Workspace
- add `.angular` cache to Prettier ignore 

## References

- https://github.com/Energinet-DataHub/team-titans/issues/230


libs/dh/market-participant/shell/src/lib/dh-market-participant-shell.routes.ts
```diff
@@ -30,9 +34,40 @@ import {
...
 export const dhMarketParticipantShellRoutes: Routes = [
+  {
+    path: '',
+    pathMatch: 'full',
+    redirectTo: dhMarketParticipantActorsPath,
+  },
+  {
+    path: dhMarketParticipantActorsPath,
+    canMatch: [
+      () => {
```
mimse, line 48: Question: Is it okay that this function has two diffent return types

mimse (approved):

libs/dh/market-participant/shell/src/lib/dh-market-participant-shell.routes.ts
dzhavat: Yes. returning `true` means that the navigation can proceed. Returning a `UrlTree` means that Angular will cancel the current navigation and make a redirect to another page.

PR #1893, opened 2023-07-05
refac(dh): rename batch to calculation

ManBearTM:
## Description

The term "batch" is outdated and should be replaced by "calculation".

## References

- [#1280](https://github.com/Energinet-DataHub/opengeh-wholesale/issues/1280)


dzhavat (approved):
🚀

PR #1896, opened 2023-07-06
Replaced ngx-slider with own implementation

mimse:
Made a new slider to replace ngx-slider


libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
```diff
@@ -52,33 +56,86 @@ export class WattSliderComponent {
...
-      enforceStep: false,
-      enforceStepsArray: false,
-      step: this.step,
-      stepsArray: this.customSteps?.map((value) => ({ value })),
-    };
+  ngAfterViewInit(): void {
+    const maxRangeElement = this.maxRange.nativeElement;
+    const minRangeElement = this.minRange.nativeElement;
+    const maxChanged$ = fromEvent(maxRangeElement, 'change');
+    const minChanged$ = fromEvent(minRangeElement, 'change');
```
dzhavat, line 79: issue: looks like the `change` event type comes with the downside that it only emits once when the user is done moving the slider. This means that the value shown in the time input field will not change continuously but only once. This makes it hard to know the current time value as the user moves the slider, and also to select a specific value "on the first try". Maybe going back to `input` event is better in this case because it emits values continuously.

libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
```diff
@@ -52,33 +56,86 @@ export class WattSliderComponent {
...
   /** The currently selected range value. */
   @Input() value: WattSliderValue = { min: this.min, max: this.max };
 
+  @ViewChild('maxRange') maxRange!: ElementRef<HTMLInputElement>;
+
+  @ViewChild('minRange') minRange!: ElementRef<HTMLInputElement>;
+
+  private _maxChangeSubscription!: Subscription;
+  private _minChangeSubscription!: Subscription;
+  private _providedStep = this.step;
```
dzhavat, line 68: issue: the `_providedStep` will be set to 1 initially (default value of `step`) and will not be updated when new values arrive in `step`. In order to update `_providedStep `, `onChanges` hook can be used (or the `step` Input is converted to a setter).

libs/ui-watt/src/lib/components/slider/watt-slider.component.ts
```diff
@@ -52,33 +56,86 @@ export class WattSliderComponent {
...
+    this._minChangeSubscription.unsubscribe();
+  }
+
+  private updateRange(minValue: number, maxValue: number) {
+    const rangeDistance = this.max - this.min;
+    const fromPosition = minValue - this.min;
+    const toPosition = maxValue - this.min;
+
+    const sliderColor = 'var(--watt-color-secondary-light)';
+    const rangeColor = 'var(--watt-color-primary)';
```
dzhavat, line 129: nitpick: suggestion: we have a `WattColorHelperService` that can be used for getting color values instead of hard-coding them. It's worth having a look.

dzhavat (approved):

PR #1901, opened 2023-07-07
feat(dh): shell grid layout

ManBearTM:
## Description

This PR tries to implement what #1297 failed to do.

### DIFF
--- a/libs/dh/wholesale/feature-calculations/src/lib/calculations.component.scss
+++ b/libs/dh/wholesale/feature-calculations/src/lib/calculations.component.scss
@@ -17,8 +17,21 @@
 
 @use "@energinet-datahub/watt/utils" as watt;
 
+:host {
+  position: absolute;
+  inset: 0;
+}
+
 .wrapper {
   @include watt.space-inset-l;
+  height: 100%;
+}
+
+watt-card {
+  display: grid;
+  grid-template-rows: auto auto minmax(0, 1fr);
+  grid-template-columns: 100%;
+  height: 100%;
 }
 
 watt-card-title {
--- a/libs/dh/wholesale/feature-calculations/src/lib/details/details.component.scss
+++ b/libs/dh/wholesale/feature-calculations/src/lib/details/details.component.scss
@@ -28,6 +28,11 @@ watt-description-list {
   margin-left: var(--watt-space-ml);
 }
 
+watt-drawer-content {
+  display: grid;
+  grid-template-rows: auto minmax(24rem, 1fr);
+}
+
 dh-calculations-grid-areas {
   margin: 0 var(--watt-space-ml);
 }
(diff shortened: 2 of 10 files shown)


libs/dh/wholesale/feature-calculations/src/lib/calculations.component.scss
```diff
@@ -17,8 +17,21 @@
...
   @include watt.space-inset-l;
+  height: 100%;
+}
+
+watt-card {
+  display: grid;
+  grid-template-rows: auto auto minmax(0, 1fr);
+  grid-template-columns: 100%;
+  height: 100%;
 }
```
dzhavat, line 35: question: does this "pattern" need to be used throughout all pages in order to achieve the same/similar layout? So `position: absolute` on the outermost page element, `height: 100%` on a wrapper, and `display: grid` on a child? Not relevant currently but we should consider how to make sure this is applied correctly throughout the app.

libs/dh/wholesale/feature-calculations/src/lib/details/details.component.scss
```diff
@@ -28,6 +28,11 @@ watt-description-list {
   margin-left: var(--watt-space-ml);
 }
 
+watt-drawer-content {
+  display: grid;
+  grid-template-rows: auto minmax(24rem, 1fr);
+}
+
```
dzhavat, line 35: question: what's the purpose of this? Doesn't it currently work somewhat similarly?

libs/dh/wholesale/feature-calculations/src/lib/calculations.component.scss
ManBearTM: Yep, pretty much. The `position: absolute` could also be replaced by `height: 100%` if desired. Perhaps we should consider creating a dh ui library for handling this specific table layout? 🙂

libs/dh/wholesale/feature-calculations/src/lib/details/details.component.scss
ManBearTM: Yeah, I started some work on the details drawer, but then later decided to do it later to keep the PR small 😄 So it sort of does nothing right now, but I'll need it later. Do you want me to remove it to keep the PR clean? 🙂

libs/dh/wholesale/feature-calculations/src/lib/calculations.component.scss
dzhavat: A library, or predefined CSS classes. Let's keep an eye in which direction things are going, then decide on a plan. But it'd be nice to apply the same style throughout.

libs/dh/wholesale/feature-calculations/src/lib/details/details.component.scss
dzhavat: Don't have to remove it. 

Well, it has a small side effect where that the `margin` on the `dl` elements inside `watt-description-list` in the drawer is now taken into account, whereas before the two `margin`s collapsed. That's why I was wondering about its purpose 😊

image

ManBearTM: Good catch! I have removed the change, that was not intended! Thanks for the help!

dzhavat (approved):

PR #1906, opened 2023-07-07
fix(dh): calculations empty state

ManBearTM:
## Description

This PR adds skeleton loading to watt-table and then uses it in calculations. Also, improves how the calculation page handles empty states (error and empty).


frontend-specialisten (dismissed):
Praise: I love it! The only thing I would love even more would be to simulate the loading in the storybook. However this should not hold the changes back in wholesale, so hereby approved :)

frontend-specialisten (approved):

PR #1909, opened 2023-07-10
fix(eo, transfers): adjust accessor of startDate of ta for table

frontend-specialisten:
## Description

Fix sorting of `startDate` under transfer agreements table


sahma19 (approved):
Approved

PR #1945, opened 2023-08-09
feat(dh): adjustments to actors page

dzhavat:
### DataHub
- rename translation property
- filter out irrelevant statuses in filter
- translate status values in filter
- translate market roles in table

## References

- https://github.com/Energinet-DataHub/team-titans/issues/215

### DIFF
--- a/libs/dh/market-participant/actors/feature-actors/src/lib/filters/dh-actors-filters.component.ts
+++ b/libs/dh/market-participant/actors/feature-actors/src/lib/filters/dh-actors-filters.component.ts
@@ -31,11 +31,7 @@ import { RxPush } from '@rx-angular/template/push';
 
 import { WATT_FORM_FIELD } from '@energinet-datahub/watt/form-field';
 import { WattButtonComponent } from '@energinet-datahub/watt/button';
-import {
-  WattDropdownComponent,
-  WattDropdownOption,
-  WattDropdownOptions,
-} from '@energinet-datahub/watt/dropdown';
+import { WattDropdownComponent, WattDropdownOptions } from '@energinet-datahub/watt/dropdown';
 import { MarketParticipantEicFunction } from '@energinet-datahub/dh/shared/domain';
 import { ActorStatus } from '@energinet-datahub/dh/shared/domain/graphql';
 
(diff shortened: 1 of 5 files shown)


libs/dh/market-participant/actors/feature-actors/src/lib/filters/dh-actors-filters.component.ts
```diff
@@ -132,22 +128,26 @@ export class DhActorsFiltersComponent implements OnInit, OnDestroy {
...
-  private buildActorStatusOptions(): WattDropdownOptions {
-    return Object.keys(ActorStatus).map((key) => ({
-      displayValue: key,
-      value: key,
-    }));
+  private buildActorStatusOptions(): Observable<WattDropdownOptions> {
+    return this.transloco.selectTranslateObject('marketParticipant.actorsOverview.status').pipe(
+      map((statusTranslations) =>
+        Object.keys(ActorStatus)
+          .filter((key) => !(key === ActorStatus.New || key === ActorStatus.Passive))
```
mimse, line 135: Question: Should new not map to Active, and Passive to Inactive, or else some of the actors will be missing a status?

mimse (changes requested):

libs/dh/market-participant/actors/feature-actors/src/lib/filters/dh-actors-filters.component.ts
dzhavat: If "New" and "Passive" statuses are going away, they will also be removed from the `enum` at some point. So when the new "Create actor" flow is implemented, it will not be possible to set an actor as "New" or "Passive".

The issue now is that an actor can still be created/edited using the existing flow (under "Organisations"). So "New" or "Passive" status can still be set on an actor.

Given that the new "Create actor" flow is going to be implemented soon, maybe we can accept that some actors can be "New"/"Passive" in the table until the existing "Create/Edit actor" is migrated. I'm not sure how many are using the "Organisations" page right now, but there's only one organisation in T with one actor.

dzhavat: So if an actor is set to "New"/"Passive", they will be in the table initially, but will not be possible to filter by "New"/"Passive" using the status filter. The "Status" cell will not show a badge though.

mimse (approved):

PR #1949, opened 2023-08-11
feat(dh): connect filters with actors table

dzhavat:
### DataHub
- create custom filter predicate function
- add search input
- connect filters and search input with actors table
- show fallback in table when market role property is not set
- update mock data
- update import of EicFunction
- add JSON util functions

## References

- https://github.com/Energinet-DataHub/team-titans/issues/215


libs/dh/market-participant/actors/feature-actors/src/lib/dh-actors-custom-filter-predicate.ts
```diff
@@ -0,0 +1,62 @@
...
+import { AllFiltersCombined } from './actors-filters';
+import { DhActor } from './dh-actor';
+import { dhParseJSON } from './dh-json-util';
+
+/**
+ * Custom filter predicate function that runs for each actor in the table.
+ *
+ * @returns `true` if an actor should be shown in the table, `false` otherwise.
+ */
+export const dhActorsCustomFilterPredicate = () => {
```
ManBearTM, line 28: question: why is this a higher-order function?

libs/dh/market-participant/actors/feature-actors/src/lib/dh-actors-overview.component.ts
```diff
@@ -102,9 +118,25 @@ export class DhActorsOverviewComponent implements OnInit, OnDestroy {
         this.loading = false;
       },
     });
+
+    this.dataSource.filterPredicate = dhActorsCustomFilterPredicate();
+
+    const filtersCombined$: Observable<AllFiltersCombined> = combineLatest([
+      this.filters$,
```
ManBearTM, line 125: question: have you considered moving the filtering to GraphQL? or is the data set small enough that we can always fetch every row?

libs/dh/market-participant/actors/feature-actors/src/lib/dh-actors-custom-filter-predicate.ts
dzhavat: No particular reason. I think I started with it initially, and kept it afterwards. Will get rid of it :)

libs/dh/market-participant/actors/feature-actors/src/lib/dh-actors-overview.component.ts
dzhavat: Yes, we did. The data set will be around 200 records, so "it shouldn't" be a problem to filter on the client. If we notice perf issues, we'll reconsider :)

libs/dh/market-participant/actors/feature-actors/src/lib/dh-actors-custom-filter-predicate.ts
```diff
@@ -0,0 +1,62 @@
...
+      showActor = filters.marketRoles.includes(actor.marketRole as EicFunction);
+    }
+
+    if (showActor && filters.searchInput !== '') {
+      showActor =
+        actor.glnOrEicNumber.includes(filters.searchInput) ||
+        actor.name.toLocaleLowerCase().includes(filters.searchInput.toLocaleLowerCase());
+    }
+
+    return showActor;
```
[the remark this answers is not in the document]
dzhavat: Nice suggestion 💯  Let's go with it.

The filtering is a bit tricky because the function runs for each row in the table. So all filters + input should be accounted for, also when some or all of them are set. That's why I tried adding some tests because the function is fairly complex. :)

ManBearTM (approved):

Issue #1930, opened 2023-08-02
Remove material legacy components


mimse:
We need to remove material legacy components. So we can upgrade to angular 17 when it lands.

Components to be converted

- [x] - legacy-card (EO) PR #2051
- [x] - legacy-paginator (EO) @frontend-specialisten
- [x] - legacy-table (EO) @frontend-specialisten 
- [x] - legacy-slide-toggle (EO) PR #2053
- [x] - legacy-radio (EO) PR #2054
- [x] - legacy-dialog
- [x] - legacy-snack-bar
- [x] - legacy-select/testing (mvt)
- [x] - legacy-select (mvt)
- [x] - legacy-paginator (mvt)
- [x] - legacy-table
- [x] - legacy-tooltip
- [x] - legacy-menu
- [x] - legacy-button (EO) [PR #2010](https://github.com/Energinet-DataHub/greenforce-frontend/pull/2010)
- [x] - legacy-checkbox
- [x] - legacy-form-field
- [x] - legacy-input
- [x] - legacy-list
- [x] - legacy-progress-spinner

PR #1956, opened 2023-08-15
watt-field, with text field

mimse:
Added a new way to handle over form fields and inputs

This only handles input in the text form, not radio and checkbox, select etc.

I have made use of the :has selector in css. I know it is not supported in firefox just yet. But it should land soon.


libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,163 @@
...
+  <watt-field [label]="label">
+    <watt-text-field [required]="required" [type]="type" [placeholder]="placeholder" [formControl]="exampleFormControl"></watt-text-field>
+  </watt-field>
+</div><p>Value: {{exampleFormControl.value}}</p>`;
+
+const howToUseGuideBasic = `
+How to use
+
+1. Import ${WattTextFieldComponent.name} in a module or component
+
```
dzhavat, line 53: suggestion: let's not promote modules anymore :)


```suggestion
1. Import ${WattTextFieldComponent.name} in a component

```

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,163 @@
...
+import { ${WattTextFieldComponent.name} } from '@energinet-datahub/watt/input';
+
+2. Create FormControl in a component
+
+exampleFormControl = new FormControl(true);
```
dzhavat, line 58:
```suggestion
exampleFormControl = new FormControl('');
```

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,163 @@
...
+    <watt-field [label]="label">
+      <watt-icon wattPrefix name="search" label="some meaningful description"/>
+      <watt-text-field [required]="required" [type]="type" [placeholder]="placeholder" [formControl]="exampleFormControl"></watt-text-field>
+    </watt-field>
+  </div>`,
+});
+
+export const WithSuffix: StoryFn<WattTextFieldComponent> = () => ({
+  props: {
+    label: 'prefix',
```
dzhavat, line 154:
```suggestion
    label: 'Suffix',
```

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,163 @@
...
+  },
+  template,
+});
+
+export const WithPrefix: StoryFn<WattTextFieldComponent> = () => ({
+  props: {
+    label: 'prefix',
+    exampleFormControl: new FormControl(null),
+  },
+  template: `<div style="display: flex; gap: var(--watt-space-m); flex-direction: column; margin-bottom: var(--watt-space-m);">
```
dzhavat, line 144: question: doesn't look like the inline styling here is necessary, is it?

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,163 @@
...
+    </watt-field>
+  </div>`,
+});
+
+export const WithSuffix: StoryFn<WattTextFieldComponent> = () => ({
+  props: {
+    label: 'prefix',
+    exampleFormControl: new FormControl(null),
+  },
+  template: `<div style="display: flex; gap: var(--watt-space-m); flex-direction: column; margin-bottom: var(--watt-space-m);">
```
dzhavat, line 157: question: doesn't look like the inline styling here is necessary, is it?

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,163 @@
...
+        WattFieldComponent,
+        WattButtonComponent,
+      ],
+    }),
+  ],
+};
+
+export default meta;
+
+const template = `<div style="display: flex; gap: var(--watt-space-m); flex-direction: column; margin-bottom: var(--watt-space-m);">
```
dzhavat, line 43: question: doesn't look like the inline styling here is necessary, is it?

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,163 @@
...
+import { Meta, moduleMetadata, StoryFn } from '@storybook/angular';
+import { FormControl, ReactiveFormsModule } from '@angular/forms';
+
+import { WattIconComponent } from '../../foundations/icon/icon.component';
+import { WattTextFieldComponent } from './watt-text-field.component';
+import { WattFieldComponent } from '../field';
+import { WattButtonComponent } from '../button';
+
+const meta: Meta<WattTextFieldComponent> = {
+  title: 'Components/Text Field',
```
dzhavat, line 26: suggestion: looks like the documentation for the now "legacy" text field component is applied to this component. Maybe because the `storybook-input-overview.mdx` contains a Meta title with the same path as this component. Update the Meta title in that file to point to the legacy component instead.

image

libs/ui-watt/src/lib/components/field/watt-field.component.scss
```diff
@@ -0,0 +1,69 @@
...
+  --watt-field-disabled-background-color: var(--watt-color-neutral-grey-300s);
+  --watt-input-border-radius: 4px;
+  --watt-border-color: var(--watt-on-light-low-emphasis);
+  --watt-label-text-color: var(--watt-text-color);
+
+  label {
+    color: var(--watt-label-text-color);
+    // :has should be implmeneted in firefox before we go live
+    // https://connect.mozilla.org/t5/ideas/when-is-has-css-selector-going-to-be-fully-implemented-in/idi-p/23794/page/3#comments
+    &:has(.disabled) .label {
```
ManBearTM, line 31: question: should this be `:disabled`?

libs/ui-watt/src/lib/components/field/watt-field.component.scss
```diff
@@ -0,0 +1,69 @@
...
+        --watt-border-color: var(--watt-color-primary);
+      }
+
+      &:has(.disabled) {
+        background-color: var(--watt-field-disabled-background-color);
+      }
+
+      &:has(.ng-dirty.ng-invalid) {
+        --watt-border-color: var(--watt-color-state-danger);
+      }
```
ManBearTM, line 66: nit: like we talked about, this `:has` selector tightly couples the `watt-field` and the implementers (`watt-text-field` etc) in a way that is a bit difficult to understand. keeping the `watt-field` as an "internal only" component makes this more acceptable, but if we are allowing `ng-content` from `watt-text-field` to pass through to `watt-field`, we could run into weird issues like this:

```html
<watt-text-field>
  <watt-button [disabled]="condition"></watt-button>
</watt-text-field>
```

perhaps we can avoid this with some clever use of content projection 🤔

libs/ui-watt/src/lib/components/field/watt-field.component.scss
```diff
@@ -0,0 +1,69 @@
...
+
+    &:has(.ng-dirty.ng-invalid) .label {
+      color: var(--watt-color-state-danger);
+    }
+
+    &:has(.ng-dirty.ng-invalid, [required]) .label::after {
+      content: " *";
+    }
+
+    .watt-field {
```
ManBearTM, line 43: nitpick: is `watt-field-input` a better name? i got confused thinking this was the component selector.

libs/ui-watt/src/lib/components/text-field/watt-text-field.component.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  /* @ignore */
+  registerOnTouched(fn: () => void): void {
+    this.onTouched = fn;
+  }
+
+  /* @ignore */
+  setDisabledState(isDisabled: boolean): void {
+    this.isDisabled = isDisabled;
+  }
```
ManBearTM, line 88: question: just curios, why are the `registerOnTouched` and `registerOnChange` not properly implemented, like in [`chip.directive.ts`](https://github.com/Energinet-DataHub/greenforce-frontend/blob/main/libs/ui-watt/src/lib/components/form-field/chip.directive.ts)? is this a common pattern that i'm not aware of? 🙂

tsconfig.base.json
```diff
@@ -437,6 +437,12 @@
       "@energinet-datahub/watt/validators": [
         "libs/ui-watt/src/lib/components/input/shared/validators/index.ts"
       ],
+      "@energinet-datahub/watt/watt-field": [
+        "libs/ui-watt/src/lib/components/field/index.ts"
+      ],
```
ManBearTM, line 442: consider not exposing this if we choose to use `watt-field` directly within `watt-text-field`.

libs/ui-watt/src/lib/components/field/watt-field.component.scss
mimse: But it would also contain a select at somepoint?

libs/ui-watt/src/lib/components/text-field/watt-text-field.component.ts
mimse: It is a copy / paste from radio button

libs/ui-watt/src/lib/components/field/watt-field.component.scss
mimse: Renamed it to .watt-field-wrapper

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,164 @@
...
+             </watt-text-field>`,
+});
+
+export const WithHint: StoryFn<WattTextFieldComponent> = () => ({
+  props: {
+    label: 'hint',
+    exampleFormControl: new FormControl(null),
+  },
+  template: `<watt-text-field [required]="true" [label]="label" [type]="type" [placeholder]="placeholder" [formControl]="exampleFormControl">
+              <span error>This field is required</span>
```
dzhavat, line 161: question: do we want to support two ways of showing the error message - (1) by using an `ngIf` on the error (like the `WithRequired` story) (which adds/removes the error element from the DOM) and (2) by simply adding the `error` attribute on an element (so the element is shown/hidden with CSS)?

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,164 @@
...
+              <span hint>This is a hint</span>
```
dzhavat, line 162: suggestion: we discussed not creating components for `error` and `hint` but now I can see why they'd be useful - for consistency. Without such elements, one wonders "Should I add the `error`/`hint` attribute on a `p`, `span`, `div` or another element?". With dedicated components it's pretty clear what should be used. What do you think?

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,164 @@
...
+            </watt-text-field>`,
```
dzhavat, line 163: issue: there's a small jump when the component goes from normal to error state (when the "hint" gets hidden and "error" is shown).

issue

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
```diff
@@ -0,0 +1,164 @@
...
+  template,
+});
+
+export const WithRequired: StoryFn<WattTextFieldComponent> = () => ({
+  props: {
+    label: 'required',
+    exampleFormControl: new FormControl(null),
+  },
+  template: `<watt-text-field [required]="true" [label]="label" [type]="type" [placeholder]="placeholder" [formControl]="exampleFormControl">
+              <span info *ngIf="exampleFormControl.dirty && exampleFormControl.hasError('required')">This field is required</span>
```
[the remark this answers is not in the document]
mimse: You are right, good catch

libs/ui-watt/src/lib/components/text-field/watt-text-field.stories.ts
mimse: Only with an error attribute, i forgot to remove it :-)

dzhavat (approved):

PR #1962, opened 2023-08-17
Handle terms differently

CodeReaper:


libs/eo/shared/atomic-design/feature-molecules/src/lib/eo-privacy-policy/eo-privacy-policy.component.ts
```diff
@@ -89,6 +94,14 @@ import { EoScrollViewComponent } from '@energinet-datahub/eo/shared/atomic-desig
 })
 export class EoPrivacyPolicyComponent {
   privacyPolicy$ = this.http.get('/assets/html/privacy-policy.html', { responseType: 'text' });
+  bag$ = this.http.get<VersionResponse>('/assets/configuration/privacy-policy.json').subscribe({
```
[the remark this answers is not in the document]
CodeReaper: @frontend-specialisten - It seems there are some git ignore rules so I did not notice the file was missing, but also yes! This file is replaced during development so it contains the correct current version.

libs/eo/terms/src/lib/eo-terms.component.ts
```diff
@@ -99,16 +114,31 @@ import { WattSpinnerComponent } from '@energinet-datahub/watt/spinner';
...
+  );
+
+  loadingPrivacyPolicyFailed = false;
+  privacyPolicy$ = this.policyVersion$.pipe(
+    switchMap((response) => {
+      this.termsService.setVersion(response.version);
+      return this.http.get('/assets/html/privacy-policy.html', { responseType: 'text' });
+    }),
+    catchError(() => {
+      this.termsService.setVersion(-1); // Is this needed anymore?
```
CodeReaper, line 132: @frontend-specialisten 
I would argue it is needed to ensure the version is kept in sync with the text of the privacy policy. Otherwise the following sequence is possible:

- client: fetch policy and version
- client: view the policy
- server: update the policy and version
- client: fetch new policy and fail to fetch version
- client: accept the new policy, but register the old version as accepted

It is unlikely to happen, but it seems like a nice safety net to have.

CodeReaper, line 132:
```suggestion
      this.termsService.setVersion(-1);
```

PR #1966, opened 2023-08-18
Test: added id to archivedMessages

RasmusGraabaek:
## References

- #0000


libs/dh/message-archive/feature-log-search/src/lib/drawer/dh-message-archive-drawer.component.ts
```diff
@@ -104,7 +104,7 @@ export class DhMessageArchiveDrawerComponent implements OnInit, OnDestroy {
     this.drawer.open();
 
     if (this.message) {
-      this.getDocument(this.message?.messageId);
+      this.getDocument(this.message?.id);
```
dzhavat, line 107: nitpick: `?` not necessary since `id` will always contain a value.

```suggestion
      this.getDocument(this.message.id);
```

libs/dh/message-archive/feature-log-search/src/lib/drawer/dh-message-archive-drawer.component.ts
```diff
@@ -113,9 +113,9 @@ export class DhMessageArchiveDrawerComponent implements OnInit, OnDestroy {
     this.message = null;
   }
 
-  getDocument(messageId: string) {
+  getDocument(id: string) {
     this.apiStore.getDocument({
-      id: messageId,
+      id: id,
```
dzhavat, line 118: nitpick: when an object property matches with the name of a variable that we want to assign to it, a shorthand notation can be used.

```suggestion
      id,
```

libs/dh/shared/data-access-msw/src/lib/mocks/data/messageArchiveSearchResponseLogs.ts
```diff
@@ -16,98 +16,128 @@
...
 export const messageArchiveSearchResponseLogs: SearchResult = {
   messages: [
     {
-      messageId: '38374f50-f00c-4e2a-aec1-70d391cade06',
+      messageId: messageId1,
       documentType: 'NotifyAggregatedMeasureData',
       createdDate: '2023-06-14T12:10:54.9519045+00:00',
       senderGln: '5790001330552',
       receiverGln: '7080000729821',
+      id: messageId1,
```
dzhavat, line 37: question: are values for `id` and `messageId` supposed to match? Some of the objects here have the same values, where others don't :)

RasmusGraabaek: Yes. The "id" and "messageId" will match whenever the archived message is a response from IDE.
But they will most likely not match, whenever the archived message is a request from an actor. Since the messageId will be the identifier of their request and the id will be something we add for uniqueness

dzhavat (approved):

PR #1979, opened 2023-08-23
feat(dh): actors drawer component

dzhavat:
### DataHub
- show drawer with details when actor is selected in table

## References

- https://github.com/Energinet-DataHub/team-titans/issues/225

### DIFF
--- a/libs/dh/market-participant/actors/feature-actors/src/lib/drawer/dh-actor-drawer.component.html
+++ b/libs/dh/market-participant/actors/feature-actors/src/lib/drawer/dh-actor-drawer.component.html
@@ -1,3 +1,19 @@
+<!--
+@license
+Copyright 2020 Energinet DataHub A/S
+
+Licensed under the Apache License, Version 2.0 (the "License2");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


libs/dh/market-participant/actors/feature-actors/src/lib/drawer/dh-actor-drawer.component.html
```diff
@@ -0,0 +1,72 @@
...
+limitations under the License.
+-->
+<ng-container *transloco="let t; read: 'marketParticipant.actorsOverview.drawer'">
+  <watt-drawer #drawer (closed)="onClose()">
+    <watt-drawer-topbar>
+      <dh-actor-status-badge [status]="actor?.status" />
+    </watt-drawer-topbar>
+
+    <watt-drawer-heading>
+      <h2 class="actor-heading">{{ actor?.name }}</h2>
```
mimse, line 24: Suggestion for another day. Maybe we should have a title / label on the watt-drawer-heading component, and move the h tag into the component. 
I never know what heading is right to put in, this would fix that :-)

mimse (approved):

libs/dh/market-participant/actors/feature-actors/src/lib/drawer/dh-actor-drawer.component.html
dzhavat: Good idea. Yeah, so far we've used `h1`, `h2` and `h3`. So I think we're all confused as to which heading level to use :D

PR #1980, opened 2023-08-23
Feature/remove legeacy mat checkbox

mimse:
Replaced mat-checkbox with our own implementation

It does not match 100% our current look, but I think it´s better :D

Also removed the hover, click effects from material, and after they got removed, I saw no need for the state stories any more. So I removed them as well.


libs/ui-watt/src/lib/components/checkbox/watt-checkbox.component.scss
```diff
@@ -0,0 +1,69 @@
...
+  input {
+    appearance: none;
+    margin: 0;
+    padding: 0;
+    width: 1.15em;
+    height: 1.15em;
+    border: 0.15em solid var(--watt-checkbox-color);
+    display: grid;
+    place-content: center;
+  }
```
dzhavat, line 33: issue: the new checkbox is a bit square-ish compared to Material (and in Figma).

Let's round the corners a bit (2px I believe) and make the checkmark a bit thinner.

New / Current
image

Design in Figma ([link](https://www.figma.com/file/fwEDQkgEcpwfUojXl0EIC5/Watt-Design-System?type=design&node-id=4761-12635&mode=design&t=vcMKrq08ouepUVuD-0)) 

image

libs/ui-watt/src/lib/components/checkbox/watt-checkbox.component.ts
```diff
@@ -14,103 +14,68 @@
...
+  template: `<label>
+    <input
+      [(ngModel)]="model"
+      [disabled]="isDisabled"
+      (ngModelChange)="onChange($event)"
+      type="checkbox"
+    />
+    <ng-content />
+  </label>`,
+  standalone: true,
```
dzhavat, line 39: question: Material's checkbox offered "indeterminate" state. Can we implement that here as well?

From Figma 

image

libs/ui-watt/src/lib/components/checkbox/watt-checkbox.component.ts
```diff
@@ -14,103 +14,68 @@
...
+      multi: true,
+    },
   ],
-  templateUrl: './watt-checkbox.component.html',
-  providers: [customValueAccessor],
-  standalone: true,
-  imports: [MatCheckboxModule, ReactiveFormsModule],
+  imports: [FormsModule],
 })
 export class WattCheckboxComponent implements ControlValueAccessor {
```
dzhavat, line 50: issue: validation is missing. Should that be part of this PR?

From Figma

image

libs/ui-watt/src/lib/components/checkbox/watt-checkbox.component.scss
```diff
@@ -0,0 +1,69 @@
...
+  &[disabled="true"] label {
+    color: var(--watt-checkbox-disabled-color);
+  }
+
+  label {
+    display: grid;
+    grid-template-columns: 1em auto;
+    gap: 0.5em;
+    align-items: center;
+  }
```
dzhavat, line 68: nitpick: let's change the mouse cursor to hand on hover, indicating that the element can be clicked.

nitpick: on double click, the label text gets selected. Can see that Material is using `user-select: none;` to disable that. Should we add it too?

libs/ui-watt/src/lib/components/checkbox/watt-checkbox.component.ts
mimse: I´m missing that :-)

dzhavat (dismissed):

PR #1992, opened 2023-08-29
feat(dh): update to actor drawer

dzhavat:
### DataHub
- show grid area in drawer
- show organisation in drawer
- update mock data
- configure actor to use partial data from in-memory cache

## References

- https://github.com/Energinet-DataHub/team-titans/issues/225


libs/dh/market-participant/actors/feature-actors/src/lib/dh-actor.ts
```diff
@@ -16,6 +16,10 @@
...
 import type { ResultOf } from '@graphql-typed-document-node/core';
 
-import { GetActorsDocument } from '@energinet-datahub/dh/shared/domain/graphql';
+import {
+  GetActorByIdDocument,
+  GetActorsDocument,
+} from '@energinet-datahub/dh/shared/domain/graphql';
 
 export type DhActor = ResultOf<typeof GetActorsDocument>['actors'][0];
+export type DhActorExtended = ResultOf<typeof GetActorByIdDocument>['actor'];
```
mimse, line 25: I think the list actor, and detailed actor is the same type.

libs/dh/market-participant/actors/feature-actors/src/lib/drawer/dh-actor-drawer.component.ts
```diff
@@ -102,4 +110,25 @@ export class DhActorDrawerComponent {
...
+    this.getActorByIdQuery$.setVariables({ id });
+
+    this.subscription = this.getActorByIdQuery$.valueChanges
+      .pipe(takeUntil(this.closed))
+      .subscribe({
+        next: (result) => {
+          this.actor = result.data?.actor;
+        },
+        error: (error) => {
+          console.log('error', error);
```
mimse, line 130: Suggestion: Maybe show the an error in the UI?

libs/dh/market-participant/actors/feature-actors/src/lib/dh-actor.ts
dzhavat: You're right, they are the same type but the GraphQL queries select different Actor properties for their respective views. Så using the `Actor` type for both will mean that a view has access to more `Actor` properties than what might be selected by a query.

That's how I understand the type - query relationship in GraphQL. Is it correct or am I missing something? :)

libs/dh/market-participant/actors/feature-actors/src/lib/drawer/dh-actor-drawer.component.ts
dzhavat: I was actually considering removing the error handler, because most of the data is already available from the table (using partial loading) and the rest has an em dash fallback in case it fails to load.

Do you think we should still show some kind of an error message?

mimse: Sounds like it is handled

libs/dh/market-participant/actors/feature-actors/src/lib/dh-actor.ts
mimse: Ah ofc, you are right :-)

mimse (approved):

PR #2006, opened 2023-09-01
feat(dh): use vater in calculations

ManBearTM:
## Description

Use vater components for layout in calculations.


libs/dh/wholesale/feature-calculations/src/lib/calculations.component.ts
```diff
@@ -22,9 +22,14 @@ import { TranslocoModule } from '@ngneat/transloco';
 import { sub, startOfDay, endOfDay } from 'date-fns';
 import { BehaviorSubject, Subject, switchMap, takeUntil } from 'rxjs';
 
+import {
+  VaterFlexComponent,
+  VaterSpacerComponent,
+  VaterStackComponent,
```
mimse, line 28: Suggestion: maybe we should have a VATER export with all of them components?

mimse (approved):

PR #2017, opened 2023-09-05
feat(watt): vater css layer

ManBearTM:
## Description

This PR moves all of the vater utility classes to a separate vater layer, making these classes always have the lowest specificy. This is useful since you do not want the utility classes to override any local styles.


frontend-specialisten (approved):
LGTM

dzhavat (approved):

PR #2021, opened 2023-09-05
experiment(watt): watt-heading

ManBearTM:
## Description

In an attempt to solve the issue regarding the special watt-card title, this PR takes a different approach. Instead of creating special components for each time headings need to be treated differently, this solution consolidates all exceptions within the watt-heading component by using the host-context selector.


ManBearTM:
Still think this could be relevant in the future, but closing it for now.

PR #2028, opened 2023-09-07
feat(watt): watt-data-table

ManBearTM:
## Description

Yet another attempt at the whole card/table abstraction problem.


libs/watt/src/lib/components/data/watt-data-table.component.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+      }
+    `,
+  ],
+  template: `
+    <watt-card vater fill="vertical">
+      <vater-flex fill="vertical" gap="m">
+        <vater-stack direction="row" gap="m">
+          <vater-stack direction="row" gap="s">
+            <ng-content select="h3" />
```
frontend-specialisten, line 83: Question: Does forcing either `h3` or `h4` make sense? Couldn't we have cases where the semantic leveling would be `h2`?

libs/watt/src/lib/components/data/watt-data-table.component.ts
```diff
@@ -0,0 +1,124 @@
...
+    WattEmptyStateComponent,
+    WattPaginatorComponent,
+    WattSearchComponent,
+  ],
+  encapsulation: ViewEncapsulation.None,
+  styles: [
+    `
+      watt-data-table h3,
+      watt-data-table h4 {
+        line-height: 44px;
```
frontend-specialisten, line 54: Question: So we have "headers" with different font sizes and potential other styles, since only the line height and height are styled here?

libs/watt/src/lib/components/table/watt-table.component.ts
```diff
@@ -291,7 +286,7 @@ export class WattTableComponent<T> implements OnChanges, AfterViewInit, OnDestro
   _element = inject<ElementRef<HTMLElement>>(ElementRef);
 
   /** @ignore */
-  _subscription!: Subscription;
+  private destroy$ = new Subject<void>();
```
frontend-specialisten, line 289: Suggestion: use `takeUntilDestroyed ` instead :) 
https://indepth.dev/posts/1518/takeuntildestroy-in-angular-v16#takeuntildestroy

tsconfig.base.json
```diff
@@ -351,6 +351,9 @@
       "@energinet-datahub/watt/danish-date-time": [
         "libs/watt/src/lib/configuration/watt-danish-datetime.providers.ts"
       ],
+      "@energinet-datahub/watt/data": [
```
frontend-specialisten, line 354: Consideration: Of course, `data` is an excellent short import. However, does it make sense for everyone that this is the data-table entry point?

libs/watt/src/lib/components/data/watt-data-table.component.ts
ManBearTM: Good question 🤔 In the DataHub app, the heading at the top of the page is an `h2`, which is why I thought it could never be `h1` or `h2`. But that is of course specific to DataHub. Do you have a use case for `h2` (or even `h5` / `h6`) for EO? I don't mind adding other levels, but perhaps that is also something we could add when the need arises 🙂

libs/watt/src/lib/components/data/watt-data-table.component.ts
ManBearTM: Yes, I based that decision on the designs in Figma, which uses `h3` on the "main" view and `h4` in the drawer:
https://www.figma.com/file/ipM0Z3VUyrk0SQTvAfmfnw/Wholesale?type=design&node-id=378-64675&mode=design&t=WClfDU1vEJ3MVmW8-0

tsconfig.base.json
ManBearTM: I'm probably getting ahead of myself here, but I was imagining the `data` category to have other composite components for data presentation, such as a `watt-data-calendar` for example. But I do agree that the naming is a bit vague.

We could make a separate `@energinet-datahub/watt/data-table` entry point, but what about the imports that are not strictly for the table? For example (and again, getting ahead of myself 😄) the `watt-data-calendar` would also need the `watt-data-filters` as it has options for filtering what is shown on the calendar.

Do you have any suggestions?

mimse (approved):

libs/watt/src/lib/components/data/watt-data-table.component.ts
frontend-specialisten: Ah yes, of course that makes sense!

tsconfig.base.json
frontend-specialisten: I buy in on that plan. Let's keep it.

frontend-specialisten (approved):
Nice work!

PR #2051, opened 2023-09-14
refactor(eo): replaced mat-card with watt-card

frontend-specialisten:
## Description

- Replaced `mat-card` with `watt-card`

## References

https://github.com/Energinet-DataHub/energy-origin-issues/issues/1771


sahma19 (approved):
Approve

PR #2052, opened 2023-09-14
feat(dh): use new Vater components in Balance responsible tab

dzhavat:
### DataHub
- use new Vater components in Balance responsible tab

### Design System
- add height to tabs so new Vater components can work inside tabs

## References

- https://github.com/Energinet-DataHub/team-titans/issues/274


ManBearTM (approved):
Looks good!

PR #2058, opened 2023-09-15
Added new textarea component

mimse:
We were missing a textarea component

image

### DIFF
--- a/libs/watt/src/lib/components/textarea-field/watt-textarea-field.component.ts
+++ b/libs/watt/src/lib/components/textarea-field/watt-textarea-field.component.ts
@@ -1,0 +1,94 @@
+/**
+ * @license
+ * Copyright 2020 Energinet DataHub A/S
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License2");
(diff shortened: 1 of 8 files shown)


libs/watt/src/lib/components/textarea-field/watt-textarea-field.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  template: `<watt-field [label]="label">
+    <textarea
+      [attr.placeholder]="placeholder"
+      [value]="value"
+      [(ngModel)]="model"
+      [disabled]="isDisabled"
+      (ngModelChange)="onChange($event)"
+      [required]="required"
+    ></textarea>
+    <ng-content />
```
dzhavat, line 52: question: is there a use case where this can be necessary? In the text field component this can be used for a suffix button, but maybe that's not relevant here?

mimse: You are right, i forgot to remove it ! :-)

dzhavat (approved):

PR #2109, opened 2023-10-03
refac: Update user role audit log.

defectiveAi:
## References

- #0000


libs/dh/admin/feature-user-management/src/lib/drawer/roles/tabs/content/dh-role-audit-logs.component.ts
```diff
@@ -71,37 +79,58 @@ import { DhAuditChangeCellComponent } from './dh-audit-change-cell.component';
...
+    entry: { accessor: null },
   };
 
-  @Input() role: MarketParticipantUserRoleWithPermissionsDto | null = null;
+  @Input({ required: true }) role!: MarketParticipantUserRoleWithPermissionsDto;
 
   ngOnInit(): void {
-    this.store.auditLogs$.pipe(takeUntil(this.store.destroy$)).subscribe((logs) => {
-      this.dataSource.data = logs;
+    console.log('init called');
```
FirestarJes, line 105: Should this still be here?

michaelmajgaard (approved):

FirestarJes (changes requested):

libs/dh/admin/feature-user-management/src/lib/drawer/roles/tabs/content/dh-role-audit-logs.component.ts
defectiveAi: Fixed.

FirestarJes (approved):

libs/dh/admin/feature-user-management/src/lib/drawer/roles/tabs/content/dh-role-audit-logs.component.ts
```diff
@@ -71,37 +79,56 @@ import { DhAuditChangeCellComponent } from './dh-audit-change-cell.component';
...
+    });
   }
 
-  reloadAuditLogs(): void {
-    if (this.role) {
-      this.store.getAuditLogs(this.role.id);
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.role.currentValue) {
+      const id = changes.role.currentValue.id;
+      this.getUserRoleAuditLogsQuery?.refetch({ id });
```
mimse, line 127: @defectiveAi I pretty sure @dzhavat and I found some issues with caching and refetch. Can you remember what that was @dzhavat ?

mimse (commented):
Looks good, just one comment that i need @dzhavat eyes on :-)

libs/dh/admin/feature-user-management/src/lib/drawer/roles/tabs/content/dh-role-audit-logs.component.ts
dzhavat: Don't remember exactly 😞 

But I think the issue in our case was related to using partial data loading, which is not set up here (not needed).

The only "issue" I see is potentially double fetching of data, because a refetch is sent during `NgOnChange` and then a subscription is created during `ngOnInit`. There's only one way to find out :)

dzhavat (approved):

PR #2119, opened 2023-10-06
feat: Initial edit actor modal.

defectiveAi:


libs/dh/globalization/assets-localization/src/assets/i18n/en.json
```diff
@@ -632,7 +633,6 @@
       "MeteringPointAdministrator": "Metering point administrator",
       "SystemOperator": "System operator",
       "DanishEnergyAgency": "Energistyrelsen",
-      "ElOverblik": "ElOverblik",
```
dzhavat, line 635: issue: this translation is also used in the "Market roles" tab under "Actors". So if we really should get rid of it, then we also need to get rid of "EicFunction.ElOverblik" enum property in `.../generated/graphql.ts`

image

dzhavat (approved):

PR #2123, opened 2023-10-08
feat: handle invalid balance responsible data

michaelmajgaard:


libs/dh/esett/feature-balance-responsible/src/lib/table/dh-table.component.html
```diff
@@ -34,7 +34,12 @@
         let message
       "
     >
-      {{ message.supplierWithName?.value ?? "" }}
+      <ng-container *ngIf="message.supplierWithName">
+        {{ message.supplier.value }} - {{ message.supplierWithName.value }}
```
dzhavat, line 38: question: Wouldn't the supplier id be shown two times in this case? Once in `message.supplier.value` and then as part of the `supplierWithName` property?

Looking at the page, it seems like it. (or maybe the test data is not correct)

\* The same comment applies to other places as well, i.e. `gridArea` + `gridAreaWithName`, and `balanceResponsible` + `balanceResponsibleWithName`.

image

michaelmajgaard: `supplier.value` is the number and `supplierWithName.value` is the name. `supplierWithName` might be null if the number is invalid or not known to us for some reason.

dzhavat: Ok, so the test data is not correct then. Because `supplierWithName` currently includes both a number and name.

dzhavat (dismissed):

dzhavat (approved):

dzhavat (approved):

PR #2133, opened 2023-10-10
chore: update Prettier to latest v3

dzhavat:
**Update 2023-11-13**: The note bellow is not completely correct. If `prettier` package is installed as a `devDependencies` in a project, the Prettier VS Code extension will use the package version instead of the built-in one. So Prettier v3 can be used right away, though I had to make a small adjustment to the `prettier.prettierPath` setting.

#### Note 

Prettier VS Code extension v10.1.0 (latest stable as of 2023-10-10) is still using Prettier v2 for formatting. This means that any formatting done in CI - where Prettier v3 is used had this PR been merged - will be reverted to v2 format during local development - after file save. Because of this, the PR cannot be merged just yet. Prettier VS Code extension needs to support Prettier v3 before the PR can be merged.

### Workspace
- update Prettier to latest v3
- configure trailing comma to reflect previous behaviour
- format files
- update `prettier.prettierPath` Workspace setting. Otherwise Prettier VS Code extension was not working correctly.

## References

- N/A


frontend-specialisten (approved):
Nice work!

Issue #981, opened 2022-10-03
Open modal dynamically


dzhavat:
**Description:**
The only way to open a modal component now is by declaring it in the HTML of the component that is going to open the modal. We want to make it possible to open a modal dynamically from a component class (similar to how the [Dialog](https://material.angular.io/components/dialog/overview) component works in Angular Material).

**AC:**

- [x] Should be possible to open a modal dynamically from a class

**Definition of Ready:**

- [ ] The issue is correctly estimated
- [ ] The issue is adequately described
- [ ] Possible dependencies are defined and aligned
- [ ] We have the necessary skills to complete this issue
- [ ] The issue can be completed withing 1 iteration
- [ ] The issue has acceptance criteria defined
- [ ] The issue has adequate Definition Of Done described

**Definition of Done:**

- [ ] Acceptance Criteria have been met
- [ ] The product has been demo’ed for relevant stakeholders
- [ ] Dependencies are handled
- [ ] The work has been documented
- [ ] The issue has been handed over and reviewed
- [ ] The PO has accepted the product

PR #2141, opened 2023-10-11
Open modal from class

frontend-specialisten:
## Description
This PR makes it possible to open a modal by a class, ex—a service, without inlining it. 

## References

- #981


libs/watt/src/lib/components/modal/watt-modal.service.ts
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Closes the modal with `true` for acceptance or `false` for rejection.
+   * @ignore
+   */
+  close(result: boolean) {
+    this.closeSubject.next(result);
+  }
+}
+
+@NgModule({
```
mimse, line 97: Question: Module, why not standalone?

mimse (dismissed):

libs/watt/src/lib/components/modal/watt-modal.service.ts
frontend-specialisten: I'm not a fan of it either, however, I was not able to make it work due to DI not having this module (at least this is only for internal usage).

libs/watt/src/lib/components/modal/+storybook/watt-modal-from-class.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    setTimeout(() => {
+      this.isLoading = false;
+    }, 300);
+  }
+}
+
+@Component({
+  standalone: true,
+  selector: 'watt-modal-from-class',
```
dzhavat, line 55: nitpick: the component has the same selector as the previous component.

libs/watt/src/lib/components/modal/+storybook/watt-modal-from-class.component.ts
```diff
@@ -0,0 +1,71 @@
...
+import { FormControl } from '@angular/forms';
+
+@Component({
+  standalone: true,
+  selector: 'watt-modal-from-class',
+  imports: [WATT_MODAL, WattTextFieldComponent, WattButtonComponent],
+  template: `
+    <watt-modal #modal [title]="title" closeLabel="Close modal" [loading]="isLoading">
+      <watt-text-field [formControl]="exampleFormControl" label="Username" />
+      <watt-text-field [formControl]="exampleFormControl" label="Password" type="password" />
```
dzhavat, line 31: nitpick: the same control is used on both input fields.

libs/watt/src/lib/components/modal/+storybook/watt-modal-from-class.component.ts
```diff
@@ -0,0 +1,71 @@
...
+import { WattTextFieldComponent } from '../../text-field/watt-text-field.component';
+import { FormControl } from '@angular/forms';
+
+@Component({
+  standalone: true,
+  selector: 'watt-modal-from-class',
+  imports: [WATT_MODAL, WattTextFieldComponent, WattButtonComponent],
+  template: `
+    <watt-modal #modal [title]="title" closeLabel="Close modal" [loading]="isLoading">
+      <watt-text-field [formControl]="exampleFormControl" label="Username" />
```
dzhavat, line 30: issue: weird but when I type in the input, there's an error in the console.

image

libs/watt/src/lib/components/modal/+storybook/watt-modal-from-class.component.ts
frontend-specialisten: It should be fixed now; the selector of the component used in the modal has now been removed.

libs/watt/src/lib/components/modal/+storybook/watt-modal-from-class.component.ts
frontend-specialisten: Should be fixed now

libs/watt/src/lib/components/modal/+storybook/watt-modal-from-class.component.ts
frontend-specialisten: The error was caused by not importing `ReactiveFormsModule` (good catch!).

dzhavat (approved):

PR #2142, opened 2023-10-11
Preview/fe 1844 automatic logout

frontend-specialisten:
## Description

Automatic Logout now works when the user leaves the current tab

## References

https://github.com/Energinet-DataHub/energy-origin-issues/issues/1844


sahma19 (approved):
Tested in Preview. Works as intended. Approving

PR #2146, opened 2023-10-12
feat: Add search for documentId to Outgoing Messages

FirestarJes:


libs/dh/esett/feature-outgoing-messages/src/lib/filters/dh-filters.component.ts
```diff
@@ -68,6 +68,10 @@ type Filters = FormControls<DhOutgoingMessagesFilters>;
       watt-dropdown {
         width: auto;
       }
+
+      .documentId-field {
+        width: 20rem;
+      }
```
dzhavat, line 74: nitpick: doesn't look like this is used.

FirestarJes: Ohh, you are right, this was used for the textbox, just forgot to remove it.

FirestarJes: Fixed

dzhavat (approved):

PR #2152, opened 2023-10-16
Feat/eo idle use watt modal

frontend-specialisten:
## Description

- Use `watt-modal` in eo idle-timer, and remove custom styling
- Update styles for `watt-modal` to new styles provided by UX
- Rename size `normal` to `medium` of watt-modal, and remove the usage of old naming
- Make it possible to pass data to a `watt-modal` opened from a class

### DIFF
--- a/build/infrastructure/eo/configuration.yaml
+++ b/build/infrastructure/eo/configuration.yaml
@@ -1,5 +1,5 @@
 name: eo-frontend-app
-version: 0.3.385
+version: 0.3.386
 repo: ghcr.io/energinet-datahub
 references:
   - file: k8s/energy-origin-apps/frontend/base/deployment.yaml


libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-idle-timer/idle-timer-countdown.component.ts
```diff
@@ -16,80 +16,37 @@
...
+    private dialogRef: WattDialogRef<EoIdleTimerCountdownModalComponent>,
+    @Inject(WATT_DIALOG_DATA) public data: { countdown$: Observable<string> }
   ) {
-    this.countdown$ = data.countdown;
-  }
-
-  close(action?: string) {
-    this.dialogRef.close(action);
+    data.countdown$.subscribe((countdown) => {
+      if (countdown === '00:00') {
```
mimse, line 47: Suggestion: Make countdown a number?

libs/eo/shared/services/src/idle-timer/idle-timer.service.ts
```diff
@@ -52,7 +46,7 @@ export class IdleTimerService {
   private destroy$ = new Subject<void>();
   private lastActivity!: number | null;
 
-  constructor(private dialog: MatDialog, private authService: EoAuthService) {}
+  constructor(private authService: EoAuthService, private modalService: WattModalService) {}
```
mimse, line 49: Suggestion: havent we switch to using inject instead of constructor DI?

libs/eo/shared/services/src/idle-timer/idle-timer.service.ts
```diff
@@ -102,14 +96,9 @@ export class IdleTimerService {
 
     if (!lastActivity) lastActivity = new Date().getTime();
 
-    const countdown = timer(0, 1000).pipe(
+    const countdown$ = timer(0, 1000).pipe(
       takeUntil(this.destroy$),
```
mimse, line 100: Suggestion: Use takeUntilDestoryed

mimse (changes requested):

libs/eo/shared/services/src/idle-timer/idle-timer.service.ts
frontend-specialisten: It does not work like a destroy$ normally does, I've renamed the subject to avoid confusion.

mimse (approved):
💯

PR #2154, opened 2023-10-17
feat: organization audit log - drawer

michaelmajgaard:


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Resolvers/MarketParticipantResolvers.cs
```diff
@@ -57,5 +58,10 @@ public class MarketParticipantResolvers
             [Parent] OrganizationDto organization,
             ActorByOrganizationBatchDataLoader dataLoader) =>
             await dataLoader.LoadAsync(organization.OrganizationId.ToString());
+
+        public Task<GetAuditIdentityResponseDto> GetIdentityAsync(
```
mimse, line 62: Suggestion, return a string to make the schema more simple

mimse (changes requested):

mimse (approved):

PR #2174, opened 2023-10-23
Clean up

mimse:
Moved FormGroupDirective to shell

### DIFF
--- a/libs/eo/core/shell/src/lib/eo-core-shell.providers.ts
+++ b/libs/eo/core/shell/src/lib/eo-core-shell.providers.ts
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 import { importProvidersFrom } from '@angular/core';
-import { FormGroupDirective } from '@angular/forms';
 import { MatDialogModule } from '@angular/material/dialog';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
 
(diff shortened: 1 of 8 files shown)


libs/eo/core/shell/src/lib/eo-core-shell.providers.ts
```diff
@@ -25,6 +26,7 @@ import { danishDatetimeProviders } from '@energinet-datahub/watt/danish-date-tim
 import { WattModalService } from '@energinet-datahub/watt/modal';
 
 export const eoCoreShellProviders = [
+  FormGroupDirective,
```
frontend-specialisten, line 29: Question: Why is this needed?

frontend-specialisten, line 29: Question: This doesn't seem to be necessary. Do you have a specific reason for adding this? Otherwise, please remove it :)

mimse: I needed to add this for reactive form validation in watt-field to work
https://github.com/Energinet-DataHub/greenforce-frontend/blob/ae88fffd8c922243373df5b7a7bb742ec0bcd907/libs/watt/src/lib/components/field/watt-field.component.ts#L63

frontend-specialisten: But if I remove it, it still work? (in EO)

mimse: Aha, must be because you added the optional param

frontend-specialisten (approved):

PR #2192, opened 2023-10-27
Feedback from review

mimse:


libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.ts
```diff
@@ -122,10 +125,10 @@ export class DhWholesaleRequestCalculationComponent {
   private _transloco = inject(TranslocoService);
   private _toastService = inject(WattToastService);
   private _destroyRef = inject(DestroyRef);
-  private _formGroupDirective = inject(FormGroupDirective);
-  private _selectedEicFunction: EicFunction | null | undefined = null;
+  private _selectedEicFunction!: EicFunction;
```
dzhavat, line 128: suggestion: the property can be `null`/`undefined` according to the return type, so let's reflect that here as well.

libs/dh/shared/ui-util/src/lib/dh-enum-to-dropdown-options.ts
```diff
@@ -16,9 +16,24 @@
...
 import { WattDropdownOptions } from '@energinet-datahub/watt/dropdown';
 
-export function enumToDropdownOptions<T extends object>(enumObj: T) {
-  return Object.keys(enumObj).map((key) => ({
-    displayValue: key,
-    value: Object.values(enumObj)[Object.keys(enumObj).indexOf(key)],
-  })) as WattDropdownOptions;
+type Sort = 'asc' | 'desc';
+
+export function DhEnumToWattDropdownOptions<T extends object>(
```
dzhavat, line 21: nitpick: throughout the project functions start with a small letter, so let's keep that convention.

libs/dh/shared/ui-util/src/lib/dh-enum-to-dropdown-options.ts
```diff
@@ -16,9 +16,24 @@
  */
 import { WattDropdownOptions } from '@energinet-datahub/watt/dropdown';
 
-export function enumToDropdownOptions<T extends object>(enumObj: T) {
-  return Object.keys(enumObj).map((key) => ({
-    displayValue: key,
-    value: Object.values(enumObj)[Object.keys(enumObj).indexOf(key)],
-  })) as WattDropdownOptions;
+type Sort = 'asc' | 'desc';
```
dzhavat, line 19: suggestion: this type can also be inlined next to the param.

libs/dh/shared/ui-util/src/lib/dh-enum-to-dropdown-options.ts
```diff
@@ -16,9 +16,24 @@
...
+
+export function DhEnumToWattDropdownOptions<T extends object>(
+  enumObj: T,
+  exclude?: string[],
```
dzhavat, line 23: nitpick: filter the unnecessary options out before calling the function. There's no need to create options just to be discarded in the next step.

Note: this comment is more of a personal preference (thus marked as "nitpick"). The current version is fine too. Ignore the comment if you prefer the current version.

mimse: But you dont have a list before this function. You only have an enum. So it needs to be made into a list before the filter can be applied :-)

dzhavat: Oh, yeah, that's true. Let's keep it then :)

dzhavat (approved):

PR #2194, opened 2023-10-27
refactor: use self-closing tags 

dzhavat:
### Workspace
- configure self-closing tags lint rule to only target files that end with `*.component.html`. This also includes inline templates.
- run lint on the workspace
- replace deprecated `dep-graph` Nx command

Note: main `index.html` files should not use self-closing custom elements (i.e. `<dh-app />`), because, as per HTML spec, this is invalid HTML.

## References

- N/A


mimse (approved):

.eslintrc.json
```diff
@@ -205,6 +205,13 @@
       "files": ["*.ts"],
       "extends": ["plugin:@ngrx/component-store-strict"]
     },
+    {
+      "files": ["apps/**/app/**/*.html", "libs/**/*.html"],
```
frontend-specialisten, line 209: Question: what about components using inline templates?

frontend-specialisten (commented):
1. EO has now been checked for failures, and everything now works as expected.  
2. When I try breaking the rule, I don't get any feedback in my editor or by running `yarn lint`?
3. Can you please explain this: `Note: main index.html files should not use self-closing custom elements (i.e. <dh-app />) as this is invalid HTML.`

.eslintrc.json
dzhavat: It looks like the parser used in the next line also checks inline templates. :)

dzhavat: I tested a few things, and the `parser` was not really needed. The inline templates are still checked, though.

frontend-specialisten (approved):
Great work!

Issue #2073, opened 2023-09-20
Filter dropdown needs a minimum width


renehossle:
Dropdowns seem to adapt to the width of their parent.

This makes the text in the dropdown useless, and user expectations are not met, making it confusing.

Solution:
Consider having a minimum width. Talk with @renehossle about what makes sense to do here and maybe adjust the acceptance criteria. 

Accept Criteria:
- [ ] Dropdown options (overlay) should have a minimum width not restricted by the "input" itself. 
- [ ] Check-in with Khatozen in Team Mandalorian when fixed (Shout out in the frontend channel to verify the change in DataHub 3.0) 

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdUM2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--c2f8f5efffbb3f50ff6f0022c6d101a5ee781f5d/Screenshot%202023-09-20%20at%2013.38.52.png)
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdUc2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--d0e9869ff491956abc6d7361ac24b288eda85516/Screenshot%202023-09-20%20at%2013.38.58.png)

PR #2198, opened 2023-10-30
set min-width and override maxheight for dropdown

Jaypeg85:
## Description

- Set minimum width for downdown panel to 250px in case of small dropdowns
- Set max-height of dropdown to 80vh to be able to show many more items in one view

## References

- #2073


mimse:
Question: Did you try getting it to dynamically resize depending on the largest item in the list? I think this would be better than a hard width of 250px, but i do not know if it is possible.

Jaypeg85:
> Question: Did you try getting it to dynamically resize depending on the largest item in the list? I think this would be better than a hard width of 250px, but i do not know if it is possible.

I didn't try that, but I'm not sure it would be very nice, I might be wrong though. Maybe the min-width could be the size of the select?

mimse:
Guess it was not a big problem

PR #2226, opened 2023-11-09
Refactored to use takeUntilDestroyed

mimse:
Unified the way we destroy subscription when the component is destroyed

This should make it easier for newcomers to the project to know how to destroy subscriptions

### DIFF
--- a/libs/dh/charges/feature-prices/src/lib/drawer/charge-content/message-tab/dh-charges-charge-messages-tab.component.ts
+++ b/libs/dh/charges/feature-prices/src/lib/drawer/charge-content/message-tab/dh-charges-charge-messages-tab.component.ts
@@ -19,16 +19,15 @@ import {
   Component,
   Input,
   OnInit,
-  OnDestroy,
   OnChanges,
   ViewChild,
   inject,
+  DestroyRef,
 } from '@angular/core';
 import { CommonModule } from '@angular/common';
 import { TranslocoModule, TranslocoService } from '@ngneat/transloco';
 import { RxPush } from '@rx-angular/template/push';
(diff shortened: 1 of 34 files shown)


libs/dh/charges/feature-prices/src/lib/drawer/charge-content/message-tab/dh-charges-charge-messages-tab.component.ts
```diff
@@ -80,9 +80,10 @@ export class DhChargesChargeMessagesTabComponent implements OnInit, OnChanges, O
 
   @Input() charge: ChargeV1Dto | undefined;
 
-  private transloco = inject(TranslocoService);
-  private chargeMessagesStore = inject(DhChargeMessagesDataAccessApiStore);
-  private dhChargesPricesDrawerService = inject(DhChargesPricesDrawerService);
+  private _transloco = inject(TranslocoService);
+  private _destroyRef = inject(DestroyRef);
+  private _chargeMessagesStore = inject(DhChargeMessagesDataAccessApiStore);
+  private _dhChargesPricesDrawerService = inject(DhChargesPricesDrawerService);
```
dzhavat, line 86: question: any reason for adding the underscore at the beginning? Is it only to distinguish private properties in the class? Does it matter :)

mimse: It does not matter, but I think it makes it easier to identify what field is private and what is public. Also they are easier to filter out when using code completion :-)

dzhavat (approved):
praise: nice changes. But better with `DestroyRef` and `takeUntilDestroyed` 💯

PR #2228, opened 2023-11-09
not needed

mimse:
Removed the @ignore comment from the code. They had no effect or use

### DIFF
--- a/libs/watt/src/lib/components/dropdown/watt-dropdown.component.ts
+++ b/libs/watt/src/lib/components/dropdown/watt-dropdown.component.ts
@@ -16,13 +16,14 @@
  */
 import {
   Component,
+  DestroyRef,
   Host,
   HostBinding,
   Input,
-  OnDestroy,
   OnInit,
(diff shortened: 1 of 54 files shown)


libs/watt/src/lib/components/dropdown/watt-dropdown.component.ts
```diff
@@ -67,33 +67,19 @@ import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
   ],
 })
 export class WattDropdownComponent implements ControlValueAccessor, OnInit {
-  /**
-   * @ignore
-   */
   private _destroyRef = inject(DestroyRef);
```
dzhavat, line 70: suggestion: consider reverting the changes.

It looks like the comment works. Here's before and after when running Storybook locally. I haven't tested with all properties, but it seems like adding the comment removes properties from Storybook.

**Notice the "Properties" section**

Before (with comment present)

image

After (with comment removed)

image

PR #2235, opened 2023-11-14
added missing gridarea dropdown, and fix some bugs

mimse:
image


libs/dh/market-participant/actors/feature-actors/src/lib/create/dh-actors-create-actor-modal.component.ts
```diff
@@ -138,6 +140,7 @@ export class DhActorsCreateActorModalComponent {
     glnOrEicNumber: ['', [Validators.required, dhEicOrGlnValidator]],
     name: [''],
     marketrole: ['', Validators.required],
+    gridArea: ['', Validators.required],
```
dzhavat, line 143: question: this control is required but the dropdown is only shown when the selected market role is `GridAccessProvider`. Can the user still go to the next step if the selected market role is something different than `GridAccessProvider`?

PR #2240, opened 2023-11-15
use inject instead of constructor DI

mimse:
This PR updates our constructor DI´s to injects instead. So our code base is align with how we do it now.
I have not done it for injectables, because we have some unit tests that parses parameters to the constructor.
We need talk about those usecases, before we can use inject here as well.


libs/eo/shared/atomic-design/ui-atoms/src/lib/eo-idle-timer/idle-timer-countdown.component.ts
```diff
@@ -40,13 +40,12 @@ import { Observable, filter, tap } from 'rxjs';
   `,
 })
 export class EoIdleTimerCountdownModalComponent {
+  private dialogRef = inject<WattDialogRef<EoIdleTimerCountdownModalComponent>>(WattDialogRef);
+  data = inject(WATT_DIALOG_DATA);
```
dzhavat, line 44: issue: this is actually not the same as before. `data` is now type `any`, where before it was `{ countdown$: Observable<number> }`

mimse: Put a type on

dzhavat (approved):
praise: nice work 💯 
DH approve ✅

Now if there only was a lint rule to check for this ... maybe there will be one in the future 😊

PR #2249, opened 2023-11-20
EO: illustrate consumption

frontend-specialisten:
## Description

Illustrate consumption

## References

[#1930](https://github.com/Energinet-DataHub/energy-origin-issues/issues/1930)


libs/eo/shared/services/src/api-versioning/api-versioning.interceptor.ts
```diff
@@ -0,0 +1,33 @@
...
+import { EoApiEnvironment, eoApiEnvironmentToken } from '@energinet-datahub/eo/shared/environments';
+
+@Injectable()
+export class EoApiVersioningInterceptor implements HttpInterceptor {
+  constructor(@Inject(eoApiEnvironmentToken) private apiEnvironment: EoApiEnvironment) {}
+
+  intercept(request: HttpRequest<unknown>, next: HttpHandler): Observable<HttpEvent<unknown>> {
+    const apiVersions = this.apiEnvironment.apiVersions;
+    const versionedPaths = Object.keys(apiVersions);
+    const isVersioned = versionedPaths.find((path) => request.url.includes(path));
```
ckr123, line 14: This looks to me that it takjes the request url to see which sub system to use.
What if we change the urls to not include the sub systems?

ckr123 (approved):

PR #2265, opened 2023-11-22
bug: Added a day to the settlement download input parameters

energinet-ajw:
Because the end period in the Wholesale API is exclusive, so we need to add one day to the period.
Example - Before the fix.

In UI we select the period 2023-11-14 to 2023-11-15
This is converted to 2023-11-13 to 2023-11-14 before requesting the Wholesale API because of a UTC conversion and a substring operation. And this is the wrong period, that is, one day back in time.

https://app.zenhub.com/workspaces/mandalorian-5fd22a1a59e4740018f5ab5a/issues/gh/energinet-datahub/team-mandalorian/80

Reach out if you have a better solution.


libs/dh/wholesale/feature-settlement-reports/src/lib/tabs/dh-wholesale-settlements-reports-tabs-balance.component.ts
```diff
@@ -218,12 +219,15 @@ export class DhWholesaleSettlementsReportsTabsBalanceComponent
       message: this.transloco.translate('wholesale.settlementReports.downloadStart'),
     });
 
+    const startPeriod = this.tryAddOneDay(this.searchForm.controls.period.value?.start);
```
frontend-specialisten, line 222: Question: I understand what you're doing, however my question is if we're interested in calling the API if the periods  are not set? (the empty string fallback).

libs/dh/wholesale/feature-settlement-reports/src/lib/tabs/dh-wholesale-settlements-reports-tabs-balance.component.ts
```diff
@@ -237,4 +241,12 @@ export class DhWholesaleSettlementsReportsTabsBalanceComponent
           }),
       });
   }
+
+  private tryAddOneDay(value: unknown): unknown {
```
frontend-specialisten, line 245: Question: From what I can see, the value should be of type either `string | undefined`, so I can't see how your if statement is ever being executed?

libs/dh/wholesale/feature-settlement-reports/src/lib/tabs/dh-wholesale-settlements-reports-tabs-balance.component.ts
energinet-ajw: Good question. The previous code had this logic:

this.searchForm.controls.period.value?.start ?? ''

libs/dh/wholesale/feature-settlement-reports/src/lib/tabs/dh-wholesale-settlements-reports-tabs-balance.component.ts
energinet-ajw: I can't use string or undefined.

frontend-specialisten (approved):

PR #2306, opened 2023-12-04
fix(dh): reset search when filters get reset

dzhavat:
### DataHub
- reset search when filters get reset
- add fallback value to table data 
- use operator to clear subscription

## References

- https://github.com/Energinet-DataHub/team-titans/issues/367


mimse (approved):
💯

PR #2315, opened 2023-12-05
feat(dh): reset search input when filters get reset

dzhavat:
### DataHub
- reset search input when filters get reset
- remove unused import

### Watt
- do not clear input if it's already empty

## References

- N/A

### DIFF
--- a/libs/watt/src/lib/components/search/watt-search.component.ts
+++ b/libs/watt/src/lib/components/search/watt-search.component.ts
@@ -58,9 +58,7 @@ export class WattSearchComponent {
   @Output() search = new EventEmitter<string>();
 
   public clear(): void {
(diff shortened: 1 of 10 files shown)


libs/watt/src/lib/components/search/watt-search.component.ts
```diff
@@ -58,6 +58,10 @@ export class WattSearchComponent {
   @Output() search = new EventEmitter<string>();
 
   public clear(): void {
+    if (this.input.nativeElement.value === '') {
+      return;
+    }
+
```
mimse, line 64:
```suggestion
    if (this.input.nativeElement.value === '') return;

```

mimse (dismissed):

defectiveAi (approved):

PR #2334, opened 2023-12-12
refactor(dh): adjustments to admin page

dzhavat:
### DataHub
- refactor filters to use chips
- move filters inside card
- organize translations
- add empty message
- remove unnecessary styling
- simplify store selector
- fix tests

## References

- https://github.com/Energinet-DataHub/team-titans/issues/363

### DIFF
--- a/libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab-status-filter.component.ts
+++ b/libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab-status-filter.component.ts
@@ -31,10 +31,11 @@ import { DhDropdownTranslatorDirective } from '@energinet-datahub/dh/shared/ui-u
       <watt-dropdown
         dhDropdownTranslator
         translate="admin.userManagement.userStatus"
-        [label]="t('filter.status')"
+        [placeholder]="t('filter.status')"
         [formControl]="statusControl"
         [options]="userStatusOptions"
         [multiple]="true"
+        [chipMode]="true"
       />
     </ng-container>
   `,
@@ -42,10 +43,6 @@ import { DhDropdownTranslatorDirective } from '@energinet-datahub/dh/shared/ui-u
     `
       :host {
         display: block;
-      }
-
-      watt-dropdown {
-        width: 15rem;
       }
     `,
   ],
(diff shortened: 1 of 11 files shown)


libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab-status-filter.component.ts
```diff
@@ -59,28 +65,16 @@ export class DhUsersTabStatusFilterComponent implements OnInit {
...
+    (entry) => ({
+      value: entry,
+      displayValue: entry,
+    })
+  );
 
   ngOnInit(): void {
-    this.buildUserStatusOptions();
     this.statusControl.valueChanges
       .pipe(takeUntilDestroyed(this._destroyRef))
```
mimse, line 77: Suggestion: I think it is possible to move this to a constructor, then it would be possible to remove the destoryRef

libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab.component.spec.ts
```diff
@@ -120,27 +125,29 @@ describe(DhUsersTabComponent, () => {
     const { store } = await setup();
 
     const inputValue = 'test';
-    const searchInput = screen.getByRole('textbox');
+    const searchInput = screen.getByRole('searchbox');
 
     userEvent.type(searchInput, inputValue);
-    tick(searchDebounceTimeMs);
+    tick(250);
```
mimse, line 131: Question: This is now a hardcoded value, is this correct?

mimse (changes requested):

libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab.component.spec.ts
dzhavat: It is correct, but also intentional (for now). That is because `searchDebounceTimeMs` was imported from `ui-search` library, which is no longer used in the `dh-users-tab` component. The idea is to expose `searchDebounceTimeMs` from `ui-util` library in a follow up PR.

mimse (approved):

PR #2335, opened 2023-12-12
Bugfix/ qa anmod om beregninger datokomponent

mimse:
Fix a few QA comments

- https://github.com/Energinet-DataHub/team-titans/issues/341


libs/dh/wholesale/feature-request-calculation/src/lib/dh-whole-request-calculation-validators.ts
```diff
@@ -0,0 +1,82 @@
...
+export const startAndEndDateCannotBeInTheFutureValidator =
+  () =>
+  (control: AbstractControl): ValidationErrors | null => {
+    const range = control.value as WattRange<string>;
+    if (!range) return null;
+
+    const endDate = parseISO(range.end);
+    const startDate = parseISO(range.start);
+
+    if (endDate > new Date() || startDate > new Date()) {
```
dzhavat, line 77: issue: the `endDate` is always "end of (the selected) day" (one millisecond before midnight). So the `endDate > new Date()` check will almost always be `true` for today's date. 
Since today's date should also be allowed (as per requirements), the logic here will need to be adjusted.

libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.ts
```diff
@@ -138,6 +127,9 @@ export class DhWholesaleRequestCalculationComponent {
       Validators.required,
       WattRangeValidators.required(),
```
dzhavat, line 128: issue: somehow a bug sneaked in the datepicker that displays the required error message even after valid values are (manually) inputted. Doesn't look like the bug was introduced in this PR, so we should address it separately.

issue

dzhavat (approved):

libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.ts
mimse: I noticed it as well

PR #2348, opened 2023-12-18
eo-cd changes

CodingDegen:


.github/workflows/eo-cd.yml
```diff
@@ -104,14 +104,11 @@ jobs:
     needs: affected
     runs-on: ubuntu-latest
     outputs:
-      version: ${{ steps.version.outputs.semver }}
+      version: ${{ steps.version.outputs }}
```
CodeReaper, line 107: You are missing the output name here.

.github/workflows/eo-cd.yml
```diff
@@ -104,14 +104,11 @@ jobs:
...
     steps:
       - name: Get version
         id: version
-        uses: Energinet-DataHub/energy-origin/.github/actions/yaml-semver@main
-        with:
-          yaml-file: build/infrastructure/eo/configuration.yaml
-          yaml-path: version
+        uses: Energinet-DataHub/energy-origin/.github/actions/docker-image-version
```
CodeReaper, line 111: This should be constricted to main

.github/workflows/eo-cd.yml
```diff
@@ -165,19 +162,3 @@ jobs:
...
-    runs-on: ubuntu-latest
-    name: Bump configuration version
-    if: ${{ github.event_name == 'pull_request' && needs.affected.outputs.is-affected == 'true' }}
-    needs:
-      - affected
-      - update-environment
-    steps:
-      - uses: actions/checkout@v3
-
-      - uses: ./.github/actions/yaml-increment-version
```
dzhavat, line 179: suggestion: as part of this PR, can you also delete the `.github/actions/yaml-increment-version` workflow? Its only purpose is to [automatically] bump the version, which will not be needed anymore.

praise: great job getting rid of the version bump 💯

CodingDegen: Can you verify that it's not being used anywhere else than the eo? We were not sure, but will gladly remove it, if its not being used anywhere.

dzhavat: Yes, it's not used anywhere else. It can be removed.

CodingDegen: Awesome, thanks for the suggestion.

CodeReaper (approved):

PR #2350, opened 2023-12-18
fix(dh): lock selected actor to page session

dzhavat:
### DataHub
- prefer `actorId` in `sessionStorage` when reading currently selected actor
- make `StorageFake` an `abstract` class
- export `LocalStorageFake` and `SessionStorageFake`
- fix failing tests
- add actor to mock
- update translation

## References

- https://github.com/Energinet-DataHub/team-titans/issues/380


libs/dh/shared/feature-authorization/src/lib/dh-actor-storage.ts
```diff
@@ -0,0 +1,68 @@
...
+  private actorIds: string[] = [];
+
+  setUserAssociatedActors = (actorIds: string[]) => (this.actorIds = actorIds);
+
+  getSelectedActor = () => {
+    const selectedActorInLS = this._localStorage.getItem(this.selectedActorKey);
+    const selectedActorInSS = this._sessionStorage.getItem(this.selectedActorKey);
+
+    if (selectedActorInSS) {
+      return selectedActorInSS;
```
defectiveAi, line 48: Check if this is the case `this.actorIds.includes(selectedActorInSS)`, similar to how it is done for LS. If a user loses access to actor, then refreshes page, we should not re-select the removed actor.

mimse (dismissed):

defectiveAi (approved):

PR #2365, opened 2023-12-21
Refactor Application Insights initialization

mimse:
Optimize the bundle size by lazy loading the app insight library

### DIFF
--- a/libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
+++ b/libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
@@ -16,9 +16,9 @@
  */
 import { ErrorHandler, Inject, Injectable } from '@angular/core';
 import { Router } from '@angular/router';
-import { AngularPlugin } from '@microsoft/applicationinsights-angularplugin-js';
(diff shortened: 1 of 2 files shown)


libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
```diff
@@ -55,11 +42,37 @@ export class DhApplicationInsights {
...
       return;
     }
 
+    const { ApplicationInsights } = await import(
+      /* webpackChunkName: "application-insights" */
+      '@microsoft/applicationinsights-web'
+    );
+
+    const { AngularPlugin } = await import(
+      /* webpackChunkName: "application-insights-angular-plugin" */
```
dzhavat, line 56: nitpick: doesn't look like a new chunk with this name gets generated. It's probably bundled in the `application-insights` chunk 🤔

dzhavat (approved):
praise: nice solution 💯

libs/dh/shared/util-application-insights/src/lib/dh-application-insights.service.ts
mimse: I saw this as well, lets leave it :-)

PR #2413, opened 2024-01-11
readd removed class

Jaypeg85:
## Description

Fix regression from change to expandable card


libs/watt/src/lib/components/expandable-card/watt-expandable-card.component.html
```diff
@@ -14,7 +14,7 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-<mat-expansion-panel [expanded]="expanded">
+<mat-expansion-panel [expanded]="expanded" class="watt-expandable-card watt-elevation">
```
dzhavat, line 17: issue: the same CSS classes are also added in the component's class, so they'll be doubled. Consider using [NgClass](https://angular.io/api/common/NgClass) to toggle between `watt-elevation`/`watt-solid` on the `mat-expansion-panel` element, while `watt-expandable-card` is always present.

libs/watt/src/lib/components/expandable-card/watt-expandable-card.component.html
```diff
@@ -14,7 +14,7 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-<mat-expansion-panel [expanded]="expanded">
+<mat-expansion-panel [expanded]="expanded" class="watt-expandable-card">
```
dzhavat, line 17: question: doesn't this still duplicate the `watt-expandable-card` class - once here and once in `WattExpandableCardComponent` class?

suggestion: I was thinking of something like

```html
<mat-expansion-panel [expanded]="expanded" class="watt-expandable-card watt-{{ variant }}">
```

then remove the `HostBinding` from the class, and adjust the CSS file to this:

```css
.watt-expandable-card.mat-expansion-panel {
  ...
  &.watt-solid {
    border: 1px solid var(--watt-color-neutral-grey-300);
    box-shadow: none;
  }
}
```

libs/watt/src/lib/components/expandable-card/watt-expandable-card.component.ts
```diff
@@ -55,10 +53,7 @@ export class WattExpandableCardComponent {
   _content?: WattExpandableCardContentDirective;
   /** Whether the card is elevated or has solid border */
   @Input() variant: 'solid' | 'elevation' = 'elevation';
-  @HostBinding('class')
-  get cssClass() {
-    return `watt-expandable-card watt-${this.variant}`;
-  }
+  variantClass = `watt-${this.variant}`;
```
dzhavat, line 56: issue: the value of `variantClass` is not updated once the component is initialized. So the class stays `watt-elevation` even if the Input is `solid`.

Using `class="watt-expandable-card watt-{{ variant }}"` should work fine.

dzhavat (approved):

PR #2433, opened 2024-01-22
feat(watt): support toggle position on expandable card

dzhavat:
### Watt Design System
- support toggle position on expandable card - `togglePosition` Input
- use built-in `if` syntax
- format file

## References

- N/A


libs/watt/src/lib/components/expandable-card/watt-expandable-card.component.html
```diff
@@ -24,9 +28,9 @@
 
   <ng-content />
 
-  <ng-container *ngIf="_content">
+  @if (_content) {
```
mimse, line 31: ❤️

mimse (approved):

PR #2435, opened 2024-01-22
chore: upgrade Node.js to v20

dzhavat:
### Workspace
- upgrade Node.js to latest v20, since Angular v17 added support for it

## References

- N/A


mimse (approved):
🔪 edge

PR #2439, opened 2024-01-23
feat(dh): set color on negative prices and format number

dzhavat:
### DataHub
- format number with min 5 and max 6 digits after the decimal
- change color of price if negative
- rename component
- set change detection strategy

## References

- N/A


libs/dh/imbalance-prices/shell/src/lib/table-day-view/dh-price-color.directive.ts
```diff
@@ -0,0 +1,43 @@
...
+@Directive({
+  selector: '[dhPriceColor]',
+  standalone: true,
+})
+export class DhPriceColorDirective {
+  private readonly elementRef = inject(ElementRef);
+  private readonly renderer = inject(Renderer2);
+  private readonly wattColorService = inject(WattColorHelperService);
+
+  @Input() set dhPriceColor(value: number | undefined) {
```
mimse, line 30: Suggestion: Would it not be better to use input signal or a transform function?

mimse: Of cause you do mutate something, so maybe not

mimse (approved):

libs/dh/imbalance-prices/shell/src/lib/table-day-view/dh-price-color.directive.ts
dzhavat: I was thinking of using a Signal input but the value comes from an API, and it's not wrapped in a Signal. So to wrap it in a Signal just to send it in ... maybe ok, but probably unnecessary.

After opening the PR, I thought of another solution - toggling a class in the `dh-table-day-view` component instead of creating a directive. This seems simpler, so I pushed a new commit.

libs/dh/imbalance-prices/shell/src/lib/table-day-view/dh-table-day-view.component.ts
```diff
@@ -43,11 +52,13 @@ import { DhPriceColorDirective } from './dh-price-color.directive';
       </ng-container>
 
       <ng-container *wattTableCell="columns['price']; let entry">
-        <span [dhPriceColor]="entry.price">{{ entry.price | number: '1.5-6' }}</span>
+        <span [ngClass]="{ 'negative-price': entry.price < 0 }">
```
mimse, line 55: Nice

mimse (approved):

PR #162, opened 2021-11-22
Search metering point

frontend-specialisten:
## Description

Add search for metering point

## References

[#135 ](https://github.com/Energinet-DataHub/greenforce-frontend/issues/135)

PR #2447, opened 2024-01-24
Update styles and components

frontend-specialisten:
## Description

- Bring back range separator for the legacy version of the datepicker
- [#168](https://github.com/Energinet-DataHub/team-mandalorian/issues/168)
- [#162](https://github.com/Energinet-DataHub/team-mandalorian/issues/162)


dzhavat:
issue: looks like Vater will need to be tweaked a bit. The page looks like this locally 

image

libs/dh/wholesale/feature-settlement-reports/src/lib/tabs/dh-wholesale-settlements-reports-tabs-balance.component.html
dzhavat, line 1: issue: looks like Vater will need to be tweaked a bit. The page looks like this locally 

image

frontend-specialisten: Should be fixed now, the view is now using the data table component instead of custom styling.

dzhavat (approved):

PR #2487, opened 2024-02-02
fix(dh): show descriptive error when domain is taken

dzhavat:
### DataHub
- show descriptive error when domain is taken when editing an organization 
- add update actor mock 
- update phone number in mock
- set `loading` on modal

## References

- N/A


mimse (approved):
💯

PR #2514, opened 2024-02-12
Watt as NPM package

frontend-specialisten:
## Description

Add `ng-packagr` to make distribution of `Watt` as an NPM package possible. 

## References

- #0000


workspace-tools/src/executors/bump-and-publish/executor.ts
```diff
@@ -0,0 +1,143 @@
...
+
+function generatePakacgeVersion(): string {
+  let branchName = execSync('git rev-parse --abbrev-ref HEAD').toString().trim();
+  const gitCommitHash = execSync('git rev-parse --short HEAD').toString().trim();
+
+  // Replace special characters in the branch name with a hyphen
+  branchName = branchName.replace(/[^a-zA-Z0-9]/g, '-');
+
+  let versionSuffix = branchName === 'main' ? gitCommitHash : `${branchName}.${Date.now()}`;
+  const version = branchName === 'main' ? '1.0.0' : '0.0.0';
```
ManBearTM, line 59: praise: good solution, no semver please 😄

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "energinet-datahub",
-  "version": "0.0.0",
+  "version": "1.0.0-chore-watt-npm-package.1710839337619",
```
ManBearTM, line 3: question: is this just out of sync, shouldn't the executor result in version `0.0.0-chore-watt-npm-package.1710839337619`?

ManBearTM (commented):
Awesome job! There are SO many changes, so I won't guarantee that I didn't miss something 😄 But overall it looks good. Only thing I'm missing is a little documentation on how the new Watt packages works, specifically regarding the new `ng-package.json`. They look to be identical, so perhaps it is enough to just make sure the `watt-component` generator includes this file as well.

libs/watt/src/lib/utils/date/dayjs.ts
```diff
@@ -14,13 +14,56 @@
...
 
 dayjs.extend(utc);
 dayjs.extend(timezone);
 dayjs.extend(customParseFormat);
 
-export { dayjs };
+@Injectable({
+  providedIn: 'root',
+})
+class WattDateUtils {
```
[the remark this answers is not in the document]
frontend-specialisten: image

PR #2542, opened 2024-02-20
feat(dh): refresh users in Admin when user profile is updated

dzhavat:
### DataHub
- refresh users in Admin when user profile is updated
- update translation

## References

- N/A

### DIFF
--- a/libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab.component.ts
+++ b/libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab.component.ts
@@ -47,6 +47,7 @@ import {
 } from '@energinet-datahub/watt/vater';
 import { WattDropdownOptions } from '@energinet-datahub/watt/dropdown';
 import { WattSearchComponent } from '@energinet-datahub/watt/search';
+import { DhProfileModalService } from '@energinet-datahub/dh/profile/feature-profile-modal';
 
 import { DhUsersTabTableComponent } from './dh-users-tab-table.component';
 import { DhUsersTabStatusFilterComponent } from './dh-users-tab-status-filter.component';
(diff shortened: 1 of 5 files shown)


libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab.component.ts
```diff
@@ -136,6 +138,10 @@ export class DhUsersTabComponent {
     this.userRolesStore.getRoles();
 
     this.onSearchInput();
+
+    this.profileModalService.onProfileUpdate$
+      .pipe(takeUntilDestroyed(this.destroyRef))
```
mimse, line 143: Suggestion: No need for this.destoryRef in constructor

libs/dh/profile/feature-profile-modal/src/lib/dh-profile-modal/dh-profile-modal.service.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class DhProfileModalService {
+  private profileUpdateSubject = new Subject<void>();
+
+  public onProfileUpdate$ = this.profileUpdateSubject.asObservable();
```
mimse, line 26: Question: Isnt a subject already a observable?

mimse (changes requested):

libs/dh/profile/feature-profile-modal/src/lib/dh-profile-modal/dh-profile-modal.service.ts
dzhavat: Yes, it is. The `asObservable()` pattern is for protecting the subject, so its `next` is only called via the method below.

libs/dh/admin/feature-user-management/src/lib/tabs/users-tab/dh-users-tab.component.ts
dzhavat: Nice catch. Should be fixed now.

mimse (approved):

Issue #2581, opened 2024-02-27
Send diagnostics data for GraphQL to Application Insights


ManBearTM:
Following a hard to debug incident in the GraphQL API, it became apparent that there was a lack of logging from the GraphQL API. This feature aims to address that by sending telemetry data to Application Insights.

PR #2584, opened 2024-02-28
feat(dh): OpenTelemetry

ManBearTM:
## Description

This PR adds support for OpenTelemetry in the BFF, sending diagnostics data to ApplicationInsights for both REST and GraphQL endpoints.

I also added a test to make sure the telemetry data actually reaches ApplicationInsights. This in turn brought me down the rabbit hole of refactoring and weird workflow issues. But the results are good I think:

- **Much simplified test setup**
  Test classes now only need to inherit from a single base class and the flow is much easier to follow IMO.
- **Switched from CookieCrumble to Verify for snapshot testing**
  The CI is set up to build the solution in Linux, then copy the DLL's to windows and run the tests there. This caused the weirdest issue with CookieCrumble, which weren't able to find the snapshot files. The issue was fixed by switching library,
which on the upside provides a better DX IMO.
- **API tests no longer runs in frontend-ci**
  By resolving the above issue, we no longer need to run api-dh:test in frontend-ci, speeding up the workflow significantly.
- **Added API helper scripts in package.json**
  There are now some new scripts for testing the API locally. For testing use `yarn api:test`, and for helping resolve snapshot mismatches, check out `yarn api:verify`! You can also quickly serve the API by using `yarn api:dev`.
- **Telemetry tests are separated**
  Since the Application Insights can take up to 20 minutes (usually around 3-5 minutes), it was changed to run in parallel with other tests to minimize the wait. Unfortunately it seems like it is still the last check to complete...
- **Renamed appsettings.Development.json to appsettings.json**
  The appsettings needed to apply during tests as well (which wasn't considered Development). Renaming to appsettings.json makes it apply to all environments. Get the latest config from dev-secrets today!

## References

- #2581


dstenroejl (commented):
Mentioned a few suggestions

Renetnielsen (approved):

mimse (approved):
Good work 💯

PR #2597, opened 2024-03-04
Switch from date-fnz to dayjs

mimse:
Removed date-fns - date-fns-tz from dh-app, and replaced it with https://day.js.org

date--fns has been updated to version 3 which is a major version bump. But date-fns-tz havent seen develoment in over a year.

So by switching to dayjs we fix our problem with setting a global locale, avoid upgrade to version 3 of date-fns and we do not use a package which seems not to be updated anymore.

https://github.com/marnusw/date-fns-tz/issues/266


libs/watt/src/lib/configuration/watt-date-adapter.ts
```diff
@@ -21,7 +21,7 @@ const danishLocale = 'da';
 
 export class WattDateAdapter extends NativeDateAdapter {
   setActiveLocale(language: WattSupportedLocales): void {
-    this.setLocale(language === danishLocale ? 'da-DK' : 'en-GB');
+    this.setLocale(language === danishLocale ? danishLocale : 'en-GB');
```
mimse, line 24: Question: Should en-GB then be en? or should danishLocale const be da-DK? seems od to mix them?

libs/dh/shared/feature-flags/src/lib/feature-flags.spec.ts
```diff
@@ -14,15 +14,15 @@
...
+import { dayjs } from '@energinet-datahub/watt/date';
 import { DhFeatureFlag, DhFeatureFlags, dhFeatureFlagsConfig } from './feature-flags';
 
 const maxAgeOfDays = 62;
 
 const featureFlagCases = Object.keys(dhFeatureFlagsConfig).map((featureFlag) => {
   const created = (dhFeatureFlagsConfig[featureFlag as DhFeatureFlags] as DhFeatureFlag).created;
-  const parsedDate = parse(created, 'dd-MM-yyyy', new Date());
-  const diffInDays = differenceInDays(new Date(), parsedDate);
+  const parsedDate = dayjs(created, 'dd-MM-yyyy');
```
dzhavat, line 24: issue: format pattern is not valid.


```suggestion
  const parsedDate = dayjs(created, 'DD-MM-YYYY');
```

libs/watt/src/lib/components/picker/datepicker/watt-datepicker.component.ts
```diff
@@ -318,7 +318,7 @@ export class WattDatepickerComponent extends WattPickerBase {
         }
         const parsedStartDate = this.parseDateShortFormat(start);
 
-        if (isValid(parsedStartDate)) {
+        if (dayjs(parsedStartDate).isValid()) {
```
dzhavat, line 321: nitpick: update comment about `dateShortFormat` on line 312.

libs/dh/imbalance-prices/shell/src/lib/drawer/dh-drawer.component.ts
```diff
@@ -117,12 +117,27 @@ export class DhImbalancePricesDrawerComponent {
...
+
+  private readonly month = computed(() => {
+    const imbalancePrice = this.imbalancePrice();
+
+    if (!imbalancePrice) return 0;
+
+    const formatter = new Intl.DateTimeFormat(this.locale, { month: 'numeric' });
+    const formattedDate = formatter.format(imbalancePrice.name);
+    return Number(formattedDate);
+  });
```
dzhavat, line 140: suggestion: let's reuse `dayjs` to format dates. It's been used everywhere else anyway.

libs/dh/imbalance-prices/shell/src/lib/drawer/dh-drawer.component.ts
```diff
@@ -162,9 +177,12 @@ export class DhImbalancePricesDrawerComponent {
...
     });
 
+    const formatter = new Intl.DateTimeFormat(this.locale, { month: 'long', year: 'numeric' });
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const formattedDate = formatter.format(this.imbalancePrice()!.name);
+
     const fileOptions = {
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      name: 'imbalance-prices-' + format(this.imbalancePrice()!.name, 'MMMM-yyyy'),
+      name: 'imbalance-prices-' + formattedDate,
```
dzhavat, line 185: suggestion: let's reuse `dayjs` to format the date. It's been used everywhere else anyway.

mimse:
@dzhavat nice findings, should be fix / changed now :-)

dzhavat (approved):
praise: nice work. 💯

Approve from my side. Not sure if someone else should look at the PR too 👀

libs/eo/core/shell/src/lib/eo-shell.component.ts
```diff
@@ -34,7 +33,6 @@ import { EoPrimaryNavigationComponent } from './eo-primary-navigation.component'
   imports: [
     EoFooterComponent,
     EoPrimaryNavigationComponent,
-    MatDateFnsModule,
```
frontend-specialisten, line 37: Have you tested this doesn't break any functionality in `eo`? Otherwise, please revert the change, and I will prioritize moving `eo` to `dayjs` also :)

libs/watt/src/lib/configuration/watt-date-adapter.ts
```diff
@@ -14,22 +14,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Inject, Optional } from '@angular/core';
-import { DateFnsAdapter } from '@angular/material-date-fns-adapter';
-import { MAT_DATE_LOCALE } from '@angular/material/core';
-import { da as daLocale, enGB as enGBLocale } from 'date-fns/locale';
+import { NativeDateAdapter } from '@angular/material/core';
 
 export type WattSupportedLocales = 'da' | 'en';
```
frontend-specialisten, line 19: Question: wouldn't it be better to actually use the real locale name `en-GB`? I know that we currently only support Danish and English but because we're doing it like this we need checks like `language === danishLocale ? danishLocale`

libs/watt/src/lib/configuration/watt-date-adapter.ts
```diff
@@ -39,8 +31,16 @@ export class WattDateAdapter extends DateFnsAdapter {
   override getDateNames(): string[] {
     const dateNames = super.getDateNames();
 
-    return this.locale.code === danishLocale
+    return this.locale === danishLocale
```
frontend-specialisten, line 34: Consideration: Couldn't we just make the replacement regardless of the locale? I know from a micro-optimization perspective it doesn't make sense, but it would remove the hard binding of doing something specific on a specific locale.

libs/watt/src/lib/configuration/watt-locale.service.ts
```diff
@@ -29,6 +30,15 @@ export class WattLocaleService {
   constructor(private dateAdapter: DateAdapter<unknown>) {}
 
   setActiveLocale(locale: WattSupportedLocales): void {
+    if (locale === 'da') {
+      import('dayjs/locale/da');
+    }
+
+    if (locale === 'en') {
+      import('dayjs/locale/en');
+    }
```
frontend-specialisten, line 39: Question: Why not make this dynamic? We already have type-safety for the locale since it should be a supported locale. So why not make it easier to add other locales and remove some if statements? :)
OR does AST come into play here?
```suggestion
     import(`dayjs/locale/${locale}`);
```

package.json
```diff
@@ -89,6 +88,7 @@
     "chart.js": "4.4.0",
     "date-fns": "2.29.2",
     "date-fns-tz": "2.0.0",
+    "dayjs": "^1.11.10",
```
frontend-specialisten, line 91:
```suggestion
    "dayjs": "1.11.10",
```

libs/eo/core/shell/src/lib/eo-shell.component.ts
mimse: It will not work with MatDateFnsModule, because angular materiale now uses the native provider and not the date-fns one.
I did launch EO and it looked fine to me, but it might be a good idea with second peer of eyes :-)

libs/watt/src/lib/configuration/watt-date-adapter.ts
mimse: @dzhavat did a the research, it should be okay :-)

libs/watt/src/lib/configuration/watt-locale.service.ts
mimse: I can try it out, but I do not thing it will work. It would have a hard time figuring out what dynamic imports to create when it cannot static analyze the code?

libs/watt/src/lib/configuration/watt-date-adapter.ts
mimse: It did not want to change to much logic around our locales, but only switch date-fns for dayjs. Other refactoring could be done in a other PR :-)
But you do have a good point.

libs/eo/core/shell/src/lib/eo-shell.component.ts
frontend-specialisten: It looks fine to me :)

libs/watt/src/lib/configuration/watt-date-adapter.ts
frontend-specialisten: Fine by me, I create an issue at the bounty hunters board for a follow-up refactoring task :)

libs/watt/src/lib/configuration/watt-date-adapter.ts
frontend-specialisten: We can talk about this in the follow-up refactoring issue. I resolve for now :)

dzhavat: @frontend-specialisten you can see some details about this in the first resolved comment in this PR. :)

libs/watt/src/lib/configuration/watt-locale.service.ts
mimse: image

As I suspected

dzhavat (approved):

frontend-specialisten (approved):

PR #2612, opened 2024-03-06
feat(dh): add permissions to Delegation tab

dzhavat:
## References

- N/A


mimse (approved):
Nikpick: I properly would have used featureflag service and a if instead.

PR #2624, opened 2024-03-07
refac: Removed not used settlement report endpoints.

energinet-ajw:
Removed 2 not used endpoints from BFF (and Wholesale API).
Updated WholesaleClient too.

Changed get settlement report test so it tests download endpoint instead.

(https://app.zenhub.com/workspaces/mandalorian-5fd22a1a59e4740018f5ab5a/issues/gh/energinet-datahub/opengeh-wholesale/1619)


apps/dh/api-dh/source/DataHub.WebApi.Tests/Integration/Controllers/WholesaleSettlementReportControllerTests.cs
```diff
@@ -37,17 +39,30 @@ public sealed class WholesaleSettlementReportControllerTests : ControllerTestsBa
         }
 
         [Fact]
-        public async Task CreateAsync_ReturnsOk()
+        public async Task DownloadAsync_ReturnsOk()
```
ManBearTM, line 42: praise: you could have taken the easy way out here, but you actually added a missing test instead. nice work!

energinet-ajw: :)

PR #2643, opened 2024-03-12
feat(dh): filter delegations

dzhavat:
## References

- N/A


libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -186,6 +186,7 @@
         "RSM017_OUTBOUND": "Anmod om engrosydelser (RSM-017)",
         "RSM019_INBOUND": "Fremsend beregnede engrosydelser (RSM-019)"
       },
+      "statusPlaceholder": "Status",
```
mimse, line 189: Suggestion: Looks like you are missing some translations
image

mimse (changes requested):
Suggestion: If possible keep the expanded state of the accordion when applying the filter. Its hard to see what you filter on when the result is hidden

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
dzhavat: Good catch. Was missing an import.

mimse (approved):

PR #2646, opened 2024-03-13
feat(dh): grid area displayName

ManBearTM:
## Description

This PR makes sure grid areas are displayed uniformly across DataHub by moving the logic to GraphQL. A new field has been added to gridArea called displayName, which combines code and name in a consistent way.


dzhavat (dismissed):

mimse (dismissed):
Praise: Nice change :-)

PR #2651, opened 2024-03-14
Bugfix/datepicker

mimse:
Switch to new pickers

**Important**: The new datepicker is based on Maskito. This means that inputting values programmatically is now different. Before, values were inputted in "from input" and "to input" (in range picker). What I have found in the new component is that values should be inputted in `input.mask-input` element (for both single and range mode). Furthermore, dates should be without dash, e.g. "14032024", whereas time should be without column, e.g "1530".


frontend-specialisten (changes requested):
Due to my test on `eo`, I found an issue when editing a transfer agreement, which needs to be fixed before merging this PR. 

image

mimse:
> Due to my test on `eo`, I found an issue when editing a transfer agreement, which needs to be fixed before merging this PR.
> 
> image

Hvordan finder jeg edit skærm billedet

frontend-specialisten (approved):

PR #2679, opened 2024-03-19
hack(dh): filter grid areas in test 001

ManBearTM:
## Description

If a user starts a calculation with a grid area that has no data (which is most of them), the calculation will fail. This creates a lot of noise for the wholesale operation. This is a temporary hack to make sure that only a specific subset of grid areas can be selected when starting new calculations. I made this as a feature flag so it could only apply to test-001 AND to periodically get the reminder that this hack exists.

### DIFF
--- a/libs/dh/shared/feature-flags/src/lib/feature-flags.ts
+++ b/libs/dh/shared/feature-flags/src/lib/feature-flags.ts
@@ -41,6 +41,10 @@ export const dhFeatureFlagsConfig = {
       DhAppEnvironment.test_002,
     ],
   },
+  'calculations-include-all-grid-areas': {
+    created: '18-03-2024',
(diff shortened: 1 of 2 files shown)


libs/dh/shared/feature-flags/src/lib/feature-flags.ts
```diff
@@ -41,6 +41,10 @@ export const dhFeatureFlagsConfig = {
       DhAppEnvironment.test_002,
     ],
   },
+  'calculations-include-all-grid-areas': {
+    created: '18-03-2024',
+    disabledEnvironments: [DhAppEnvironment.test_001],
```
dzhavat, line 46: issue: the PR description says "a feature flag so it could only apply to test-001" but doesn't the code above do the opposite - it disables the feature in `test_001` and enables it everywhere else 😊

PR #2682, opened 2024-03-19
Fixed date formtting

mimse:


libs/watt/src/lib/utils/date/watt-date.pipe.ts
```diff
@@ -42,10 +42,16 @@ export class WattDatePipe implements PipeTransform {
   ): string | null {
     if (!input) return null;
 
-    return input instanceof Date || typeof input === 'string'
-      ? dayjs(input).tz(timeZone).format(formatStrings[format])
-      : typeof input === 'number'
-        ? dayjs(new Date(input)).tz(timeZone).format(formatStrings[format])
-        : `${this.transform(input.start, format)} ― ${this.transform(input.end, format)}`;
+    if (input instanceof Date || typeof input === 'string') {
+      return dayjs(input).year() === 10000 // Max Date is the same as null so dont show it
```
dzhavat, line 46: nitpick: slighly changed wording hopefully making it more clear 


```suggestion
      // Treat year 1000 as `null` in the UI
      // Needed because in some cases `input.end` is set to a value that is far into the future
      // in order to signify that the value is not defined  
      return dayjs(input).year() === 10000 
```

libs/watt/src/lib/utils/date/watt-date.pipe.ts
```diff
@@ -42,10 +42,16 @@ export class WattDatePipe implements PipeTransform {
...
+        ? null
```
dzhavat, line 47: suggestion: add a test that covers this case

libs/watt/src/lib/utils/date/watt-date.pipe.spec.ts
```diff
@@ -44,6 +44,11 @@ describe(WattDatePipe, () => {
     expect(pipe.transform(range)).toBe('25-03-2019 ― 27-03-2019');
   });
 
+  it('transforms date range with end null', () => {
+    const range = { start: '2019-03-25T22:00:00Z', end: '9999-12-31T23:59:59Z' };
+    expect(pipe.transform(range)).toBe('25-03-2019');
+  });
+
```
dzhavat, line 51: suggestion: noticed that we're missing a test for the same day

```suggestion
  it('transforms date range with end null', () => {
    const range = { start: '2019-03-25T22:00:00Z', end: '9999-12-31T23:59:59Z' };
    expect(pipe.transform(range)).toBe('25-03-2019');
  });

  it('transforms date range with start and end dates falling on the same day', () => {
    const range = { start: '2019-03-25T22:00:00Z', end: '2019-03-25T22:59:59Z' };
    expect(pipe.transform(range)).toBe('25-03-2019');
  });
```

mimse: Good catch

dzhavat (approved):

PR #2692, opened 2024-03-21
fix(dh): set default encoding for csharp files

ManBearTM:
## Description

Small change to always set encoding to UTF8 with BOM when writing C#, since the compiler otherwise throws an error.


dzhavat (approved):

mimse:
OMG YES Champ !

PR #2711, opened 2024-03-25
fix(dh): set button in loading state while a request is active

dzhavat:
### DataHub
- set button in loading state while a request is active
- update mock

### Watt
- extend step component to allow setting next button in loading state

## References

- N/A


mimse (approved):
Great addition

PR #2737, opened 2024-04-05
Set the correct flag when init with a value

mimse:


ManBearTM (approved):
Would it make sense to add a test for this functionality?

mimse:
> Would it make sense to add a test for this functionality?

It would, I will make a task for it in the bounty board

libs/watt/src/lib/components/phone-field/watt-phone-field.component.ts
dzhavat, line 1: praise: the flag appears correctly now. Good job 💯

issue: when opening the profile modal for a second time, a console error is thrown about `metadata` argument.

bug

Loading the  `metadata` eagerly seems to fix the issue. Does loading the `metadata` eagerly affect the final bundle a lot?

dzhavat, line 1: issue:: the phone is not formatted correctly when opening the modal initially. Eagerly loading the  `metadata` seems to fix that (on opening the profile modal for the second time).

bug2

PR #2738, opened 2024-04-08
feat(watt): add support for sorting options in dropdown

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/shared/ui-util/src/lib/dh-dropdown-translator.directive.ts
+++ b/libs/dh/shared/ui-util/src/lib/dh-dropdown-translator.directive.ts
@@ -16,31 +16,23 @@
  */
 import { DestroyRef, Directive, Input, OnInit, inject } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
-import { WattDropdownComponent } from '@energinet-datahub/watt/dropdown';
 import { TranslocoService } from '@ngneat/transloco';
-import { withLatestFrom } from 'rxjs';
+
+import { WattDropdownComponent } from '@energinet-datahub/watt/dropdown';
 
 @Directive({
   standalone: true,
   selector: '[dhDropdownTranslator]',
 })
 export class DhDropdownTranslatorDirective implements OnInit {
-  private trans = inject(TranslocoService);
+  private translocoService = inject(TranslocoService);
   private host = inject(WattDropdownComponent);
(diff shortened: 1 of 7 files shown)


libs/dh/shared/ui-util/src/lib/dh-dropdown-translator.directive.ts
```diff
@@ -16,31 +16,23 @@
...
+
+import { WattDropdownComponent } from '@energinet-datahub/watt/dropdown';
 
 @Directive({
   standalone: true,
   selector: '[dhDropdownTranslator]',
 })
 export class DhDropdownTranslatorDirective implements OnInit {
-  private trans = inject(TranslocoService);
+  private translocoService = inject(TranslocoService);
```
mimse, line 28: Praise: better name

libs/dh/shared/ui-util/src/lib/dh-dropdown-translator.directive.ts
```diff
@@ -16,31 +16,23 @@
...
-      .pipe(
-        takeUntilDestroyed(this.destroyRef),
-        withLatestFrom(this.trans.selectTranslateObject<object>(this.translate))
-      )
-      .subscribe(([, keys]) => {
-        this.setTranslation(keys);
-      });
-
-    this.trans
+    this.translocoService
```
mimse, line 35: Suggestion: I think this could be moved to the constructor. Then we wont not need the destroyRef

libs/watt/src/lib/components/dropdown/watt-dropdown.module.spec.ts
```diff
@@ -457,5 +463,46 @@ describe(WattDropdownComponent, () => {
         expect(actualLabel).toBe(noOptionsFoundLabel);
       });
     });
+
+    describe('sorting', () => {
```
mimse, line 467: Praise: Nice what you added tests for this scenario as well

mimse (changes requested):

libs/dh/shared/ui-util/src/lib/dh-dropdown-translator.directive.ts
dzhavat: Unfortunately the value of `translate` Input will not be set at this point.

mimse (approved):

PR #2740, opened 2024-04-08
feat: Add GLN filter param to graphQL

FirestarJes:


schema.graphql
```diff
@@ -282,7 +282,6 @@ type Mutation {
 
 type Subscription {
   calculationCreated: Calculation!
-  calculationProgress: Calculation!
```
dzhavat, line 285: question: should this be removed? Doesn't seem related to the other changes.

FirestarJes: Ehh, i have no idea where that comes from...

dzhavat (approved):

PR #2742, opened 2024-04-09
Better mitId ux

mimse:


libs/dh/shared/feature-authorization/src/lib/dh-actor-token.service.ts
```diff
@@ -71,9 +73,19 @@ export class DhActorTokenService {
     return this.marketParticipantUserHttp.v1MarketParticipantUserGetUserActorsGet().pipe(
       tap(({ actorIds }) => this.actorStorage.setUserAssociatedActors(actorIds)),
       switchMap(() =>
-        this.tokenHttp
-          .v1TokenPost(this.actorStorage.getSelectedActor())
-          .pipe(map(({ token }) => token))
+        this.tokenHttp.v1TokenPost(this.actorStorage.getSelectedActor()).pipe(
+          map(({ token }) => token),
+          tap(async () => {
```
dzhavat, line 78: question: does it have to be `async`?

mimse: Nope, good catch, was a left over, from testing

dzhavat (approved):

PR #2745, opened 2024-04-09
feat(dh): add energy supplier filter to eSett

dzhavat:
## References

- #0000


libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -611,7 +611,8 @@
       "calculationType": "Beregningstype",
       "messageType": "Meddelelsestype",
       "gridArea": "Netområde",
-      "status": "Status"
+      "status": "Status",
+      "energySupplier": "Elleverandør"
```
mimse, line 615: Question: not possible to use shared here as well?

mimse (approved):

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
dzhavat: This is the topmost `shared` in `eSett`. Couldn't find other `shared` higher up that contain `Elleverandør`. Did you mean to add it?

PR #2754, opened 2024-04-11
fix: resubscribe to graphql subscriptions on token refresh

ManBearTM:
## Description

Currently, when a subscription runs for longer than the token expiration time, the GraphQL server will start to operate on an expired token, causing requests to fail in the backend. This PR makes sure that subscriptions are disposed and then resubscribed with the new token, once MSAL deems it necessary.


mimse (approved):
Nice rxjs magic :-)

PR #2765, opened 2024-04-12
feat: default apollo options

ManBearTM:
## Description

This PR makes the most commonly used query options the default in Apollo, avoiding a fair amount of config code. I am not sure if we even need the "notifyOnNetworkStatusChange: true", but I'm not touching that in this PR. At least now it is easy to remove.


mimse (approved):
Praise: Nice change 💯

dzhavat (approved):

PR #2768, opened 2024-04-15
feat: balance responsible ui - bff swagger and graph gen

michaelmajgaard:


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Mutation.cs
```diff
@@ -19,6 +19,7 @@
 using HotChocolate.Subscriptions;
 using NodaTime;
 using EdiB2CWebAppProcessType = Energinet.DataHub.Edi.B2CWebApp.Clients.v1.ProcessType;
+using MeteringPointType = Energinet.DataHub.Edi.B2CWebApp.Clients.v1.MeteringPointType;
```
mimse, line 22: Question: Why did you add this?

mimse (approved):

PR #2778, opened 2024-04-16
refactor(watt): use new built-in control flow

dzhavat:
## References

- #0000


libs/watt/src/lib/components/data/watt-data-table.stories.ts
```diff
@@ -58,7 +60,7 @@ export const DataTable: StoryObj<WattDataTableComponent> = {
   render: (args) => ({
     props: { ...args, ...tableStoryArgs },
     template: `
-      <watt-data-table vater inset="m" [dataSource]="dataSource">
+      <watt-data-table vater inset="m">
```
mimse, line 63: Question: Is this right?

mimse (approved):
Praise: Nice Work!

libs/watt/src/lib/components/data/watt-data-table.stories.ts
dzhavat: Yes. The `watt-data-table` doesn't have `dataSource` Input.

PR #2802, opened 2024-04-22
feat: Add balance responsible and energy supplier name

FirestarJes:


apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
```diff
@@ -69,10 +74,10 @@ type AssociatedActors {
 
 type BalanceResponsibilityAgreement {
   meteringPointType: String
+  energysupplierName: ActorNameWithId
+  balanceresponsibleId: ActorNameWithId
```
dzhavat, line 78: suggestion: can we call them `energySupplierWithName` and `balanceResponsibleWithName` instead? There's already something similar under `BalanceResponsibleType`

dzhavat (approved):

PR #2830, opened 2024-04-29
Download

mimse:


libs/dh/market-participant/actors/feature-actors/src/lib/drawer/balance-responsible-relation-tab/dh-balance-responsible-relation-tab.component.ts
```diff
@@ -246,4 +251,41 @@ export class DhBalanceResponsibleRelationTabComponent {
...
+    const lines = balanceResponsibleRelations.map((balanceResponsibleRelation) => [
+      `"${balanceResponsibleRelation.balanceResponsibleWithName?.id}"`,
+      `"${balanceResponsibleRelation.balanceResponsibleWithName?.actorName.value}"`,
+      `"${balanceResponsibleRelation.energySupplierWithName?.id}"`,
+      `"${balanceResponsibleRelation.energySupplierWithName?.actorName.value}"`,
+      `"${balanceResponsibleRelation.gridAreaId}"`,
+      `"${balanceResponsibleRelation.meteringPointType}"`,
+      `"${balanceResponsibleRelation.status}"`,
+      `"${balanceResponsibleRelation.validPeriod.start}"`,
+      `"${balanceResponsibleRelation.validPeriod.end}"`,
```
dzhavat, line 286: suggestion: some of these properties can be `undefined`/`null`. In such case, what will the output in the CSV file be? `"undefined"`/`"null"`? If that's the case, let's set it to en empty string

dzhavat (approved):

PR #2835, opened 2024-04-30
refac: Use Orchestrations app for starting calculations

dstenroejl:
## Description

StartCalculation in Wholesale has been moved to a Azure Function HttpTrigger, so the endpoint has changed.
The request DTO with the calculation info has also changed.

So with this PR we:
- [x] Create another client to call the HttpTrigger
- [x] Create another DTO
- [x] Use the new client and DTO when starting a calculation
- [x] Add a health check for the Function App

Part of:
https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/182


ManBearTM:
Looks good to me! It would be great if you could also add the url of the new endpoint to the dev-secrets [appsettings.json](https://github.com/Energinet-DataHub/dh3-dev-secrets/blob/main/greenforce-frontend/apps/dh/api-dh/source/DataHub.WebApi/appsettings.json) file (it can come after merge, it just helps us when working with the BFF locally).

ManBearTM (approved):

PR #2839, opened 2024-05-01
feat(dh): load settlement reports and show empty state

dzhavat:
## References

- #0000


libs/dh/wholesale/feature-settlement-reports-v2/src/lib/dh-wholesale-feature-settlement-reports-v2.component.ts
```diff
@@ -14,11 +14,94 @@
...
+  isLoading = false;
+
+  constructor() {
+    this.fetchData();
+  }
+
+  private fetchData(): void {
+    this.isLoading = true;
+
+    this.settlementReports$.pipe(takeUntilDestroyed(this.destroyRef)).subscribe({
```
mimse, line 96: Suggestion: The destoryRef can be removed, because its called from the constructor.

mimse (changes requested):

libs/dh/wholesale/feature-settlement-reports-v2/src/lib/dh-wholesale-feature-settlement-reports-v2.component.ts
dzhavat: Even better 🙂 I always forget it 😊

mimse (approved):

PR #2843, opened 2024-05-01
Feature/469 error and empty state handling

mimse:


dzhavat (approved):
praise: good job trying the `signalStore`. It looks quite nice. We should definitely consider using it more often in the future 💯

PR #2858, opened 2024-05-06
feat: use min and max in datepicker mask

ManBearTM:
This PR includes the min and max date configuration in the Maskito options. This means we can skip validating min and max in form controls, since maskito will prevent entering dates outside of this range.


dzhavat (approved):
praise: 💯💯

PR #2860, opened 2024-05-06
Feature/604 - show actor name column if fas

mimse:


dzhavat (approved):
praise: nice changes. We'll definitely be using the `isFas` method in more places in the future 💯

PR #2862, opened 2024-05-07
Added status indication to table overview

mimse:


libs/dh/wholesale/feature-settlement-reports-v2/src/lib/table/dh-settlement-reports-status.component.ts
```diff
@@ -0,0 +1,42 @@
...
+    @switch (status()) {
+      @case ('IN_PROGRESS') {
+        <watt-badge type="info">{{ t(status()) }}</watt-badge>
+      }
+      @case ('ERROR') {
+        <watt-badge type="warning">{{ t(status()) }}</watt-badge>
+      }
+      @case ('COMPLETED') {
+        <watt-badge type="success">{{ t(status()) }}</watt-badge>
+      }
```
dzhavat, line 35: nitpick: according to Figma, the complete status is actually a download link, so this badge is not needed. It's ok for now. It can be replaced in the PR that implements the download functionality.

dzhavat (dismissed):

libs/dh/wholesale/feature-settlement-reports-v2/src/lib/table/dh-settlement-reports-status.component.ts
mimse: It will be done in the next PR :-)

dzhavat (approved):

PR #2864, opened 2024-05-07
Download link

mimse:


libs/dh/wholesale/feature-settlement-reports-v2/src/lib/table/dh-settlement-reports-table.component.html
```diff
@@ -55,7 +55,11 @@
     </ng-container>
 
     <ng-container *wattTableCell="columns['status']; header: t('columns.status'); let entry">
-      <dh-settlement-reports-status [status]="entry.statusType" />
+      <dh-settlement-reports-status
+        [status]="entry.statusType"
+        [reportId]="entry.id"
+        (download)="downloadReport($event)"
+      />
```
dzhavat, line 62: suggestion: wouldn't it be better if the Download button was not part of `dh-settlement-reports-status`? It could simple be an `if/else` here depending on the status instead of going in a component only to send an event out.

mimse: When we end up with logic around the status in two places instead of one. Right now the status component handle every case the status enum has.

dzhavat, line 62: suggestion: `reportId` is not doing anything special in `dh-settlement-reports-status` so it can be skipped. Instead, it can be directly sent to `downloadReport()`, e.g.

```suggestion
      <dh-settlement-reports-status
        [status]="entry.statusType"
        (download)="downloadReport(entry.id)"
      />
```

The click event in `dh-settlement-reports-status` can then just emit.

mimse: Good idea, looks better :100:

dzhavat (approved):

PR #2872, opened 2024-05-08
feat: request calculations validations changes

ManBearTM:


libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation-validators.ts
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+import { AbstractControl, ValidatorFn } from '@angular/forms';
+import { WattRange, dayjs } from '@energinet-datahub/watt/date';
+
+type rangeControl = AbstractControl<WattRange<string>>;
+
+export const max31DaysDateRangeValidator: ValidatorFn = ({ value }: rangeControl) => {
+  if (!value?.end || !value?.start) return null;
+  return dayjs(value.end).diff(value.start, 'days') > 30 ? { max31DaysDateRange: true } : null;
```
dzhavat, line 24: question: the error message says ".. cannot be longer than 31 days". Is the 31st day included or not? Seems like the error will be shown when selecting 31 days 🤔

libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.html
```diff
@@ -33,14 +33,8 @@
           [max]="maxDate"
           [formControl]="form.controls.period"
         >
-          @if (form.controls.period.errors?.maxOneMonthDateRange) {
+          @if (form.controls.period.errors?.max31DaysDateRange) {
           <watt-field-error> {{ t('maxPeriodLength')}} </watt-field-error>
-          } @if (form.controls.period.errors?.startDateCannotBeAfterEndDate) {
-          <watt-field-error> {{ t('startDateCannotBeAfterEndDate')}} </watt-field-error>
```
dzhavat, line 39: suggestion: remove unused translations

ManBearTM: Good catch!
https://github.com/Energinet-DataHub/greenforce-frontend/pull/2872/commits/f21154d53e11fba9ec2a88db2c9cb40283cf07c2

libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation-validators.ts
ManBearTM: You would think so (and it also caught me by surprise), but that is not the case. I believe it is because it is diffing in whole days, rounding down. So for march, the date range for all 31 days looks something like this:

`01-03-2024 00:00:00:000 – 31-03-2024 23:59:59:999`

Which is just one millisecond shy of 31 days.

But it is definitely confusing, so I added a comment explaining the condition: https://github.com/Energinet-DataHub/greenforce-frontend/pull/2872/commits/7c8e1107bc02652cb5b61f3239d2db4d8ce8331d

dzhavat: What about selecting 10th to 10th next month? Is that a valid value or must the period be one full month (1st to 28th/29th/30th/31st)?

ManBearTM: Yes, 10th to 10th is a valid selection and yeah, it should always be 31 days, regardless of the current month 🤷 I will be adding new calculation types that are restricted to full months only, but those will be using the month selection mode 🙂

dzhavat: Sounds good :)

dzhavat (approved):

PR #2884, opened 2024-05-14
fix: targetDefaults for test targets

ManBearTM:
## Description

This PR should prevent app-dh:generate from running before EO tests.


frontend-specialisten (changes requested):
It's better but it seems that `nx run dh-shared-domain:generate` is still executed **(IT WORKS MY MISTAKE)**

frontend-specialisten (approved):

PR #2895, opened 2024-05-15
Feature/navigation tabs on admin

mimse:


dzhavat (approved):
praise: looks good 💯 

nitpick: `routing` libs begin to look less relevant. Maybe we can drop them entirely at some point 🤔

PR #2901, opened 2024-05-16
EDI: Add endpoint for requesting wholesale settlements

MWO1024:


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Mutation.cs
```diff
@@ -123,17 +123,48 @@ await client
...
+                },
+                cancellationToken)
+            .ConfigureAwait(false);
+        return true;
+    }
+
+    public async Task<bool> CreateWholesaleSettlementRequestAsync(
+        EdiB2CWebAppCalculationType calculationType,
+        string startDate,
+        string? endDate,
```
ManBearTM, line 145: suggestion: use NodaTime.Interval here instead of separate arguments for start and end. we have logic that automatically "translates" frontend intervals to backend intervals (they are not fully equal).

ManBearTM (commented):
Looks good, just a single suggestion and I think we can merge this. There are a few things I would like to change, but I'd rather do that in a new PR together with my frontend changes. In short, I would like to make the GraphQL mutations match the frontend better by:

1. Consolidating the mutations into a single mutation that covers both cases. That way it matches the behavior of the frontend, which is a single form.
2. Remove the arguments that are not needed by the frontend, for now. I'd rather add them later as needed, it makes it easier to evolve the schema and may prevent confusion caused by unused arguments.

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Mutation.cs
MWO1024: Should we make this particular interval change now, or as part of the changes mentioned above, and thus in a later PR?

ManBearTM: Preferably in this PR, since you are the expert on how to serialize the values in accordance with what the backend client expects 🙂

PR #2907, opened 2024-05-18
use bun

mimse:
Replaced node and yarn with a new javascript runtime called [Bun](https://bun.sh/)
Bun can replace both yarn and node, because it is both a package manger like npm, yarn and a runtime.

It is a lot faster than node and yarn

Node setup and dependency install with yarn

image

Bun set and dependency install

image

If this is merged we can expect to see a minimum of 30 secs pr job taken of the CI time


package.json
```diff
@@ -202,18 +203,17 @@
...
     "typescript": "5.4.5",
     "typescript-eslint": "7.3.0"
   },
   "volta": {
     "//node-comment": "Keep this in sync with 'engines' config at the top of the current file",
     "node": "20.11.0",
-    "//yarn-comment": "Keep this in sync with GitHub actions pipelines",
-    "yarn": "4.0.2"
+    "//bun-comment": "Keep this in sync with GitHub actions pipelines",
+    "bun": "1.1.8"
```
dzhavat, line 214: issue: doesn't look like Volta supports Bun yet https://github.com/volta-cli/volta/issues/1465

mimse: Can you try volta install bun? It worked on mac

dzhavat: It worked on Windows too (had to use Admin privileges when running the command). The app started without issues as well. Should we wait for official support then? If not, things might break once it's added (and it works in a different way). 🤔

mimse: I guess we will fix it, if it breaks when the official lease lands :-)

NicolajAaH (approved):

PR #2928, opened 2024-05-24
feat: Add option to select all energy suppliers in request settlement report

FirestarJes:


libs/dh/wholesale/feature-settlement-reports-v2/src/lib/modal/dh-request-settlement-report-modal.component.ts
```diff
@@ -173,7 +175,14 @@ export class DhRequestSettlementReportModalComponent extends WattTypedModal {
     CalculationType.ThirdCorrectionSettlement,
   ]);
   gridAreaOptions$ = this.getGridAreaOptions();
-  energySupplierOptions$ = getActorOptions([EicFunction.EnergySupplier]);
+  energySupplierOptions$ = getActorOptions([EicFunction.EnergySupplier]).pipe(
+    tap((options) =>
```
defectiveAi, line 179: Can you use startsWith instead of tap? Should be simpler.

FirestarJes: I'm not sure we can do this here, as getActorOptions returns an WattsOptions with nested Wattoption objects in it, so i need to add to the WattsOptions.

PR #2947, opened 2024-05-31
fix(watt): make dropdown emit final value after filter + selection

dzhavat:
### Watt
- dropdown component did not emit final value when using filter + selection in multi mode. This PR attempts to fix that. See test case to understand the issue.

## References

- #0000


mimse (approved):
Praise: Nice test

Lets hope there is no side effects :-)

PR #2967, opened 2024-06-11
bug: Add timezone to date when requesting report

FirestarJes:


libs/dh/wholesale/feature-settlement-reports-v2/src/lib/modal/dh-request-settlement-report-modal.component.ts
```diff
@@ -275,8 +278,8 @@ export class DhRequestSettlementReportModalComponent extends WattTypedModal {
             calculationType: calculationType as CalculationType,
             includeBasisData,
             period: {
-              start: new Date(period.start),
-              end: period.end ? new Date(period.end) : null,
+              start: dayjs(period.start).tz(danishTimeZoneIdentifier).toDate(),
```
mimse, line 281: Suggestion: Try to change `WattRange<string>` to `WattRange<Date>` when we dont need to parse the date

mimse (changes requested):

dzhavat (approved):

libs/dh/wholesale/feature-settlement-reports-v2/src/lib/modal/dh-request-settlement-report-modal.component.ts
FirestarJes: We could do that, but that will then require changes to the "dhIsPeriodOneFullMonth" function, which expects a string, When looking at this function, it might also be susecptible to the same problem, but i'm not sure.

This current change, keeps the changes only to this component. that said it might be worth looking into updating it to WattRange<Date>

dzhavat: Looks like `dhIsPeriodOneFullMonth` requires the same fix as well (i.e. converting the date to Danish timezone). Let's make it. At a later point, a `dayjsWithTZ` (for example) function can be consider in order to remove the need to always specify a timezone when working with dates.

FirestarJes: I've updated it to use `WattRange<Date>` it appears to work.

mimse (approved):

dzhavat (approved):

PR #2999, opened 2024-06-13
feat: add progress field to graphql calculation

ManBearTM:
## Description

This PR adds a progress field to the GraphQL CalculationType. This new field is used to show a watt-progress-tracker on the calculations page.

## References

- [team-mosaic/#225](https://app.zenhub.com/workspaces/mosaic-60a6105157304f00119be86e/issues/gh/energinet-datahub/team-mosaic/225)


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Enums/CalculationProgressStatus.cs
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Energinet.DataHub.WebApi.GraphQL.Enums;
+
+public enum CalculationProgressStatus
+{
+    [GraphQLName("pending")]
```
mimse, line 19: Question: Why lower case here, and not in ProgressStep?

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/Calculation/CalculationType.cs
```diff
@@ -55,7 +55,57 @@ protected override void Configure(IObjectTypeDescriptor<CalculationDto> descript
...
+
+    private static CalculationProgress GetCalculationProgress(
+        CalculationProgressStep step,
+        CalculationOrchestrationState state) =>
+        step switch
+        {
+            CalculationProgressStep.Schedule => state switch {
+                CalculationOrchestrationState.Scheduled => new(step, CalculationProgressStatus.Pending, true),
+                CalculationOrchestrationState.Calculating => new(step, CalculationProgressStatus.Completed, false),
+                CalculationOrchestrationState.CalculationFailed => new(step, CalculationProgressStatus.Completed, false),
```
mimse, line 83: Suggestion: Should this not be CalculationProgressStatus.Failed?

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/Calculation/CalculationType.cs
```diff
@@ -55,7 +55,57 @@ protected override void Configure(IObjectTypeDescriptor<CalculationDto> descript
...
             });
     }
+
+    private static CalculationProgress GetCalculationProgress(
+        CalculationProgressStep step,
+        CalculationOrchestrationState state) =>
+        step switch
+        {
+            CalculationProgressStep.Schedule => state switch {
+                CalculationOrchestrationState.Scheduled => new(step, CalculationProgressStatus.Pending, true),
```
mimse, line 81: Question: What does current do? I dont see a pattern :-)

mimse (changes requested):
Looks good, nice with tests. Have a few quetions

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/Calculation/CalculationType.cs
ManBearTM: No, but I understand why the logic is a bit tricky to follow. There are 3 steps, `Schedule`, `Calculate` and `ActorMessageEnqueue`. That particular line gets the status of the first step (`Schedule`) based on the value of `CalculationOrchestrationState`. If the value of `CalculationOrchestrationState` is `CalculationFailed`, then it is implied that the first step must be completed, since the calculation cannot start before it is scheduled.

Take the following screenshot of the progress tracker (which is what this field is going to be used for). This would be how it should look if the state is `CalculationOrchestrationState.CalculationFailed`. Notice how the first step (Schedule) is completed. It is only the second step (Calculate) that should have the failed status.

progress-tracker

ManBearTM: But if you have any ideas on how to make this logic more clear, then that would be awesome!

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/Calculation/CalculationType.cs
ManBearTM: `current` tells the frontend what the currently active step is. It could probably be inferred from context in the frontend by doing things like getting the last step that is executing, failed or completed, but I wanted to keep it explicit in the API, since it is easy to accidentally break this assumption (by introducing optional steps, non-linear steps etc). There is a pattern though, you just have to keep in mind that we have the `CalculationOrchestrationState` enum with 8 steps that is broken down into 3 steps:

```
Scheduled -> Schedule
Calculating -> Calculate
CalculationFailed -> Calculate
Calculated -> Calculate
ActorMessagesEnqueuing -> Message Enqueue
ActorMessagesEnqueued -> Message Enqueue
ActorMessagesEnqueuingFailed -> Message Enqueue
Completed -> Message Enqueue
```

Fx. if the current `CalculationOrchestrationState` is `Scheduled`, then `current` should be true for the first step (Schedule), but false for the other two steps. Likewise, if the `CalculationOrchestrationState` is any of `Calculating`, `CalculationFailed` or `Calculated`, then `current` should be true for the second step (Calculate) and false for the others.

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Enums/CalculationProgressStatus.cs
ManBearTM: So it can be used as direct input to `watt-progress-tracker` ([code](https://github.com/Energinet-DataHub/greenforce-frontend/blob/c39edd45add167f0bb2cf58cde2f40c020eb0b1b/libs/watt/src/lib/components/progress-tracker/src/watt-progress-tracker-step.component.ts#L58)). But I'm not sure what the correct approach is here. The options are:

1. Convert this specific enum to lowercase in GraphQL, then pass the value directly (this is the current approach). Simple, but breaks the naming convention of enums.
3. Do NOT convert to lowercase in GraphQL, but then call `toLowerCase()` in the template. Kinda annoying.
4. Do NOT convert to lowercase in GraphQL, but require UPPERCASE in `watt-progress-tracker` input instead. Goes against the convention of our watt components.
5. Do NOT convert to lowercase in GraphQL, but accept both lowercase and UPPERCASE in `watt-progress-tracker`. Adds additional complexity to the component.
6. Convert ALL enums to lowercase in GraphQL. Major refactor required, but keeps things consistent at least.
7. Use some sort of transform middleware in Apollo/codegen that automatically transforms enums to lowercase in the frontend. Adds a lot of complexity and possibly confusion (why is the type of this enum different from the schema?).

I think I just went with 1, since it was the simpler approach. Perhaps if we make the `CalculationProgressStatus` ObjectType more generic (`ProgressStatus` instead) and add a comment that it is a `watt-progress-tracker` status?

mimse: Maybe the mapping between the enum and watt-progress-tracker should be in the frontend. So the BFF doesnt have to know about the implementation details of watt-progress-tracker?
Maybe watt-progress-tracker can take a transform function that does the custom mapping?

mimse:
Praise: I like the new code flow, easier to reason about and understand. Great job 💯

mimse (approved):

PR #3270, opened 2024-06-14
refactor(dh): rename property of dropdown translator directive

dzhavat:
## Description

Necessary to avoid clashes with global `translator` attribute. See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate

## References

- #0000


mimse (approved):
Praise: Good to know :-)

Issue #3016, opened 2024-06-13
refactor dh-filters to use signals


mimse:
Refactor ./dh/esett/feature-outgoing-messages/src/lib/filters/dh-filters.component.ts to use the new signal syntax

PR #3301, opened 2024-06-20
Refactor/refactor dh filters to use signals

mimse:
https://github.com/Energinet-DataHub/greenforce-frontend/issues/3016


libs/dh/esett/feature-outgoing-messages/src/lib/filters/dh-filters.component.ts
```diff
@@ -114,17 +119,32 @@ export class DhOutgoingMessagesFiltersComponent implements OnInit {
...
+      actorNumber: dhMakeFormControl(actorNumber),
+      status: dhMakeFormControl(status),
+      period: dhMakeFormControl(period),
+      created: dhMakeFormControl(created),
+      latestDispatch: dhMakeFormControl(latestDispatch),
     });
 
+    this.formReset$
+      .pipe(takeUntilDestroyed(this.destoryRef))
+      .subscribe(() => this.formGroup.reset(this.initial()));
```
dzhavat, line 146: question: is there a need to reset the form this way? Doesn't the reset button do it properly?

mimse: As fare as I can tell it only resets the search input field, and not the reset

dzhavat: Hmmm ... just tested it without the `formReset$` and it seems to work. :)

mimse: You are right, has been removed.

dzhavat (approved):

PR #3303, opened 2024-06-20
feat(dh): apollo signal api

ManBearTM:
This PR aims to simplify how you fetch data from GraphQL by using signals while at the same time fix many of the quirks and inconsistencies of the Apollo API. To that end, I have created a new library for use within components. The API is intentionally very close to that of the Apollo hooks, since hooks and signals share many of the same characteristics. It looks like this:

```ts
import { Component, computed } from '@angular/core';
import {
  CreateCalculationDocument,
  GetCalculationByIdDocument,
} from '@energinet-datahub/dh/shared/domain/graphql';
import { mutation, query } from '@energinet-datahub/dh/shared/util-apollo';

@Component({
  selector: 'dh-zoneless',
  template: `
    @if (result.loading()) {
      <p>Loading...</p>
    } @else if (result.error()) {
      <p>Error</p>
    } @else {
      <p>State: {{ calculation()?.state }}</p>
    }

    <button (click)="result.refetch()">Refetch</button>
    <button (click)="create.mutate()">Create</button>
  `,
  standalone: true,
})
export class DhZonelessComponent {
  result = query(GetCalculationByIdDocument, {
    variables: {
      id: '43ab77c6-4404-4621-905e-fa0ab61a95a1',
    },
  });

  calculation = computed(() => this.result.data()?.calculationById);

  create = mutation(CreateCalculationDocument, {
    variables: {
      input: {
        // ...
      }
    }
  });
}
```


mimse (approved):
Praise: Outstanding work! cannot wait to try it out 🥇

PR #3306, opened 2024-06-21
Refactored user overview to use graphql

mimse:


libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
```diff
@@ -176,23 +185,32 @@ export class DhAdminUserManagementDataAccessApiStore
...
+          userRoleIds: userRoleFilter,
+          searchText,
+          userStatus: statusFilter,
+        },
+      })
+      .pipe(
+        map((response) => ({
+          loading: response.loading,
+          error: Boolean(response.error) || (response.errors?.length ?? 0) > 0,
+          data: response.data?.userOverviewSearch,
```
dzhavat, line 211: suggestion: should probably fallback to an empty array otherwise it will end up being `undefined`


```suggestion
          data: response.data?.userOverviewSearch ?? [],
```

libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
```diff
@@ -122,24 +136,22 @@ export class DhAdminUserManagementDataAccessApiStore
           users: [],
         });
       }),
-      switchMap((fetchUsersParams) =>
-        this.getUsers(fetchUsersParams).pipe(
-          tapResponse(
-            (response) => {
+      switchMap((fetchUsersParams) => {
+        return this.getUsers(fetchUsersParams).pipe(
+          tap((response) => {
```
dzhavat, line 141: suggestion: add `catchError` either here (after the `tap`) or as part of `getUsers`. The `tapResponse` operator handled the errors for us before (e.g. 404, 500) but now we'll need to do it. Or maybe we can still use `tapResponse`. I don't think it's tied to NgRx anymore.

libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.spec.ts
dzhavat, line 1: praise: nice to see how Apollo can be used with unit tests 💯

libs/dh/admin/feature-user-management/src/lib/overview/dh-users-overview-table.component.html
```diff
@@ -20,7 +20,7 @@
     [columns]="columns"
     (rowClick)="onRowClick($event)"
     [activeRow]="activeRow"
-    sortBy="email"
+    [sortBy]="UserOverviewSortProperty.Email"
```
dzhavat, line 23: issue: noticed that this doesn't work because of capitalization. The `columns` object has an `email` property (starts with small `e`) whereas `UserOverviewSortProperty.Email` value starts with capital `E`.

dzhavat (approved):

PR #3319, opened 2024-06-28
Refactor/create graphql to replace v1marketparticipantuserupdateuseriden

mimse:


apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
```diff
@@ -376,6 +389,7 @@ type Query {
   userAuditLogs(id: UUID!): [UserAuditedChangeAuditLogDto!]!
   userRoleById(id: UUID!): UserRoleWithPermissionsDto!
   userProfile: GetUserProfileResponse!
+  userById(id: UUID!): GetUserResponse!
```
ManBearTM, line 392: nitpick: I know you are just using the dto from markpart which is totally fine, but maybe we should consider creating our own `User` type, so it looks like this `userById(id: UUID!): User!`. And then we could even consider using that for the `UserOverviewItemDto` as well, since they look almost the same. But doesn't have to be in this PR.

libs/dh/admin/data-access-api/src/lib/dh-admin-edit-user.store.ts
```diff
@@ -36,11 +43,11 @@ const initialState: State = {
 
 @Injectable()
 export class DhAdminEditUserStore extends ComponentStore<State> {
-  private readonly userRolesStore = inject(DhAdminUserRolesStore);
+  private readonly apollo = inject(Apollo);
```
ManBearTM, line 46: question: Have you tried with the new `mutation` utility? I'm just interested to know how or if it works with stores. Not that you have to rewrite everything again just for that 🙂

libs/dh/admin/data-access-api/src/lib/dh-admin-edit-user.store.ts
```diff
@@ -70,31 +77,55 @@ export class DhAdminEditUserStore extends ComponentStore<State> {
...
+                        response.data?.updateUserRoleAssignment.errors[0].statusCode,
+                        response.data?.updateUserRoleAssignment.errors[0]
+                      );
+                    }
+                  },
+                  () => {
+                    this.setSaving(ErrorState.GENERAL_ERROR);
+                  }
+                )
+              );
```
ManBearTM, line 128: suggestion: Have you considered the possibility of combining these two mutations into one (in the BFF)? There is quite a bit of complexity in determining if the mutation is successful or not. It would be nice if it was just something like this:

```ts
return this.apollo
  .mutate({
    mutation: UpdateUserDocument,
    refetchQueries: [UserOverviewSearchDocument, GetUserByIdDocument],
    variables: {
      id,
      firstName,
      lastName,
      phoneNumber,
      roles,
    },
  })
  .pipe(
    tapResponse(
      (response) => {
        if (response.data?.success)  {
          this.setSaving(SavingState.SAVED);
          onSuccessFn();
        } else if (response.data.errors) {
          this.setSaving(ErrorState.GENERAL_ERROR);
          onErrorFn(
            response.data.errors[0].statusCode,
            response.data.errors[0]
          );
        } 
      },
      () => {
        this.setSaving(ErrorState.GENERAL_ERROR);
      }
    )
  );
```

Or with the new util:

```ts
updateUser = mutation(UpdateUserDocument, {
  refetchQueries: [UserOverviewSearchDocument, GetUserByIdDocument],
});

// ...

await updateUser
  .mutate({
    input: {
      id,
      firstName,
      lastName,
      phoneNumber,
      roles,
    }
  })
  .then(response => {
    if (response.data?.success)  {
      this.setSaving(SavingState.SAVED);
      onSuccessFn();
    } else if (response.data.errors) {
      this.setSaving(ErrorState.GENERAL_ERROR);
      onErrorFn(response.data.errors[0].statusCode, response.data.errors[0]);
    } else if (response.error) {
      this.setSaving(ErrorState.GENERAL_ERROR);
    }
  },
});
```

ManBearTM (commented):
Nice work, just a few comments/questions, shouldn't block merge 🙂

libs/dh/admin/data-access-api/src/lib/dh-admin-edit-user.store.ts
mimse: I had already written the mutation before your util were merged in :-)

apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
mimse: Good idea, will do it in other PR

ManBearTM (approved):

PR #3322, opened 2024-07-01
Look at permissions instead of marketrole

mimse:
Fixes: https://github.com/Energinet-DataHub/team-mosaic/issues/237


libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.ts
```diff
@@ -186,7 +193,7 @@ export class DhWholesaleRequestCalculationComponent {
           this._selectedEicFunction
         );
 
-        const excludeProcessTypes = this.getExcludedProcessTypes(this._selectedEicFunction);
+        const excludeProcessTypes = await this.getExcludedProcessTypes();
```
ManBearTM, line 196: suggestion: I don't like this await here, the subscribe `next` handler does not accept promises like this. Things might get weird when selecting different actors. May I suggest this signal-based approach which also gets rid of that `combineLatest`/`reduce` logic which seems more complicated than it needs to be:

```ts
private _permissionService = inject(PermissionService);

requestAggregatedMeasuredDataView = this.hasPermission('request-aggregated-measured-data:view');
requestWholesaleSettlementView = this.hasPermission('request-wholesale-settlement:view');

excludeCalculationTypes = computed(() => {
  const aggregation = this.requestAggregatedMeasuredDataView() ? [] : aggregationCalculationTypes;
  const wholesale = this.requestWholesaleSettlementView() ? [] : wholesaleCalculationTypes;
  return [...aggregation, ...wholesale];
});

calculationTypeOptions = computed(() =>
  dhEnumToWattDropdownOptions(CalculationType, null, this.excludeCalculationTypes())
);

hasPermission(permission: Permission) {
  return toSignal(this._permissionService.hasPermission(permission));
}
```

mimse: I like ❤️

libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.ts
```diff
@@ -295,16 +308,7 @@ export class DhWholesaleRequestCalculationComponent {
...
-      selectedEicFunction === EicFunction.EnergySupplier
-    )
-      return wholesaleCalculationTypes;
-
-    if (selectedEicFunction === EicFunction.SystemOperator)
-      return [CalculationType.Aggregation, CalculationType.BalanceFixing];
-
-    return [];
+  hasPermission(permission: Permission) {
+    return toSignal(this._permissionService.hasPermission(permission));
```
ManBearTM, line 312: suggestion: maybe add options with `{ initialValue: false }` so that the signal becomes of type `Signal<boolean>` instead of `Signal<boolean | undefined>`.

ManBearTM: I just went ahead and fixed it:
d6ec0066f1a86e630c7da923a275e8aa6f8c62f2

ManBearTM (approved):

PR #3329, opened 2024-07-03
refac(dh): grid area in create calculation

ManBearTM:
This PR moves all of the logic concerning grid areas in create calculations into its own component. I had to skip the cypress test since I just couldn't get it to work (the query would not return the mocked data). I couldn't justify spending any more time on that, so I'm skipping it for now. We will have a talk about it later.

### DIFF
--- a/libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
+++ b/libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
@@ -46,7 +46,8 @@ limitations under the License.
           }}"
         />
       } @else if (
-        latestPeriodEnd && formGroup.controls.calculationType.value === "BALANCE_FIXING"
+        latestPeriodEnd &&
+        formGroup.controls.calculationType.value === CalculationType.BalanceFixing
       ) {
         <watt-validation-message
           size="normal"
--- a/libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
+++ b/libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
@@ -27,7 +27,8 @@ limitations under the License.
       novalidate
       [formGroup]="formGroup"
       (ngSubmit)="
-        latestPeriodEnd && formGroup.controls.calculationType.value === 'BALANCE_FIXING'
+        latestPeriodEnd &&
+        formGroup.controls.calculationType.value === CalculationType.BalanceFixing
           ? (showPeriodWarning = true)
           : modal.close(true)
       "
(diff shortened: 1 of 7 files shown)


libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
```diff
@@ -21,162 +21,111 @@
...
-              {{ t("scheduling.SCHEDULED") }}
-            </watt-filter-chip>
-          </div>
+    <form
+      id="create-calculation"
+      class="grid"
+      novalidate
+      [formGroup]="formGroup"
+      (ngSubmit)="
+        latestPeriodEnd && formGroup.controls.calculationType.value === 'BALANCE_FIXING'
```
mimse, line 30: Suggestion: import and expose CalculationType enum as a public field, so you can use it in the template. To have type safety

libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
```diff
@@ -21,162 +21,111 @@
...
-            ) {
-              <watt-field-error>
-                {{ t("create.period.invalid") }}
-              </watt-field-error>
-            }
-          </watt-datepicker>
+      <!-- Process type -->
+      <watt-dropdown
+        [label]="t('create.calculationType.label')"
+        [formControl]="formGroup.controls.calculationType"
```
mimse, line 83: Praise: Nice refactor to formControl

mimse (changes requested):
Good work, just a few comments

libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
ManBearTM: Good idea!
37008b8fc41d90453fbff87a6bf18607f67c9f07

libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
```diff
@@ -21,162 +21,112 @@
...
+          type="danger"
+          icon="danger"
+          message="{{
+            t('create.quarterlyResolutionTransitionError', {
+              resolutionTransitionDate: resolutionTransitionDate | wattDate
+            })
+          }}"
+        />
+      } @else if (
+        latestPeriodEnd && formGroup.controls.calculationType.value === "BALANCE_FIXING"
```
mimse, line 49: Suggestion: Do the same here :-)

ManBearTM: b0a7711c7f0e94b92570696dac1126837411c45f

mimse (approved):

PR #3332, opened 2024-07-03
refactor(dh): change update user role to use GraphQL 

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/admin/data-access-api/src/lib/dh-admin-user-role-edit-data-access-api.store.ts
+++ b/libs/dh/admin/data-access-api/src/lib/dh-admin-user-role-edit-data-access-api.store.ts
@@ -19,14 +19,15 @@ import { HttpErrorResponse, HttpStatusCode } from '@angular/common/http';
 import { ComponentStore } from '@ngrx/component-store';
 import { tapResponse } from '@ngrx/operators';
 import { exhaustMap, filter, Observable, tap } from 'rxjs';
+import { Apollo } from 'apollo-angular';
 
 import { ErrorState, LoadingState } from '@energinet-datahub/dh/shared/data-access-api';
-import {
-  MarketParticipantUserRoleHttp,
-  MarketParticipantUpdateUserRoleDto,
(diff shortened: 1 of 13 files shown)


libs/dh/admin/data-access-api/src/lib/dh-admin-user-role-edit-data-access-api.store.ts
```diff
@@ -38,59 +39,86 @@ const initialState: DhEditUserRoleState = {
...
+                    this.userRolesStore.updateRoleById({
+                      id: userRoleId,
+                      name: updatedUserRole.name,
+                    });
+
+                    onSuccessFn();
+                  }
+
+                  if (data?.updateUserRole.errors?.length) {
+                    const [firstError] = data.updateUserRole.errors;
```
mimse, line 99: Praise: nice synxtax, need to rememeber that one

libs/dh/shared/data-access-mocks/src/lib/admin.ts
```diff
@@ -189,10 +190,20 @@ function getUserOverviewQuery() {
...
 }
 
-function putMarketParticipantUserRoleUpdate(apiBase: string) {
-  return http.put(`${apiBase}/v1/MarketParticipantUserRole/Update`, async () => {
+function updateUserRoleMutation() {
+  return mockUpdateUserRoleMutation(async () => {
     await delay(mswConfig.delay);
-    return new HttpResponse(null, { status: 200 });
+
+    return HttpResponse.json({
```
mimse, line 197: Suggestion: I have started to look at the variables and ex. if id = 2 then make it fail. Then it is easier to the error handling when using mock data

mimse (changes requested):

libs/dh/shared/data-access-mocks/src/lib/admin.ts
dzhavat: Tried to include an error state as well.

mimse (approved):

PR #3339, opened 2024-07-04
refac(dh): clean up create calculations

ManBearTM:
This PR removes obsolete fields from the create calculations modal + uses vater for layout.


libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
```diff
@@ -57,75 +60,61 @@
...
-        </watt-filter-chip>
-      </div>
+        <!-- Execution type -->
+        <vater-stack gap="s" fill="horizontal" align="flex-start">
+          <label>{{ t("create.executionType") }}</label>
+          <vater-stack direction="row" gap="s" align="center" style="height: 44px">
+            <watt-filter-chip choice name="executionType" value="ACTUAL" [selected]="true">
+              {{ t("executionTypes.ACTUAL") }}
+            </watt-filter-chip>
+            <watt-filter-chip choice name="executionType" value="SIMULATION" [disabled]="true">
```
mimse, line 84: Question: Should value be translated?

mimse (approved):
Looks good

libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
ManBearTM: No, but it probably should come from an enum - problem is that we do not yet have this functionality, so this is just a visual placeholder

PR #3340, opened 2024-07-04
fix(dh): feature flag for minimum date in create calculation

ManBearTM:
Remember to set min date in template


libs/dh/wholesale/feature-calculations/src/lib/create/create.component.html
```diff
@@ -93,6 +93,7 @@
           [formControl]="formGroup.controls.dateRange"
           [rangeMonthOnlyMode]="monthOnly.includes(formGroup.controls.calculationType.value)"
           [range]="true"
+          [min]="minDate"
```
dzhavat, line 96: question: not sure whether it's still relevant, but in my magic #3333 PR, I also removed `minDate` in `dh-wholesale-request-calculation.html`

ManBearTM: Right, good catch. I added it back: 263c06e97e664e8fb655092e40e98d14b5dbeeb9

I'll later make it so that request-calculation uses the same grid area dropdown component as in calculations

dzhavat (approved):

PR #3342, opened 2024-07-05
feat(watt): new vater functions

ManBearTM:
- Add new center utility to vater
- Change offset to use padding instead of margin
- Make it possible to customize flex properties

### DIFF
--- a/libs/watt/src/lib/components/vater/vater-flex.component.ts
+++ b/libs/watt/src/lib/components/vater/vater-flex.component.ts
@@ -62,7 +62,7 @@ export class VaterFlexComponent {
   @Input()
   offset?: Spacing;
 
-  @HostBinding('style.margin')
+  @HostBinding('style.padding')
   get _offset() {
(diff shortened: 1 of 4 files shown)


libs/watt/src/lib/components/vater/vater-flex.component.ts
```diff
@@ -49,6 +49,18 @@ export class VaterFlexComponent {
   @HostBinding('style.flex-direction')
   direction: Direction = 'column';
 
+  @Input()
+  @HostBinding('style.--grow')
```
mimse, line 53: Question: Why not refactor to use signals? :-)

mimse (approved):

libs/watt/src/lib/components/vater/vater-flex.component.ts
ManBearTM: I will, eventually 😄 I'm trying to keep the PR's light

PR #3353, opened 2024-07-09
refactor(dh): use GraphQL to create user role

dzhavat:
## References

- #0000


mimse (approved):
Looks good, would properly get rid of the store in another PR :-)

PR #3360, opened 2024-07-10
refactor(dh): use GraphQL in user roles overview

dzhavat:
## References

- #0000


libs/dh/admin/data-access-api/src/lib/dh-admin-user-roles-management-data-access-api.store.ts
```diff
@@ -197,19 +111,17 @@ export class DhAdminUserRolesManagementDataAccessApiStore
...
-          }))
-        )
-      );
+    this.rolesOptions$ = this.transloco.selectTranslateObject('marketParticipant.marketRoles').pipe(
+      combineLatestWith(this.select((state) => state.roles)),
+      filter(([, roles]) => roles.length > 0),
+      // eslint-disable-next-line @ngrx/avoid-mapping-component-store-selectors
+      map(([keys, roles]) =>
+        roles.map((role: UserRoleItem) => ({
+          displayValue: `${role.name} (${keys[role.eicFunction]})`,
```
mimse, line 120: Suggestion: Lets make a displayValue on the UserRoleItem in BFF, like we have with actor and gridarea

libs/dh/admin/feature-user-roles/src/lib/drawer/dh-role-drawer.component.html
```diff
@@ -32,7 +32,7 @@ <h2>{{ userRoleWithPermissions.name }}</h2>
       </watt-drawer-heading>
     }
 
-    @if (basicUserRole?.status !== "Inactive") {
```
mimse, line 35: Suggestion: Use UserRole Status Enum in the template, by binding as a public field in the component

libs/dh/admin/shared/src/dh-role-status.component.ts
```diff
@@ -18,22 +18,21 @@ import { Component, Input, ChangeDetectionStrategy } from '@angular/core';
...
 
 @Component({
   selector: 'dh-role-status',
   standalone: true,
   template: `<ng-container *transloco="let t; read: 'admin.userManagement.roleStatus'">
-    @if (status === 'Active' || status === 'ACTIVE') {
-      <watt-badge type="info">{{ t('Active') }}</watt-badge>
-    } @else if (status === 'Inactive' || status === 'INACTIVE') {
-      <watt-badge type="warning">{{ t('Inactive') }}</watt-badge>
+    @if (status === 'ACTIVE') {
```
mimse, line 27: Suggestion: Use UseRoleStatus in the template here as well

mimse (changes requested):
Praise: Good work, just a few comments

libs/dh/admin/data-access-api/src/lib/dh-admin-user-roles-management-data-access-api.store.ts
dzhavat: This will not work with translations. The `eicFunction` value is translated here.
I can see that the same issue appears in the "Market participant" dropdown on the Users page.

PR #3362, opened 2024-07-10
fix(dh): improvements to `query` function

dzhavat:
### DataHub
- do not complete stream on error by moving `catchError` operator inside `valueChanges`
- add `fetchPolicy` on `refetch` method
- add test to verify behaviour. The test show that a query can be recovered from error state when calling `refetch`. The test DO NOT show that `loading` Signal is set to `true` after `refetch` is called because I couldn't figure out how to set the `query` in that state. However, this is what happens in the browser.

## References

- #0000

### DIFF
--- a/libs/dh/shared/util-apollo/src/lib/query.spec.ts
+++ b/libs/dh/shared/util-apollo/src/lib/query.spec.ts
@@ -107,6 +107,27 @@ describe('query', () => {
       expect(result.networkStatus()).toBe(NetworkStatus.ready);
     })));
 
+  it('should start a new query on refetch after it has errored', fakeAsync(() =>
+    TestBed.runInInjectionContext(() => {
+      const result = query(TEST_QUERY);
+      const op = controller.expectOne(TEST_QUERY);
+      op.flush({ errors: [new GraphQLError('TestError')] });
+      tick();
+
+      expect(result.error()).toBeInstanceOf(ApolloError);
+      expect(result.loading()).toBe(false);
+
+      result.refetch();
+
+      const newOp = controller.expectOne(TEST_QUERY);
+      const data = { __type: { name: 'Success' } };
+      newOp.flush({ data });
+      tick();
+
+      expect(result.data()).toEqual(data);
+      expect(result.loading()).toBe(false);
+    })));
+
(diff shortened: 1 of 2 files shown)


libs/dh/shared/util-apollo/src/lib/query.spec.ts
```diff
@@ -107,6 +107,27 @@ describe('query', () => {
       expect(result.networkStatus()).toBe(NetworkStatus.ready);
     })));
 
+  it('should start a new query on refetch after it has errored', fakeAsync(() =>
```
mimse, line 110: Praise: nice test

libs/dh/shared/util-apollo/src/lib/query.ts
```diff
@@ -150,7 +154,12 @@ export function query<TResult, TVariables extends OperationVariables>(
     refetch: (variables?: Partial<TVariables>) => {
       const result = firstValueFrom(result$.pipe(filter((result) => !result.loading)));
       const mergedVariables = { ...options$.value?.variables, ...variables } as TVariables;
-      options$.next({ ...options$.value, skip: false, variables: mergedVariables });
+      options$.next({
+        ...options$.value,
+        skip: false,
+        variables: mergedVariables,
+        fetchPolicy: 'network-only',
```
mimse, line 161: Question: Is it a good idea to have this hard-coded here? It needs be we can provide default option values here libs/dh/shared/data-access-graphql/src/lib/dh-graphql.providers.ts

mimse (approved):

libs/dh/shared/util-apollo/src/lib/query.ts
dzhavat: It might be worth moving to `dh-graphql.providers.ts` but this will affect all `watchQuery`s, not only the ones executed through the `query` util function. Not sure if that can make Apollo fetch data in cases where it could just take it from cache. Besides that, the setting here is only used when calling `refetch` so its limited even more.

mimse: Ah, if its only on refetch, I think its fine :-)

ManBearTM (approved):

Issue #3178, opened 2024-06-13
refactor watt-field to use signals


mimse:
Refactor ./watt/src/lib/components/field/watt-field.component.ts to use the new signal syntax

PR #3367, opened 2024-07-11
refac(watt): refactor watt-field to use signals + more

ManBearTM:
Closes https://github.com/Energinet-DataHub/greenforce-frontend/issues/3178

I also had to make the component subscribe to statusChanges, since otherwise the `isRequired` state would not be properly updated when dynamically changing form control validators (yes that is a use case now 😄). Also meant that I could change the component to use OnPush for change detection.


libs/watt/src/lib/components/field/watt-field.component.ts
```diff
@@ -14,88 +14,96 @@
...
+  wrapper = viewChild.required<ElementRef>('wrapper');
 
-  @HostBinding('class.watt-field--invalid')
-  get _hasError() {
-    return (
-      (this.control?.status === 'INVALID' && !!this.control?.touched) ||
-      (this._formGroupDirective?.submitted && this.control?.status === 'INVALID')
+  constructor() {
+    const status$ = toObservable(this.control).pipe(
+      filter((control) => control !== null),
```
mimse, line 91: Question: Should there be a unsubscribe somewhere? Or a takeUntilDestoryed?

mimse (approved):

mimse (commented):
Praise: Good job

libs/watt/src/lib/components/field/watt-field.component.ts
ManBearTM: No need, it completes itself:
https://github.com/angular/angular/blob/main/packages/core/rxjs-interop/src/to_observable.ts#L65

mimse: Perfect!

PR #3392, opened 2024-08-01
feat: receive completed time when getting calculation from wholesale

MadsDue:
## Description

To display when the calculation flow is completed for the b2c users (including enqueuing messages) we need to get the completed time when getting the calculation.

## References

- #0000


lasrinnil (approved):
LGTM

PR #3399, opened 2024-08-06
refactor(dh): use GraphQL to deactivate user role

dzhavat:
### DataHub
- replace REST endpoint to deactivate user role with GraphQL mutation
- remove unused REST endpoint
- set button in loading state while a request is pending
- adjust mocks
- rename store

## References

- #0000


libs/dh/admin/data-access-api/src/lib/dh-user-role-management.store.ts
```diff
@@ -22,38 +22,46 @@ import { Apollo } from 'apollo-angular';
 import type { ResultOf } from '@graphql-typed-document-node/core';
```
mimse, line 22: Suggestion: Maybe replace the store with signals in other PR

mimse (approved):
Praise: Looks good, just a comment

PR #3400, opened 2024-08-07
fix(dh): make gridArea in request calculation optional

ManBearTM:
- Make gridArea optional in GraphQL
- Remove empty check for nullable fields


mimse (approved):
Smukt!

dzhavat (approved):

PR #3414, opened 2024-08-13
feat(dh): set validation to market participant name

dzhavat:
### DataHub
- validate market participant name to be of max 150 characters
- use large modal when editting market participant

## References

- #0000


libs/dh/market-participant/actors/feature-actors/src/lib/create/dh-actors-create-actor-modal.component.ts
```diff
@@ -60,6 +60,7 @@ import { DhChooseOrganizationStepComponent } from './steps/dh-choose-organizatio
 import { DhNewOrganizationStepComponent } from './steps/dh-new-organization-step.component';
 import { DhNewActorStepComponent } from './steps/dh-new-actor-step.component';
 import { ActorForm } from './dh-actor-form.model';
+import { dhNameMaxLengthValidatorFn } from '../dh-name-max-length.validator';
```
mimse, line 63: Question: Should this be used on all names or just actor names? If so maybe the name should dhActornameMaxLengthValidatorFn?

mimse (dismissed):

libs/dh/market-participant/actors/feature-actors/src/lib/create/dh-actors-create-actor-modal.component.ts
dzhavat: Just the actor/market participant name. Just pushed a fix :)

mimse (approved):

PR #3422, opened 2024-08-14
fix(dh): prevent flaky tests by making GraphQLTestService non-static

ManBearTM:
## Description

This PR attempts to fix the flakyness of the GraphQL dotnet tests. In my debugging, I discovered that the mocked data was not always as expected, but instead it looked like it served "stale" mock data. I suspect it has to do with the static GraphQLTestService, since all tests were using the same mock instances. To fix it, I now require that you instantiate the GraphQLTestService in each test file. This makes the tests run a bit slower, since each test has to create the server from scratch. It is still in the order of seconds though, so it is no big deal (yet).


mimse (approved):
Good catch

PR #3427, opened 2024-08-15
fix(dh): better support for chips as form fields

ManBearTM:


libs/watt/src/lib/components/field/watt-field.component.ts
```diff
@@ -79,6 +80,8 @@ export class WattFieldComponent {
   chipMode = input(false);
   tooltip = input<string>();
 
+  untitled = computed(() => !this.label());
```
frontend-specialisten, line 83: Looks good to me, however maybe we should use `unlabelled` instead of `untitled` to keep the wording? 
```suggestion
  unlabelled = computed(() => !this.label());
```

ManBearTM: Good suggestion!

[0f76d81](https://github.com/Energinet-DataHub/greenforce-frontend/pull/3427/commits/0f76d81a7c6365ff51224c82186bd04a9db9406c)

frontend-specialisten (approved):

PR #3437, opened 2024-08-16
refetch overview when changing the user state

mimse:


dzhavat (approved):
nitpick: it might be that some of these calls are "too aggressive", meaning they will execute even if the mutation fails. Let's keep an eye on them/feedback and adjust if necessary.

PR #3441, opened 2024-08-19
feat(dh): download users 

dzhavat:
## References

- #0000


libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
```diff
@@ -213,6 +228,28 @@ export class DhAdminUserManagementDataAccessApiStore
     this.patchState({ sortProperty, direction });
   }
 
+  downloadUsers(onSuccess: (items: UsersToDownload) => void, onError: () => void) {
+    this.getAllUsersQuery.query({
```
mimse, line 232: Nikpick: Instead of using the callback pattern, I think I would have return an promise or observable with the users. In my opinon it makes it abit clear.

mimse (approved):

libs/dh/admin/data-access-api/src/lib/dh-admin-user-management-data-access-api.store.ts
dzhavat: Tried refactoring it to use a Promise in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3441/commits/31229030ce196f8e1035e3dd259d8a17dba560a3

mimse (approved):
Praise: looks good

PR #3443, opened 2024-08-20
Fill in the gaps

mimse:
Solves: https://github.com/Energinet-DataHub/team-raccoons/issues/463

### DIFF
--- a/apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Query/ImbalancePriceQuery.cs
+++ b/apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Query/ImbalancePriceQuery.cs
@@ -59,8 +59,8 @@ public partial class Query
             {
                 PriceAreaCode = priceAreaCode switch
                 {
-                    PriceAreaCode.AreaCode1 => Energinet.DataHub.WebApi.GraphQL.Enums.PriceAreaCode.Dk1,
-                    PriceAreaCode.AreaCode2 => Energinet.DataHub.WebApi.GraphQL.Enums.PriceAreaCode.Dk2,
+                    PriceAreaCode.AreaCode1 => Enums.PriceAreaCode.Dk1,
+                    PriceAreaCode.AreaCode2 => Enums.PriceAreaCode.Dk2,
                     _ => throw new ArgumentOutOfRangeException(nameof(priceAreaCode)),
                 },
                 Name = from,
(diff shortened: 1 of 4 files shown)


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Query/ImbalancePriceQuery.cs
```diff
@@ -78,16 +78,55 @@ static async Task<ImbalancePricePeriod> GetPricesAsync(DateTimeOffset from, Date
...
+                            Timestamp = missingTimestamp,
+                            Price = 0,
+                        };
+
+                        missingTimestamps.Add(missingPrice);
+                    }
+                }
+
+                previousTimestamp = currentTimestamp;
+            }
```
dzhavat, line 122: issue: the logic should also handle 15 min gaps. There could be days where the prices are in 15 min resolution, e.g.

From T-001

image

libs/dh/imbalance-prices/shell/src/lib/table-day-view/dh-table-day-view.component.ts
```diff
@@ -53,13 +57,17 @@ import { DhImbalancePricesForDay, DhImbalancePricesForDayProcessed } from '../dh
       </ng-container>
 
       <ng-container *wattTableCell="columns['price']; let entry">
-        <span [ngClass]="{ 'negative-price': entry.price < 0 }">
-          {{ entry.price | number: '1.5-6' }}
-        </span>
+        @if (entry.price === 0) {
```
dzhavat, line 60: question: can there ever be a price that is actually 0? In that case the UI will show a missing price badge. Wouldn't it be better if the prices was `null`/`undefined`?

mimse: The DTO does not allow for null / undefined. When we need to change the data structure in markpart @FirestarJes What do you think?

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Query/ImbalancePriceQuery.cs
mimse: @FirestarJes what gives? :-D what about the cut off date we talked about? :D

FirestarJes: I was under the impression we had a cut-off date where this was to implemented, wasn't aware it was already implemented. Well if we already have these we need to handle them

libs/dh/imbalance-prices/shell/src/lib/table-day-view/dh-table-day-view.component.ts
FirestarJes: I don't think there can actually ever be a "0" price, that would be missing a price, but should probably consult Irene on this one, as she will be able to answer this.

dzhavat: If the price can never be exactly 0, then it's fine to keep it like that.

mimse: Prices can in rare cases be 0, so need to find another solution

mimse: Is now set as null

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Query/ImbalancePriceQuery.cs
mimse: We are not yet switched to quaterly values, so have overridden it with hourly instead, so the data is right.

dzhavat (approved):
praise: perfect 💯

PR #3444, opened 2024-08-20
Fas permission for history

mimse:
Solves: https://github.com/Energinet-DataHub/team-raccoons/issues/761


libs/dh/admin/feature-user-roles/src/lib/drawer/tabs/dh-drawer-role-tabs.component.html
```diff
@@ -22,7 +22,7 @@
     <watt-tab [label]="t('permissions.tabLabel')">
       <dh-role-permissions [role]="role" />
     </watt-tab>
-    <watt-tab *dhPermissionRequired="['user-roles:manage']" [label]="t('history.tabLabel')">
+    <watt-tab *dhPermissionRequired="['fas']" [label]="t('history.tabLabel')">
```
dzhavat, line 25: suggestion: I can remember that some "History" tabs lazy-load their data (first when the tab is open), while others load data when the drawer gets open. Can you test whether history queries are not executed when the drawer gets open. Also test what happens when a user/actor is updated. History queries are refetched after that as well.

libs/dh/market-participant/actors/feature-organizations/src/lib/drawer/tabs/dh-organization-history.component.ts
```diff
@@ -0,0 +1,116 @@
...
+  selector: 'dh-organization-history',
+  standalone: true,
+  template: ` <ng-container
+    *transloco="let t; read: 'marketParticipant.organizationsOverview.drawer'"
+  >
+    <vater-stack [fill]="'horizontal'" [align]="'center'">
+      @if (isLoading()) {
+        <watt-spinner />
+      }
+    </vater-stack>
```
dzhavat, line 45: suggestion: it's just a copy/paste from before but it looks like the `vater-stack` element will still be in the DOM and be empty when `isLoading` is `false`. Let's move the `if` one level up.

```suggestion
    @if (isLoading()) {
      <vater-stack fill="horizontal" align="center">
        <watt-spinner />
      </vater-stack>
    }
```

libs/dh/market-participant/actors/feature-organizations/src/lib/drawer/tabs/dh-organization-history.component.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  isLoading = this.getAuditLogByOrganizationIdQuery.loading;
+  hasError = computed(() => this.getAuditLogByOrganizationIdQuery.error !== undefined);
+
+  auditLog = computed(
+    () =>
+      new WattTableDataSource<OrganizationAuditedChangeAuditLogDto>(
+        this.getAuditLogByOrganizationIdQuery.data()?.organizationAuditLogs
+      )
+  );
```
dzhavat, line 99: suggestion: this will create a new `WattTableDataSource` on new data. Material's advice is to update the `.data` property instead. Let's use that. Also let's have an empty array as fallback in case `.data()?.organizationAuditLogs` is `undefined` (as it looks like it could be).

mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3444/commits/32483794bfde07552fe022da0a2129e6dda3deb6

dzhavat (approved):
praise: looks good 💯

PR #3459, opened 2024-08-23
fix(dh): show 0 imbalance price in day view

dzhavat:
## References

- #0000


mimse (approved):
Praise: Nice find, Kim shown it to me yesterday, and I could not figure out why 0 showed up as "Mangler", but now I do !

PR #3470, opened 2024-08-27
feat(dh): log failed MitID logins to AppInsights

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/core/shell/src/lib/dh-core-login.component.ts
+++ b/libs/dh/core/shell/src/lib/dh-core-login.component.ts
@@ -17,11 +17,12 @@
 import {
   AfterViewInit,
   Component,
-  ViewChild,
   ViewEncapsulation,
   inject,
   signal,
+  viewChild,
 } from '@angular/core';
+import { MatProgressBarModule } from '@angular/material/progress-bar';
 import { TranslocoPipe } from '@ngneat/transloco';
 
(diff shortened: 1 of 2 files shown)


libs/dh/core/shell/src/lib/dh-core-login.component.ts
```diff
@@ -117,31 +119,42 @@ import { DhFeatureFlagDirective } from '@energinet-datahub/dh/shared/feature-fla
...
+
       setInterval(() => {
         this.progressBarValue.set(this.progressBarValue() + 1);
       }, 200);
+
       setTimeout(() => {
         localStorage.removeItem('mitIdRelogin');
-        this.mitIdButton.redirectToMitIdSignup();
-        this.showProgressBar.set(false);
-      }, 20000);
```
mimse, line 138: Issue: Before it redirected after 20 sec, now it does it after 2 sec?

mimse (changes requested):

libs/dh/core/shell/src/lib/dh-core-login.component.ts
dzhavat: Not really. Now it redirects after 22 sec - 20 sec for the outer `setTimeout` + 2 sec for the inner where the redirect is triggered.

mimse: You are right :-)

mimse (approved):

Issue #3062, opened 2024-06-13
refactor dh-market-participant-gridarea-overview to use signals


mimse:
Refactor ./dh/market-participant/grid-areas/overview/src/lib/dh-market-participant-gridarea-overview.component.ts to use the new signal syntax

Issue #3059, opened 2024-06-13
refactor dh-table to use signals


mimse:
Refactor ./dh/market-participant/actors/feature-organizations/src/lib/table/dh-table.component.ts to use the new signal syntax

Issue #3058, opened 2024-06-13
refactor dh-edit-modal to use signals


mimse:
Refactor ./dh/market-participant/actors/feature-organizations/src/lib/edit/dh-edit-modal.component.ts to use the new signal syntax

Issue #3045, opened 2024-06-13
refactor dh-actors-table to use signals


mimse:
Refactor ./dh/market-participant/actors/feature-actors/src/lib/table/dh-actors-table.component.ts to use the new signal syntax

PR #3471, opened 2024-08-27
Refactor-input-signal-v2

dzhavat:
## References

Closes 
- #3027
- #3043
- #3045
- #3058
- #3059
- #3062

### DIFF
--- a/libs/dh/market-participant/actors/feature-actors/src/lib/status-badge/dh-actor-status-badge.component.ts
+++ b/libs/dh/market-participant/actors/feature-actors/src/lib/status-badge/dh-actor-status-badge.component.ts
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
 import { TranslocoDirective } from '@ngneat/transloco';
 
 import { WattBadgeComponent } from '@energinet-datahub/watt/badge';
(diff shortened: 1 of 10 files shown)


libs/dh/market-participant/actors/feature-actors/src/lib/status-badge/dh-actor-status-badge.component.ts
```diff
@@ -43,5 +43,5 @@ import { DhEmDashFallbackPipe } from '@energinet-datahub/dh/shared/ui-util';
   imports: [TranslocoDirective, WattBadgeComponent, DhEmDashFallbackPipe],
 })
 export class DhActorStatusBadgeComponent {
-  @Input({ required: true }) status: ActorStatus | null | undefined = null;
+  status = input.required<ActorStatus | null | undefined>();
```
mimse, line 46: Question: If it is required, should it then be allowed to be null or undefined as well?

mimse (approved):

libs/dh/market-participant/actors/feature-actors/src/lib/status-badge/dh-actor-status-badge.component.ts
mimse: Is where a good reason for it to also be null or undefined?

dzhavat: According to the types, it can be `undefined`. So maybe we can get rid of the `null`

mimse (approved):

PR #3472, opened 2024-08-27
feat(dh): hide permission  tab behind a permission

dzhavat:
## References

- #0000


libs/dh/admin/feature-user-roles/src/lib/drawer/tabs/content/dh-role-permissions.component.ts
```diff
@@ -30,20 +30,20 @@ import { WattTableDataSource, WattTableColumnDef, WATT_TABLE } from '@energinet-
...
 })
-export class DhRolePermissionsComponent implements OnChanges {
-  @Input() role: DhUserRoleWithPermissions | null = null;
+export class DhRolePermissionsComponent {
+  role = input.required<DhUserRoleWithPermissions | null>();
 
-  readonly dataSource: WattTableDataSource<DhUserRolePermissionDetails> = new WattTableDataSource(
-    undefined
-  );
+  readonly dataSource = new WattTableDataSource<DhUserRolePermissionDetails>(undefined);
```
mimse, line 36: Suggestion: I think you can compute the datasource as well, and get rid of the effect. But maybe it well access the input to early.

mimse (changes requested):

libs/dh/admin/feature-user-roles/src/lib/drawer/tabs/content/dh-role-permissions.component.ts
dzhavat: Used `effect` here only because of the Material's recommendation to update the `.data` property with new data arrives, instead of creating a new data source.

mimse (approved):

PR #3489, opened 2024-08-30
feat(dh): preselect grid areas

ManBearTM:


libs/dh/shared/data-access-graphql/src/lib/graphql/get-grid-areas.graphql
```diff
@@ -6,5 +6,6 @@ query GetGridAreas {
     displayName
     validTo
     validFrom
+    includedInCalculation
```
dzhavat, line 9: nitpick: a bit odd to include a page specific property to a "global" query but I don't have a better suggestion right now.

dzhavat (approved):

PR #3491, opened 2024-08-30
Feature/behov for flere filtreringer p en gang

mimse:
## References

- #0000


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/Balance/BalanceResponsibilityMeteringPointType.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Energinet.DataHub.WebApi.GraphQL.Types.Balance;
+
+public class BalanceResponsibilityMeteringPointType : EnumType<ESett.MeteringPointType>
+{
+    protected override void Configure(IEnumTypeDescriptor<ESett.MeteringPointType> descriptor)
+    {
+        descriptor.Name("BalanceResponsibilityMeteringPointType");
+        descriptor.AsIsCase();
+    }
+}
```
dzhavat, line 27: issue: looks like this causes an issue with the metering point type translations in the overview table. The value is not translated anymore.

image

libs/dh/shared/data-access-mocks/src/lib/data/esett-balance-responsible-messages.ts
```diff
@@ -39,13 +39,13 @@ export const eSettBalanceResponsibleMessages = (apibase: string): BalanceRespons
       receivedDateTime: new Date('2021-02-01T00:00:00.000Z'),
       supplier: '123',
       balanceResponsible: '321',
-      gridArea: '344',
+      gridAreas: '344',
```
dzhavat, line 42: nitpick: this should probably stay `gridArea`

```suggestion
      gridArea: '344',
```

libs/dh/shared/data-access-mocks/src/lib/data/esett-balance-responsible-messages.ts
```diff
@@ -89,7 +89,7 @@ export const eSettBalanceResponsibleMessages = (apibase: string): BalanceRespons
       balanceResponsible: '999',
       gridArea: '000',
       gridAreaWithName: null,
-      meteringPointType: TimeSeriesType.Production,
+      meteringPointType: BalanceResponsibilityMeteringPointType.Production,
```
dzhavat, line 92: suggestion: let's remove the type casting on line 98 in order to get better type checking.

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/Balance/BalanceResponsibilityMeteringPointType.cs
mimse: Same translation was used for both timeseriestype and metering point time. Fixed now

PR #3498, opened 2024-09-03
feat(dh): graphql eslint

ManBearTM:
This PR adds linting rules for `.graphql` files (both operations AND schema, each with separate config), hopefully preventing some obscure bugs (such as the one outlined in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3494).

I have turned off some of the rules for now, since the PR would be way too big if I had to fix all the linting errors at once (yeah there are a few 😄). I plan to follow up with smaller PR's where I fix each lint error one at a time and then a final PR where I turn the rule(s) on.


mimse (approved):
Praise: nice change!

dzhavat (approved):
praise: great 🚀

PR #3512, opened 2024-09-09
feat(dh): put first and last name in one column

dzhavat:
### DataHub
- put first and last name in one column. The change is visual only. The back-end systems still process user name as "firstName" and "lastName". 
- refactor code to use Signals
- update mock data

\* A side-effect of this is that the user can only sort by `FirstName`. 

## References

- #0000


libs/dh/admin/feature-user-management/src/lib/overview/dh-users-overview-table.component.ts
```diff
@@ -94,40 +93,42 @@ export class DhUsersTabTableComponent implements AfterViewInit {
...
   ngAfterViewInit(): void {
-    this.usersTable.sortChange.pipe(takeUntilDestroyed(this._destroyRef)).subscribe((x) => {
-      const property = (x.active.charAt(0).toUpperCase() +
-        x.active.slice(1)) as UserOverviewSortProperty;
-      const direction = (x.direction.charAt(0).toUpperCase() +
-        x.direction.slice(1)) as MarketParticipantSortDirctionType;
-
-      this.sortChanged(property, direction);
-    });
+    this.usersTable()
```
mimse, line 116: Question: Could this be moved into an effect. When we can drop the destoryRef thing.

mimse (approved):

libs/dh/admin/feature-user-management/src/lib/overview/dh-users-overview-table.component.ts
dzhavat: Apparently not. Moving the logic inside an `effect` fails with 

image

PR #3517, opened 2024-09-10
feat(dh): apollo data source

### LINKED ISSUE #3192: refactor watt-paginator to use signals
Refactor ./watt/src/lib/components/paginator/watt-paginator.component.ts to use the new signal syntax

ManBearTM:
## Description

This PR introduces the new ApolloDataSource, which is a Material DataSource implementation that is meant to be used with the new auto-generated DataSource classes. When you add `UsePaging` and `UseSorting` to a list query in the BFF (and update the related `.graphql` operation file), a corresponding DataSource class will be generated in the `graphql.ts` file. For example, if you were to add paging and sorting to the `gridAreas` query, a new class named `GetGridAreasDataSource` will be available in `graphql.ts`. This can be used directly with `watt-table` to get a fully server-side paginated and sort enabled table. The DataSource also enables the `filter` function (search field in the table). Just add a string parameter named `filter` to the paginated query in the BFF.

The PR also implements this new functionality for the calculations table. Since that table is now fully server-side paginated, I had to make some changes to the subscription part. It ended up simplifying the logic quite a bit, with the only caveat that calculations that others create are not automatically added to the list (this might not even be a bad thing). Instead we use `refetchQueries` now when you create a new calculation.

## References

- Closes #3192


mimse (approved):
Praise: Looks, not understanding all of it :P

PR #3522, opened 2024-09-11
fix(dh): update dropdown translation on options change

ManBearTM:


libs/dh/shared/ui-util/src/lib/dh-dropdown-translator.directive.ts
```diff
@@ -31,26 +31,33 @@ export class DhDropdownTranslatorDirective implements OnInit {
...
+      }));
+
+      this.host.options = this.host.sortDirection
+        ? this.host.sortOptions(translatedOptions)
+        : translatedOptions;
+    });
+  }
+
   ngOnInit(): void {
     this.translocoService
```
mimse, line 57: Suggestion: If possible i would move this to the constructor as well, and remove the ngOnInit

ManBearTM: I tried that, but since it is using a required input signal, I got the usual error with "signal used before it is ready" thing

mimse: Aha, that one again !

PR #3526, opened 2024-09-12
fix(dh): allow signal writes on language change

ManBearTM:
- Add new combined enum for priceType and meteringPointType
- Update query to use new RequestCalculationDataType
- Update request calculation view to use new mutation
- Small improvement to mutation
- Update schema snapshot
- Allow signal writes in translation subscribers


mimse (dismissed):
Praise: Good comment

dzhavat (approved):

PR #3542, opened 2024-09-16
fix(dh): request calculation validation issues

ManBearTM:


libs/watt/src/lib/components/picker/datepicker/watt-datepicker.component.ts
```diff
@@ -205,6 +207,25 @@ export class WattDatepickerComponent extends WattPickerBase {
...
         this.rangePlaceholder = this.getRangePlaceholder();
       });
+
+    effect(() => {
+      this.rangeMonthOnlyMode();
+      this.ngControl?.control?.updateValueAndValidity();
+    });
+  }
+
+  override ngAfterViewInit() {
```
dzhavat, line 217: question: Isn't it dangerous to override this method? The one in the base class does something :)

libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.ts
```diff
@@ -244,23 +244,15 @@ export class DhWholesaleRequestCalculationComponent {
...
-          )
-        ),
+        tap(({ marketRole }) => {
+          if (marketRole !== EicFunction.GridAccessProvider) {
+            this.gridAreaIsRequired.set(false);
+            this.form.controls.gridArea.removeValidators(Validators.required);
+          } else {
+            this.gridAreaIsRequired.set(true);
+            this.form.controls.gridArea.setValidators(Validators.required);
+          }
```
dzhavat, line 254: question: no need for `updateValueAndValidity` any longer?

libs/watt/src/lib/components/picker/datepicker/watt-datepicker.component.ts
```diff
@@ -205,6 +207,25 @@ export class WattDatepickerComponent extends WattPickerBase {
...
+  }
+
+  validate({ value }: AbstractControl<WattRange<string>>) {
+    if (!value?.end || !value?.start) return null;
+    if (!this.rangeMonthOnlyMode()) return null;
+    const start = dayjs(value.start);
+    const end = dayjs(value.end);
+    return start.isSame(start.startOf('month')) && end.isSame(start.endOf('month'))
+      ? null
+      : { monthOnly: true };
```
dzhavat, line 228: issue: looks like this causes some issues. The most obvious I can see is the required indicator next to the label.

Before

image

After

image

I think the `null` return at the start of the method overwrites the other validation errors

libs/watt/src/lib/components/picker/datepicker/watt-datepicker.component.ts
ManBearTM: Yes, what was I thinking!? So that was why it was complaining about a missing override keyword 😄 It is so rare to extend classes these days, so totally missed that. Good catch!

973d58b28330f38c02de6df2729caac4ae2b166e

libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.ts
ManBearTM: Doesn't seem like it is needed any more, since changing actors reloads the entire view. Before it was needed since the validators would be updated when selecting different options in the dropdown.

libs/watt/src/lib/components/picker/datepicker/watt-datepicker.component.ts
ManBearTM: Thanks for trying it out, well spotted! I think the `null` return is fine, since that is the way for validators to tell that there is no error. I traced the issue to be with the `setValidators` call in `ngAfterViewInit`. I initially thought that the `setValidators` would be called before other validators are set, but that was wrong, since the form + validators are always instantiated in the parent component before child components render. I fixed it be changing it to a "addValidators" call instead.

0ab3309b636065fff7f13b5f57507984f017956e

libs/dh/wholesale/feature-request-calculation/src/lib/dh-wholesale-request-calculation.ts
dzhavat: My understanding was that `updateValueAndValidity` needs to be called every time a validator is added/removed at run time. The documentation also mentions that. But trying the code locally it seems like everything works fine. Let's go without it and keep it in mind in case some weird behaviour is observed. 

https://angular.dev/api/forms/AbstractControl#addValidators

image

dzhavat (approved):
praise: looks and works much better :)

PR #3549, opened 2024-09-17
feat(dh): show checkbox to select usage of new API for requesting settlement reports

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/wholesale/feature-settlement-reports/src/lib/modal/dh-request-settlement-report-modal.component.html
+++ b/libs/dh/wholesale/feature-settlement-reports/src/lib/modal/dh-request-settlement-report-modal.component.html
@@ -95,11 +95,12 @@ limitations under the License.
         </div>
       }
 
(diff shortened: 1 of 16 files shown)


libs/dh/wholesale/feature-settlement-reports/src/lib/modal/dh-request-settlement-report-modal.component.html
```diff
@@ -94,6 +94,12 @@
           }
         </div>
       }
+
+      <ng-container *dhFeatureFlag="'settlement-report-use-api'">
```
mimse, line 98: Question: Why not use the directive on checkbox component?

mimse (approved):

libs/dh/wholesale/feature-settlement-reports/src/lib/modal/dh-request-settlement-report-modal.component.html
dzhavat: Even better. Fixed :)

PR #3563, opened 2024-09-19
feat(dh): allow grid access provider to select energy supplier

dzhavat:
## References

- #0000


dzhavat:
note: no longer relevant

PR #3564, opened 2024-09-19
fix(dh): split graphql generated file

ManBearTM:
This PR fixes a very mysterious bug with what I assume is a problem with the bundling/code splitting. When the code related to the new DataSource functionality was included in `graphql.ts`, I would get weird errors in message archive (only there, not in the other domains) when importing from `graphql.ts` (some problem with RxJS not being properly packaged and resulting in weird "observable not a stream" kind of errors). By splitting the file in two, the errors went away. I'm still not sure why, but I think the solution is a good change anyway.


libs/dh/shared/feature-graphql-codegen/dist/index.js
```diff
@@ -14,7 +17,4 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-'use strict';
-Object.defineProperty(exports, '__esModule', { value: true });
-const tslib_1 = require('tslib');
-tslib_1.__exportStar(require('./apollo-data-source'), exports);
```
dzhavat, line 20: nitpick: the license header can be removed. Strange why it wasn't removed automatically when the file was recreated but anyway.

dzhavat (approved):
praise: nice work.

The error in message archive is probably because something imports something that imports something else and message archive is the first page after login. But nice that it's fixed now. 🚀

libs/dh/shared/feature-graphql-codegen/dist/index.js
ManBearTM: Yeah I noticed that, it seems to be the `tsc` command that doesn't remove them. Not sure why it keeps that comment when it removes it from the other file 🤔 Anyway, I didn't think it was important enough to bother fixing, I'd rather just keep the default output of `tsc`

PR #3585, opened 2024-09-27
refactor(dh): use Signals in components

dzhavat:
## References

- #0000


libs/dh/admin/feature-permissions/src/lib/details/tabs/dh-admin-permission-market-roles.component.html
```diff
@@ -14,63 +14,57 @@
...
-        } @else if (marketRoleCount !== 1 && !error) {
-          <h4>
-            {{ t("rolesPlural", { marketRoleCount }) }}
-          </h4>
-        }
+<watt-card
+  variant="solid"
+  *transloco="let t; read: 'admin.userManagement.permissionDetail.tabs.marketRoles'"
+>
+  @if (isLoading()) {
```
mimse, line 21: Nitpick: maketRoleCount and hasError is used quite abit, maybe use a @let

libs/dh/admin/feature-permissions/src/lib/details/tabs/dh-admin-permission-roles.component.html
```diff
@@ -14,68 +14,62 @@
...
-          </h4>
-        }
+<watt-card
+  variant="solid"
+  *transloco="let t; read: 'admin.userManagement.permissionDetail.tabs.userRoles'"
+>
+  @if (loading()) {
+    <div class="spinner">
+      <watt-spinner />
+    </div>
```
mimse, line 24: Nitpick: @let ?

libs/dh/admin/shared/src/dh-permissions-table.component.ts
```diff
@@ -87,12 +83,11 @@ export class DhPermissionsTableComponent implements OnChanges {
...
-  ngOnChanges() {
-    this.dataSource.data = this.permissions;
-    if (this.permissionsTable) this.permissionsTable.clearSelection();
-  }
+  ngOnChanges(): void {
+    this.dataSource.data = this.permissions();
 
-  onSelectionChange(selections: PermissionDetailsDto[]): void {
-    this.selectionChanged.emit(selections);
+    if (this.permissionsTable) {
```
mimse, line 89: Issue: permissionTable will always be something, because its a signal

const table = this.permissionTable();
if(table) ...

Or maybe make its a viewChild.required

mimse (changes requested):
Praise: Cools so clean now

libs/dh/admin/feature-permissions/src/lib/details/tabs/dh-admin-permission-roles.component.html
dzhavat: I considered it too bug didn't see a big difference since it will only save us a couple of characters - (). In terms of perf, calling the same Signal multiple times shouldn't be an issue.

libs/dh/admin/shared/src/dh-permissions-table.component.ts
dzhavat: I tried with `viewChild.required` but got a run-time error because the table couldn't be found. It's probably something related to `ngOnChanges` running too early. It's worth looking at this component again on its own.

mimse: Ok, but dont you still need to 

const table = this.permissionTable();
if(table) ... or you are checking if the signal is there and not what the signal returns

PR #3592, opened 2024-10-01
feat(dh): add include related filter to message archive

ManBearTM:


libs/dh/message-archive/feature-search/src/lib/table.component.ts
```diff
@@ -92,10 +111,13 @@ import { ArchivedMessage } from '@energinet-datahub/dh/message-archive/domain';
...
   `,
 })
-export class DhMessageArchiveSearchTableComponent {
+export class DhMessageArchiveSearchTableComponent implements OnInit {
   new = output();
   open = output<ArchivedMessage>();
   selection = signal<ArchivedMessage | undefined>(undefined);
+  dataTable = viewChild.required(WattDataTableComponent);
+  form = new FormGroup({ includeRelated: new FormControl(false) });
+  filterChanges = this.form.valueChanges.pipe(takeUntilDestroyed());
```
mimse, line 120: Question: Can toSignal be used?

ManBearTM: Yes, but then I had to add an effect like the example below. Is that more readable to you? I'm not sure myself, so I went with the classic `ngOnInit` approach.
```ts
filters = toSignal(this.form.valueChanges);
filterEffect = effect(() => {
  const filters = this.filters();
  if (!filters) return;
  this.dataSource.refetch(filters);
});
```

mimse: I dont know if its more readable. Just like getting rid of takeUntilDestoryed and life cycle hooks :-)

mimse (dismissed):

dzhavat (approved):

mimse (approved):

PR #3593, opened 2024-10-01
get user roles based on actor id when not admin

vbelinschi:
## References

- #0000


libs/dh/admin/feature-user-roles/src/lib/overview/dh-roles-overview.component.ts
```diff
@@ -102,8 +106,9 @@ type Filters = {
 export class DhUserRolesOverviewComponent {
   private readonly destroyRef = inject(DestroyRef);
   private readonly transloco = inject(TranslocoService);
+  private readonly actorStorage = inject(DhActorStorage);
 
-  private rolesGraphQL = query(GetUserRolesDocument);
+  private rolesByEicFunctionIdGraphQL = lazyQuery(GetUserRolesByEicfunctionDocument);
```
dzhavat, line 111: praise: the changes look good and the feature works quite well!

suggestion: I think we should keep using the `GetUserRolesDocument` and use the user's token in the BFF to figure out the user's role. The `GetUserRolesDocument` is used in other places too, e.g. in the "User roles" dropdown under the "Users" tab, and it's also called when a role is created/edited to refresh the table. 

image

So by using `GetUserRolesDocument` we'll actually solve the same issue in the "User roles" dropdown. What do you think?

vbelinschi: another nice input 
I updated again, third time's the charm :D

defectiveAi (approved):

PR #3594, opened 2024-10-01
Send actornumber instead of actorId

FirestarJes:


dzhavat (approved):

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Mutation/SettlementReportMutation.cs
```diff
@@ -22,8 +23,13 @@ public partial class Mutation
 {
     public async Task<bool> RequestSettlementReportAsync(
         RequestSettlementReportInput requestSettlementReportInput,
+        [Service] IMarketParticipantClient_V1 marketPartClient,
         [Service] ISettlementReportsClient client)
     {
+        var requestAsActor = string.IsNullOrEmpty(requestSettlementReportInput.RequestAsActorId)
```
michaelmajgaard, line 29: Use `Guid.TryParse`

michaelmajgaard, line 29: Instead of `string.IsNull..` you could use `Guid.TryParse(..)`

michaelmajgaard: Also, indentation seems of :D

FirestarJes: Fixed

michaelmajgaard (approved):

PR #3596, opened 2024-10-02
feat: use pagination api of search archived messages

MadsDue:
## Description

We want to avoid loading all archived messages into memory when searching. This pull request introduces pagination for archived messages, allowing users to navigate forwards and backwards through the dataset without loading more data than necessary into memory. It also supports sorting, enabling users to choose a specific field to sort the dataset by.

EDI pr: 
- https://github.com/Energinet-DataHub/opengeh-edi/pull/1260

## References

- #0000


apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/MessageArchive/ArchivedMessageSortInput.cs
```diff
@@ -0,0 +1,24 @@
...
+using Energinet.DataHub.WebApi.GraphQL.Enums;
+
+namespace Energinet.DataHub.WebApi.GraphQL.Types.MessageArchive;
+
+public record ArchivedMessageSortInput(
+    SortDirection? MessageId,
+    SortDirection? DocumentType,
+    SortDirection? Sender,
+    SortDirection? Receiver,
+    SortDirection? CreatedAt);
```
MadsDue, line 24: I think we can drop some code, if this look like this instead

```suggestion
public record ArchivedMessageSortInput(
    SortDirection? SortDirection,
    FieldToSortBy? FieldToSortBy);
```

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Query/MessageArchiveQuery.cs
```diff
@@ -56,6 +62,93 @@ public async Task<IEnumerable<ArchivedMessageResult>> GetArchivedMessagesAsync(
...
-        return await client.ArchivedMessageSearchAsync("1.0", search);
+        var (field, direction) = order switch
+        {
+            { MessageId: not null } => (FieldToSortBy.MessageId, order.MessageId),
+            { DocumentType: not null } => (FieldToSortBy.DocumentType, order.DocumentType),
+            { Sender: not null } => (FieldToSortBy.SenderNumber, order.Sender),
+            { Receiver: not null } => (FieldToSortBy.ReceiverNumber, order.Receiver),
+            { CreatedAt: not null } => (FieldToSortBy.CreatedAt, order.CreatedAt),
+            _ => (FieldToSortBy.CreatedAt, SortDirection.Desc),
+        };
```
MadsDue, line 73: if you update ArchivedMessageSortInput as commented. Then this should not be needed any more

MadsDue (commented):
Should we have et behind of a feature flag or how should this be handled in the frontend?

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Types/MessageArchive/ArchivedMessageSortInput.cs
ManBearTM: Yes I agree, that would be a simpler solution 👍 But there is a reason I'm not doing that. Long story short, the structure of this SortInput class follows the same convention as Hot Chocolate when you use the `UseSorting` attribute (https://chillicream.com/docs/hotchocolate/v13/fetching-data/sorting). This is because, in the frontend, we have code that can automatically add server-side paging with sorting to any table, provided that the underlying query follows the same convention. In other words, if you were to add `UsePaging` and `UseSorting` to any query that returns a list, we have a setup that automatically generates a class for the frontend to use in a table component ([this is all it takes to make a table server-side paginated with sorting](https://github.com/Energinet-DataHub/greenforce-frontend/blob/daa5f98ad9e59a9e8522d0493c4035799a6e3bac/libs/dh/message-archive/feature-search/src/lib/table.component.ts#L130-L136)). AFAIK there is no way to alter the sort input type convention of `UseSorting`, so I'd rather just go with that over having to support multiple conventions 🙂

ManBearTM (approved):

PR #3608, opened 2024-10-07
EO: Power of attorney

frontend-specialisten:
## Description

Add functionality for changing actor (mocked data only)

## References

- #0000


sahma19 (approved):

sahma19 (approved):

libs/watt/src/lib/foundations/typography/_typography.import.scss
```diff
@@ -49,6 +49,10 @@ $font-sizes: (
     "font-size": 14px,
     "line-height": 20px,
   ),
+  "xs": (
+    "font-size": 12px,
+    "line-height": 22px,
```
TopSwagCode, line 54: line-height seems abit of compared to other font sizez. It seems to get smaller each time except for the last time? This time we increase by 2 ?

libs/eo/auth/data-access/actor.service.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { computed, inject, Injectable, signal } from '@angular/core';
+
+import { Actor } from '@energinet-datahub/eo/auth/domain';
```
TopSwagCode, line 19: To beeeee, or not to be. That is the question.

libs/watt/src/lib/foundations/typography/_typography.import.scss
frontend-specialisten: I know, however the font-sizes are taken from Pernille's figma so alternatives are only using it for this use-case or not follow the sizes she has used.

libs/eo/auth/domain/index.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @license
```
TopSwagCode, line 2: Do we really need all this ? Seems like waste of E-Ink :P

libs/eo/consent/data-access-api/consent.service.ts
```diff
@@ -31,6 +31,12 @@ export interface EoConsent {
   idpClientId: string;
 }
 
+export interface EoReceivedConsent {
```
TopSwagCode, line 34: Eo is so yesterday.

libs/eo/auth/domain/index.ts
frontend-specialisten: I'm not sure if we still have too, however from the beginning we had and it's added automatically by our CI pipeline.

libs/eo/consent/data-access-api/consent.service.ts
frontend-specialisten: True dat, however the rename has not been made so they are still named eo to avoid usage of both eo and ett

TopSwagCode (dismissed):

TopSwagCode (approved):

PR #3618, opened 2024-10-10
feat(dh): message archive improvements

ManBearTM:
## Description

This PR fixes the following things in message archive:

- Show correct actor name in message archive result (by using gln AND role when looking up name)
- Likewise, it fixes searching for sender/receiver to use both gln and role in search criteria
- Only shows relevant document types and business reasons
- Updates translations
- Remove old controller logic that was unused


apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
```diff
@@ -527,7 +529,7 @@ type Query {
...
   imbalancePricesForMonth(year: Int! month: Int! areaCode: PriceAreaCode!): [ImbalancePriceDaily!]!
-  archivedMessages(created: DateRange! senderNumber: String receiverNumber: String documentTypes: [DocumentType!] businessReasons: [BusinessReason!] includeRelated: Boolean filter: String "Returns the first _n_ elements from the list." first: Int "Returns the elements in the list that come after the specified cursor." after: String "Returns the last _n_ elements from the list." last: Int "Returns the elements in the list that come before the specified cursor." before: String order: ArchivedMessageSortInput): ArchivedMessagesConnection
+  archivedMessages(created: DateRange! senderId: UUID receiverId: UUID documentTypes: [DocumentType!] businessReasons: [BusinessReason!] includeRelated: Boolean filter: String "Returns the first _n_ elements from the list." first: Int "Returns the elements in the list that come after the specified cursor." after: String "Returns the last _n_ elements from the list." last: Int "Returns the elements in the list that come before the specified cursor." before: String order: ArchivedMessageSortInput): ArchivedMessagesConnection
```
RasmusGraabaek, line 532: senderId and receiverId, arn't they just a numbers of length 13 or 16 ?

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -202,126 +202,20 @@
...
-      "NOTIFY_BILLING_MASTER_DATA": "NotifyBillingMasterData (RSM-031)",
-      "REJECT_REQUEST_BILLING_MASTER_DATA": "RejectRequestBillingMasterData (RSM-032)",
-      "REQUEST_BILLING_MASTER_DATA": "RequestBillingMasterData (RSM-032)",
-      "CONFIRM_REQUEST_CHANGE_OF_PRICELIST": "ConfirmRequestChangeOfPricelist (RSM-033)",
-      "REJECT_REQUEST_CHANGE_OF_PRICELIST": "RejectRequestChangeOfPricelist (RSM-033)",
-      "REQUEST_CHANGE_OF_PRICELIST": "RequestChangeOfPricelist (RSM-033)",
-      "NOTIFY_PRICELIST": "NotifyPricelist (RSM-034)",
-      "REJECT_REQUEST_PRICELIST": "RejectRequestPricelist (RSM-035)",
-      "REQUEST_PRICELIST": "RequestPricelist (RSM-035)"
+      "NOTIFY_AGGREGATED_MEASURE_DATA": "Notifikation om aggregerede tidsserier (RSM-014)",
```
RasmusGraabaek, line 205: nit: can we have to same sortin?
that is, our message has the prefixes (NOTIFY,...) in the same order :)

RasmusGraabaek: If you agree to this, give "libs/dh/globalization/assets-localization/src/assets/i18n/en.json" a look :)

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Query/MessageArchiveQuery.cs
```diff
@@ -37,26 +40,36 @@ public async Task<Connection<ArchivedMessageResultV2>> GetArchivedMessagesAsync(
...
+        async Task<ActorDto?> GetActorAsync(Guid? id)
+        {
+            return id.HasValue
+                ? await markPartClient.ActorGetAsync(id.Value)
+                : null;
+        }
+
+        var sender = GetActorAsync(senderId);
+        var receiver = GetActorAsync(receiverId);
         var searchCriteria = !string.IsNullOrWhiteSpace(filter)
```
RasmusGraabaek, line 55: So the input "filter" is just the messageId of the wanted message? Of so. Can we give it a more precise name?

ManBearTM: Good point, but unfortunately not. It has to be named `filter` due to the way the table is implemented in the frontend. When you use the search field in the table view, it automatically updates the query to filter by that value - and thus it expects the query to have a parameter named `filter`. This is of course just a convenience thing so that you do not have to set up that binding for every table.

We have an ongoing discussion in the frontend guild regarding these search fields, whether they should only be free text filters or something else. So it may change in the future.

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
ManBearTM: Good idea to align sorting 🙂 Question is, do we actually want alphabetical sorting in the UI dropdown? It used to be sorted by "RSM" number, which is also how it is sorted in the translation files (this has no effect on the UI though).

ManBearTM: Meyer says to go with RSM number ordering for now. So I might take a different approach here, keep the order as it is in the translation files and then let the translation files dictate the order. Alternatively we could change the order in the backend client, or even in the BFF, but I'm not sure if that's any better.

apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
ManBearTM: They used to be gln numbers, now they are GUID's:

https://github.com/Energinet-DataHub/greenforce-frontend/pull/3618/files#diff-9417ffc0e16aac51c332b0d2d3a888f4db2a2a76c5273691bfed95901c80769fR134

RasmusGraabaek (approved):

PR #3621, opened 2024-10-11
feat(dh): add copy to clipboard to user roles tab

dzhavat:
### DataHub
- add copy to clipboard icon to expandable card under user roles tab
- fix badge in expandable card

image

## References

- #0000


michaelmajgaard (approved):
LGTM

PR #3622, opened 2024-10-11
feat(dh): util apollo subscription

ManBearTM:


dzhavat (approved):
praise: amazing! Works quite well with the new messages being pushed from the BFF. These notifications are going to be a great feature.

PR #3634, opened 2024-10-15
Feat/ett 3317 align footers

frontend-specialisten:
## Description

- Refactor: remove the atomic-design structure for the shared components (pattern is not used anywhere else, so it does not make sense to use it)
- Refactor: clean-up unused shared components
- Refactor: Move the landing page footer to a shared component
- Chore: Allow UI libraries to depend on environment libraries
- Feat: Use the shared footer component on both landing-page and the portal (after login)
- Feat: "Hide" the footer by having a min-height of 90vh of the content (after login)

## References

- [#3317](https://github.com/Energinet-DataHub/energy-origin-issues/issues/3317)


libs/eo/shared/styles/src/lib/@energinet-datahub/eo/shared/styles/themes/_themes-watt.scss
```diff
@@ -30,3 +30,13 @@ a,
 [routerLink] {
   text-decoration: underline; // This overrides WATT styles (which removes the underline on a tags)
 }
+
+// Layout
+.watt-toolbar {
+  width: 100% !important;
```
TopSwagCode, line 36: !important; I hate important :D It can't be that important! :P

TopSwagCode (approved):

PR #3658, opened 2024-10-23
Upgrade Hot Choco.

defectiveAi:


apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
```diff
@@ -9,22 +9,22 @@ interface Error {
...
+  displayName: String! @cost(weight: "10")
+  marketRole: EicFunction @cost(weight: "10")
+  userRoles: [ActorUserRole!]! @cost(weight: "10")
+  status: ActorStatus! @cost(weight: "10")
+  gridAreas: [GridAreaDto!]! @cost(weight: "10")
+  contact: ActorContactDto @cost(weight: "10")
+  organization: Organization! @cost(weight: "10")
+  balanceResponsibleAgreements: [BalanceResponsibilityAgreement!] @cost(weight: "10")
+  credentials: ActorCredentialsDto @cost(weight: "10")
+  publicMail: ActorPublicMail @cost(weight: "10")
```
ManBearTM, line 23: The operation cost can be a cool feature, but I don't really think we need it at the moment. Maybe consider disabling it for now so that it doesn't clutter up the schema unneccesarily? Looking at the docs it seems like you can disable it like this:

```csharp
services
    .AddGraphQL()
    .ModifyRequestOptions(o =>
    {
        o.Complexity.Enable = false;
    });
```

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Resolvers/MarketParticipantResolvers.cs
```diff
@@ -50,7 +50,7 @@ public async Task<IEnumerable<GridAreaDto>> GetGridAreasAsync(
                 .Distinct()
                 .Select(async gridAreaId => await dataLoader.LoadAsync(gridAreaId)));
 
-        return gridAreas.OrderBy(g => g.Code);
+        return gridAreas.Select(gridArea => gridArea!).OrderBy(g => g.Code);
```
ManBearTM, line 53: Why do we need the null-forgiving operator here? Those always seem like code smell to me.

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Resolvers/WholesaleResolvers.cs
```diff
@@ -24,14 +24,14 @@ public class WholesaleResolvers
...
-        (await dataLoader.LoadAsync(batch.CreatedByUserId)).DisplayName;
+        (await dataLoader.LoadAsync(batch.CreatedByUserId))?.DisplayName;
 
     public async Task<IEnumerable<GridAreaDto>> GetGridAreasAsync(
         [Parent] CalculationDto batch,
         GridAreaByCodeBatchDataLoader dataLoader)
     {
         var gridAreas = await Task.WhenAll(batch.GridAreaCodes.Select(async c => await dataLoader.LoadAsync(c)));
-        return gridAreas.Where(g => g != null).OrderBy(g => g.Code);
+        return gridAreas.Where(g => g != null).Select(g => g!).OrderBy(g => g.Code);
```
ManBearTM, line 34: Also here, although the preceding `Where` prevents runtime errors at least. Does the compiler not infer the grid area to be non-null after the `Where` operator?

defectiveAi: Fixed.

defectiveAi: And no, Where() does not change nullability, so compiler will not see the change.

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/Resolvers/MarketParticipantResolvers.cs
defectiveAi: Fixed.

apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
defectiveAi: Have no luck disabling the feature. There does not seem to be an option in v14 that works; docs are useless.

defectiveAi: Worst setting name ever!

ManBearTM:
LGTM!

ManBearTM (approved):

PR #3662, opened 2024-10-24
feat(dh): message archive filters

ManBearTM:


libs/dh/message-archive/feature-search/src/lib/filters.component.ts
```diff
@@ -0,0 +1,136 @@
...
+})
+export class DhMessageArchiveSearchFiltersComponent {
+  isSearchingById = input(false);
+  filter = output<GetArchivedMessagesQueryVariables>();
+  clear = output();
+
+  form = inject(DhMessageArchiveSearchFormService);
+
+  getDocumentTypeTrigger = (value: string | string[]) =>
+    this.form.getDocumentTypeIdentifier(value[0] as DocumentType);
```
mimse, line 126: Issue, if this is a string getDocumentTypeIdentifier will be set to the first letter of the string?

mimse (changes requested):

libs/dh/message-archive/feature-search/src/lib/filters.component.ts
ManBearTM: True, but that is only if the dropdown is not in `multiple` selection mode, which it always is. So this type is just to satisfy the compiler, `value` will always be an array. But I agree that it looks like a bug - I will add a comment.

ManBearTM: bfc2ba026497dbe4cd8936fb76419f47b0b13169

mimse (approved):

PR #3686, opened 2024-10-31
Refactor/improvements for organization on actor

mimse:


libs/dh/market-participant/actors/feature-organizations/src/lib/dh-organizations-overview.component.html
```diff
@@ -40,4 +40,6 @@ <h3>{{ t("organizations") }}</h3>
   </vater-flex>
 </watt-card>
 
-<dh-organization-drawer [id]="id()" (closed)="navigate(null)" />
+<dh-organization-details [id]="id()" (closed)="navigate(undefined)" />
```
dzhavat, line 43: issue: since the component is related to the `details/:id` a route, it shoudn't be included in here too. Angular will take care of instantiating it. That also means that the `id` should be a property of the component (which already is), where it's value is taken from the route params. The `closed` event should be handled in the component instead of emitting it.

mimse: Good find, it also fixed the issue with the edit modal not getting closed ! :-)

libs/dh/market-participant/actors/feature-organizations/src/lib/dh-organizations-overview.component.ts
```diff
@@ -99,11 +98,10 @@ export class DhOrganizationsOverviewComponent {
...
 
-  navigate(id: string | null) {
-    this.router.navigate([], {
+  navigate(id: string | undefined) {
+    this.id.set(id);
+    this.router.navigate(['details', id], {
       relativeTo: this.route,
-      queryParams: { id },
-      queryParamsHandling: 'merge',
     });
```
dzhavat, line 105: suggestion: a similar method that handles the closing of the drawer should be inside the `dh-organization-details.component.ts` as well. Right now the url doesn't update when the drawer is closed.
Then this class should also look at the route params in order to set the correct row as active, and clearing it when the drawer closes.

dzhavat (approved):

PR #3691, opened 2024-11-01
feat(dh): download settlement report from notification

dzhavat:
## References

- #0000


libs/dh/core/feature-notifications/src/lib/dh-notifications-center.component.ts
```diff
@@ -207,6 +201,12 @@ export class DhNotificationsCenterComponent {
     });
   }
 
+  onActionButtonClicked({ notificationType, relatedToId }: DhNotification): void {
+    if (notificationType === NotificationType.SettlementReportReadyForDownload && relatedToId) {
+      this.notificationsService.downloadSettlementReport(relatedToId);
```
mimse, line 206: Suggestion: it seems wrong to me that noticication center should know about settlement reports. If this component should be generic for all of our notifications. I think we need to move this logic out some how

mimse (changes requested):

mimse (approved):

PR #3708, opened 2024-11-07
feat: Use Process Manager from BFF

dstenroejl:
## Description

- [x] Implement feature flag for the Process Manager feature
- [x] Use Process Manager for managing Calculations (start, cancel, query)
- [x] Add health checks for Process Manager API's

## References

https://app.zenhub.com/workspaces/product-goal-board-6375df2fd6f08e0015e1e0e6/issues/gh/energinet-datahub/team-mosaic/346


apps/dh/api-dh/source/DataHub.WebApi/DataHub.WebApi.csproj
```diff
@@ -53,6 +53,7 @@ limitations under the License.
 
   <ItemGroup>
     <PackageReference Include="Azure.Monitor.OpenTelemetry.AspNetCore" Version="1.2.0" />
+    <PackageReference Include="Energinet.DataHub.ProcessManager.Client" Version="0.9.1-alpha-7" />
```
dstenroejl, line 56: Use release when ready

dstenroejl: Fixed

apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/ProcessManager/NotifyAggregatedMeasureDataClientAdapter.cs
```diff
@@ -0,0 +1,84 @@
...
+        CalculationQueryInput input,
+        CancellationToken calculationToken = default)
+    {
+        // TODO:
+        // We currently only support filtering on one state initially, and only
+        // top-level states (orchestration instance states, not steps)
+        var states = input.States ?? [];
+        var state = states
+            .Select(x => x.MapToLifecycleState())
+            .FirstOrDefault();
```
ManBearTM, line 40: question: is the plan to then support multiple states in the `SearchCalculationsAsync`? 🙂

ManBearTM: Okay thanks! Should we implement it "correctly" here at this time then? The way I understand this code, is that if you filter for multiple states, you would actually only get results for the first state. Wouldn't it be more correct to search for all states and then filter the result instead (as is already happening here: https://github.com/Energinet-DataHub/greenforce-frontend/pull/3708/files#diff-f27e9c5d48c523aceeac4ccbf4c80aa206ed6afa5a065f9dfadf573923fd4061R70)

apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/ProcessManager/OrchestrationInstanceMapperExtensions.cs
```diff
@@ -0,0 +1,185 @@
...
+    public static CalculationTypes MapToCalculationType(
+        this v3.CalculationType calculationType)
+    {
+        return Enum
+            .TryParse<CalculationTypes>(
+                calculationType.ToString(),
+                ignoreCase: true,
+                out var calculationTypeResult)
+            ? calculationTypeResult
+            : throw new InvalidOperationException($"Invalid CalculationType '{calculationType}'; cannot be mapped.");
```
ManBearTM, line 72: suggestion: i know this is temporary code, but is it possible to use an explicit mapping here so that we can benefit from the compilers exhaustiveness checks?

apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/ProcessManager/OrchestrationInstanceMapperExtensions.cs
```diff
@@ -0,0 +1,185 @@
...
+    public static v3.CalculationType MapToV3CalculationType(
+        this CalculationTypes calculationType)
+    {
+        return Enum
+            .TryParse<v3.CalculationType>(
+                calculationType.ToString(),
+                ignoreCase: true,
+                out var calculationTypeResult)
+            ? calculationTypeResult
+            : throw new InvalidOperationException($"Invalid CalculationType '{calculationType}'; cannot be mapped.");
```
ManBearTM, line 118: suggestion: same here

apps/dh/api-dh/source/DataHub.WebApi/Program.cs
```diff
@@ -85,6 +88,9 @@
 var apiClientSettings = configuration.GetSection("ApiClientSettings").Get<ApiClientSettings>() ?? new ApiClientSettings();
 services.AddDomainClients(apiClientSettings);
 
+services.AddFeatureManagement();
+services.AddProcessManager(configuration);
```
ManBearTM, line 92: nitpick: is there a reason this call is placed here and not within the `AddDomainClients` extension method?

ManBearTM (commented):
LGTM, just a few comments 🙂

apps/dh/api-dh/source/DataHub.WebApi/Program.cs
dstenroejl: 1: The call to `AddFeatureManagement` should be registered independent of any other code, as we should support feature flags ins any .NET code, not just clients. So I believe it should be registered "in the root".
2: With regards to `AddProcessManager` we can move it. I kept it separate from `AddDomainClients` because I have to follow another pattern and wasn't sure if other developers would wan't me to mix the two coding patterns/styles.

apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/ProcessManager/NotifyAggregatedMeasureDataClientAdapter.cs
dstenroejl: Nice catch, I will change the filter values to `null` and write a TODO for later.

ManBearTM (approved):
🥇

MadsDue (approved):

PR #3742, opened 2024-11-15
refactor(dh): drop ComponentStore from service

dzhavat:
## References

- #0000


mimse (approved):
Not crazy about the store, maybe we could co locate some of the logic in the  components at some point. 
But other than that it looks much better than before

PR #3743, opened 2024-11-15
feat(dh): CollectionSegmentDataSource

ManBearTM:
## Description

Turns out that the Connection abstraction was not adequate for use with offset paging, due to issues with backwards pagination. To fix this, I have added support for the `UseOffsetPaging` attribute in Hot Chocolate. This also simplifies the implementation in the BFF, since you do not have to worry about edges and cursors.


mimse (approved):
Praise: Great job, you are the man 🥇

PR #3745, opened 2024-11-18
refactor(dh): use Signals in components

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/market-participant/actors/feature-actors/src/lib/create/steps/dh-new-actor-step.component.ts
+++ b/libs/dh/market-participant/actors/feature-actors/src/lib/create/steps/dh-new-actor-step.component.ts
@@ -55,7 +55,7 @@ import { dhMarketParticipantNameMaxLength } from '../../dh-market-participant-na
         display: block;
       }
 
-      .grid > .grid__item {
+      .container > .container__item {
         width: 50%;
       }
 
@@ -69,10 +69,10 @@ import { dhMarketParticipantNameMaxLength } from '../../dh-market-participant-na
     gap="xl"
     align="flex-start"
     direction="row"
-    class="grid"
+    class="container"
     *transloco="let t; read: 'marketParticipant.actor.create'"
   >
-    <vater-stack class="grid__item" fill="horizontal" align="flex-start" direction="column">
+    <vater-stack class="container__item" fill="horizontal" align="flex-start" direction="column">
       <h4>{{ t('marketParty') }}</h4>
 
       <watt-text-field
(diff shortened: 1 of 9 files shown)


libs/dh/market-participant/actors/feature-actors/src/lib/create/steps/dh-new-actor-step.component.ts
```diff
@@ -27,36 +27,36 @@ import {
...
   ],
   styles: [
     `
       :host {
         display: block;
       }
 
-      watt-dropdown {
-        width: 100%;
+      .grid > .grid__item {
```
mimse, line 58: Suggestion: Seems odd to me that we have a flexbox layout, but apply grid to it? Maybe vater should support grid layout as well?

libs/dh/market-participant/actors/feature-delegation/src/lib/create/dh-delegation-create-modal.component.ts
```diff
@@ -150,7 +149,7 @@ export class DhDelegationCreateModalComponent extends WattTypedModal<DhActorExte
 
     this.isSaving.set(true);
 
-    this._apollo
+    this.apollo
```
mimse, line 152: Suggestion: Use mutation here

libs/dh/market-participant/actors/feature-delegation/src/lib/create/dh-delegation-create-modal.component.ts
```diff
@@ -201,7 +200,7 @@ export class DhDelegationCreateModalComponent extends WattTypedModal<DhActorExte
       eicFunctions.push(EicFunction.GridAccessProvider);
     }
 
-    return this._apollo
+    return this.apollo
```
mimse, line 203: Suggestion: Use query here

libs/dh/market-participant/actors/feature-delegation/src/lib/create/dh-delegation-create-modal.component.ts
dzhavat: This component will be refactored in a follow-up PR.

libs/dh/market-participant/actors/feature-actors/src/lib/create/steps/dh-new-actor-step.component.ts
dzhavat: We are actually using the vater layout here, but apparently it's not enough. 
As per the screenshot below, the layout is divided into two columns. Selecting grid areas in the dropdown expands the first column because the dropdown doesn't have a specific width. The change above tries to address that. Even though the class is called `grid` there's no `grid` applied to the layout. It's just the name of the class. I can change it to something else if it causes confusion. 

image

mimse: another name might be a good idea, to avoid confusion :-)

mimse (approved):

mimse (approved):

PR #3756, opened 2024-11-19
Refactor/user management part 5

mimse:


libs/dh/admin/feature-user-management-new/src/lib/edit/edit.component.ts
```diff
@@ -127,32 +127,56 @@ export class DhEditUserComponent {
...
+            lastName,
+            phoneNumber: formattedPhoneNumber,
+          },
+        },
+      },
+    });
+
+    if (result.data?.updateUserIdentity.success && result.data?.updateUserRoleAssignment.success) {
+      this.showSuccessToast();
     }
```
dzhavat, line 175: suggestion: in some cases `onCompleted` and `onError` callbacks on the mutation are used instead. For consistency sake, should we stick to only one way?

dzhavat: I'm fine with both solutions.

dzhavat (approved):

PR #3761, opened 2024-11-20
feat(dh): watt-datetime-field

ManBearTM:


libs/watt/src/lib/components/field/watt-field.component.ts
```diff
@@ -90,8 +109,20 @@ export class WattFieldComponent {
   wrapper = viewChild.required<ElementRef>('wrapper');
 
   constructor() {
-    const status$ = toObservable(this.control).pipe(
-      filter((control) => control !== null),
+    const control$ = toObservable(this.control).pipe(filter((control) => control !== null));
+
+    const value$ = control$.pipe(
```
mimse, line 114: Suggestion: This might need a comment :-)

mimse (commented):
Praise: Think it looks great. Love the use of new browser functions !

libs/watt/src/lib/configuration/watt-locale.service.ts
```diff
@@ -14,18 +14,24 @@
...
+import { BehaviorSubject } from 'rxjs';
 import { dayjs } from '@energinet-datahub/watt/date';
+import { toSignal } from '@angular/core/rxjs-interop';
 
 @Injectable({
   providedIn: 'root',
 })
 export class WattLocaleService {
-  onLocaleChange$ = new Subject<WattSupportedLocales>();
+  private locale: WattSupportedLocales = inject(LOCALE_ID) as WattSupportedLocales;
```
dzhavat, line 29: nitpick:

```suggestion
  private locale = inject<WattSupportedLocales>(LOCALE_ID)
```

libs/watt/src/lib/configuration/watt-locale.service.ts
```diff
@@ -14,18 +14,24 @@
...
+
+  onLocaleChange$ = new BehaviorSubject(this.locale);
+  current = toSignal(this.onLocaleChange$, { requireSync: true });
```
dzhavat, line 32: suggestion: thinking about it, isn't this just a fancy `signal` with initial value? I wonder whether simply 👇 will do it 🤔

```ts
locale = signal(inject<WattSupportedLocales>(LOCALE_ID));
```

libs/watt/src/lib/components/datetime-field/watt-datetime-field.component.ts
```diff
@@ -0,0 +1,204 @@
...
+      this.calendar().activeDate = date ?? new Date();
+    });
+  }
+
+  // Implementation for ControlValueAccessor
+  writeValue = (value: Date | null) => this.control.setValue(this.modelToView(value));
+  setDisabledState = (x: boolean) => (x ? this.control.disable() : this.control.enable());
+  registerOnTouched = (fn: () => void) => this.blur.subscribe(fn);
+  registerOnChange = (fn: (value: Date | null) => void) => this.dateChange.subscribe(fn);
+}
```
dzhavat, line 204: praise: beautiful 🤩

libs/watt/src/lib/components/datetime-field/watt-datetime-field.component.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  /** Converts date from outer FormControl to format of inner FormControl. */
+  protected modelToView = (value: Date | null, format = DATETIME_FORMAT) =>
+    value ? dayjs(value).tz(DANISH_TIME_ZONE_IDENTIFIER).format(format) : '';
+
+  /** Converts value of inner FormControl to type of outer FormControl. */
+  protected viewToModel = (value: string) => {
+    const date = dayjs(value, DATETIME_FORMAT, true);
+    if (!date.isValid()) return null;
+    return this.inclusive() ? date.endOf('m').toDate() : date.toDate();
```
dzhavat, line 138: question: when `inclusive` is set, the initial control value should be end-of-day, right? `2024-11-20T00:00Z` will not be converted to ``2024-11-20T23:59:59.999Z``

libs/watt/src/lib/components/datetime-field/watt-datetime-field.component.ts
```diff
@@ -0,0 +1,204 @@
...
+import { maskitoDateTimeOptionsGenerator } from '@maskito/kit';
+import { map } from 'rxjs';
+import { dayjs } from '@energinet-datahub/watt/date';
+import { WattFieldComponent } from '../field';
+import { WattButtonComponent } from '../button/watt-button.component';
+import { outputFromObservable } from '@angular/core/rxjs-interop';
+import { WattLocaleService } from '@energinet-datahub/watt/locale';
+
+const DA_FILLER = 'dd-mm-åååå, tt:mm';
+const EN_FILLER = 'dd-mm-yyyy, hh:mm';
```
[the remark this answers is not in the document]
ManBearTM: Good point on being consistent, however I do not understand why the time part should be capitalized and the date not? That is just inconsistency within the filler itself 😄 I took inspiration from the maskito examples, which use all lowercase. Maybe we should just update the timepicker instead (which is about to become "obsolete" for the time being, since this component supersedes the last component that was using it).

libs/watt/src/lib/components/datetime-field/watt-datetime-field.component.ts
ManBearTM: No, only "end-of-minute". I attempted to describe the behavior on the input [here](https://github.com/Energinet-DataHub/greenforce-frontend/pull/3761/files/97d3677ce46f9b2465eadcad6d5b8cc05ffbdddb#diff-c4c57765c0164d8b665398b0901bdfb7ae5ef6fc011d7959263ce16637e3bdf8R157), but maybe I need to make it clearer 🤔 I need this behavior so you can use two datetime fields as an alternative to a "datetime-range field", as the UX of such a field would be a nightmare to implement 😄

ManBearTM: Your example of `2024-11-20T00:00Z` should result in `2024-11-20T00:00:59.999Z` if using `[inclusive]=true`. I don't think it immediately converts the value and I'm not sure that it should, seems like the callers responsibility IMO 🙂

libs/watt/src/lib/configuration/watt-locale.service.ts
ManBearTM: Good question, I would assume that works, although it looks a bit funky 😄 I'm not sure how you want me to use that signal though, can you show me the rest of the implementation?

libs/watt/src/lib/components/datetime-field/watt-datetime-field.component.ts
ManBearTM: Appreciated, it wasn't easy cooking it down to that 😄

libs/watt/src/lib/configuration/watt-locale.service.ts
dzhavat: Something like this

```ts
current = signal(inject<WattSupportedLocales>(LOCALE_ID));

isDanish = computed(() => this.current() == 'da');
isEnglish = computed(() => this.current() == 'en');

constructor(private dateAdapter: DateAdapter<unknown>) {}

async setActiveLocale(locale: WattSupportedLocales): Promise<void> {
  if (locale === 'da') {
    await import('dayjs/locale/da');
  }

  if (locale === 'en') {
    await import('dayjs/locale/en');
  }

  dayjs.locale(locale);
  (this.dateAdapter as WattDateAdapter).setActiveLocale(locale);
  this.current.set(locale);
}
```

Can see that `watt-datepicker.component.ts` will need to be adjusted slightly should we go with this.

libs/watt/src/lib/components/datetime-field/watt-datetime-field.component.ts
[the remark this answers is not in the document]
dzhavat: I think the "HH:MM" format comes partly from the formatting library, where `HH` means `00 01 ... 22 23` values and partly from "Why use capital HH with lower-case mm, let's have both capital"-reasoning. Because `MM` means something else when formatting dates. So I think the lower-case is more correct.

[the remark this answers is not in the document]
ManBearTM: Yeah, as developers we are used to the format strings, but no users are ever going to understand the difference between `dd-MM-yyyy, HH:mm` and `dd-mm-yyyy, hh:mm` 😄 It will just look off to them I think. We could also consider using all capitals instead (`DD-MM-YYYY, HH:MM`)? 🙂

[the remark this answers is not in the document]
dzhavat: I think lowercase is fine until someone comes with better arguments for doing something else. We can adjust other components as follow-up PRs.

libs/watt/src/lib/configuration/watt-locale.service.ts
ManBearTM: Good idea and it appears to work 🎉 

https://github.com/Energinet-DataHub/greenforce-frontend/pull/3761/commits/acab1188056030b06eed71590587a4655a828a16

dzhavat: Thinking about it, let's not introduce more changes just for the sake of it. We can make it in a follow-up PR if we still think this is a good idea.

dzhavat (approved):

PR #3762, opened 2024-11-20
Service Provider Terms

MJ1227UB:
## Description

Introduces the Service Provider Terms, that should be agreed before being able to request consent.

## Parts

- Creates new service for calling the new API
- Adapts the Request Consent component to contain service provider terms
- Validated and refactored all service to use new Dependency Injection


libs/eo/shared/components/ui-scroll-view/eo-scroll-view.component.ts
```diff
@@ -117,6 +117,7 @@ export class EoScrollViewComponent implements OnDestroy {
       if (entry.isIntersecting) {
         this.checkScroll();
         this.observeContentChanges();
+        console.log('NOOO!!!');
```
frontend-specialisten, line 120:
```suggestion
```

frontend-specialisten (changes requested):

frontend-specialisten (approved):

PR #3782, opened 2024-11-26
refactor(dh): use default export with routes

dzhavat:
## References

- #0000


mimse (approved):
Praise: More simple i like

PR #3785, opened 2024-11-27
Made hasError property on query, and change the code to use that

mimse:


libs/dh/esett/data-access-outgoing-messages/src/lib/dh-outgoing-messages.store.ts
```diff
@@ -112,7 +112,7 @@ export const DhOutgoingMessagesSignalStore = signalStore(
     ),
     loading: outgoingMessageQuery().loading,
     downloading: downloadDocumentQuery().loading,
-    hasError: computed(() => outgoingMessageQuery().error() !== undefined),
+    hasError: computed(() => outgoingMessageQuery().hasError()),
```
dzhavat, line 115:
```suggestion
    hasError: outgoingMessageQuery().hasError,
```

libs/dh/imbalance-prices/shell/src/lib/dh-imbalance-prices-shell.component.html
```diff
@@ -41,7 +41,7 @@ <h3>{{ t("topBarTitle") }}</h3>
     <dh-table-month-view
       [tableDataSource]="tableDataSource"
       [isLoading]="isLoading()"
-      [hasError]="hasError"
+      [hasError]="hasError()"
```
dzhavat, line 44: praise: fixing bugs along the way too :)

libs/dh/shared/util-apollo/src/lib/query.ts
```diff
@@ -154,6 +155,7 @@ export function query<TResult, TVariables extends OperationVariables>(
   // Signals holding the result values
   const data = signal<TResult | undefined>(initial.data);
   const error = signal<ApolloError | undefined>(initial.error);
+  const hasError = signal(false);
```
dzhavat, line 158: question: since this property is derived from the value of `error`, maybe we can make it a `computed` instead?

suggestion: adjust the tests to include this property has well.

libs/dh/esett/data-access-outgoing-messages/src/lib/dh-outgoing-messages.store.ts
mimse: outgoingMessageQuery is a signal as well, so i think it needs to be in computed

libs/dh/shared/util-apollo/src/lib/query.ts
mimse: Good ideas, now implemented

dzhavat (approved):
praise: awesome change! Really like how we could move the repetitive pattern out of the components and make the code much simpler.

PR #3826, opened 2024-12-10
feat(dh): support two new badge statuses 

dzhavat:
## References

- #0000


mimse (dismissed):
Nice dotnet work!

apps/dh/api-dh/source/DataHub.WebApi/GraphQL/DataLoaders/ConsolidationByIdBatchDataLoader.cs
FirestarJes, line 1: Should probably be named ConsolidationByActorFromIdBatchLoader, otherwise it could be confused with one that could load them by their own id

FirestarJes (changes requested):

FirestarJes (approved):

PR #3828, opened 2024-12-10
Fix download

mimse:
Fixes https://github.com/Energinet-DataHub/team-raccoons/issues/969

### DIFF
--- a/libs/dh/admin/feature-user-roles/src/lib/download.component.ts
+++ b/libs/dh/admin/feature-user-roles/src/lib/download.component.ts
@@ -21,10 +21,22 @@ import { Component, inject, input } from '@angular/core';
 import { take } from 'rxjs';
 import { translate, TranslocoPipe, TranslocoService } from '@ngneat/transloco';
 
+import { wattFormatDate } from '@energinet-datahub/watt/date';
+import { WattToastService } from '@energinet-datahub/watt/toast';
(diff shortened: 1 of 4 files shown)


libs/dh/admin/feature-user-roles/src/lib/download.component.ts
```diff
@@ -35,30 +47,60 @@ import { DhUserRoles } from '@energinet-datahub/dh/admin/data-access-api';
...
-          `"${rolesTranslations[role.eicFunction]}"`,
-          `"${translate('admin.userManagement.roleStatus.' + role.status)}"`,
-        ]);
+          const lines = userRoles.map((role) => [
+            `"${role.name}"`,
+            `"${rolesTranslations[role.eicFunction]}"`,
+            `"${translate('admin.userManagement.roleStatus.' + role.status)}"`,
+          ]);
+          const env = translate(`envinronementName.${this.environment.current}`);
+          const fileName = `DataHub-User-roles-${wattFormatDate(new Date(), 'long')}-${env}`;
```
dzhavat, line 94: praise: nice work for making the file name more descriptive. The user's download uses something similar which was appreciated by our service desk.

suggestion: it's a small details but it would be awesome to unify the download file names (where it makes sense). It will make the experience more unified. Let's track it in a separate task (will create one on our board)

dzhavat (approved):

PR #3838, opened 2024-12-13
Chore/upgrade angular to 19 and nx to 21

mimse:
Updated angular to 19
Updated Storybook 8
Updated NX to 20.3


libs/dh/shared/feature-flags/src/lib/dh-feature-flag.directive.spec.ts
```diff
@@ -40,31 +40,30 @@ describe(DhFeatureFlagDirective, () => {
...
-    expect(queryContent()).toBeInTheDocument();
+    const { queryByText } = await setup('enabled-flag');
+    expect(queryByText(/SOME CONTENT/i));
   });
 
   it('should not render content, if feature flag is disabled', async () => {
-    await setup('disabled-flag');
-    expect(queryContent()).not.toBeInTheDocument();
+    const { queryByText } = await setup('disabled-flag');
+    expect(queryByText(/SOME CONTENT/i));
```
dzhavat, line 67: issue: if the content here can be found, this is actually an issue 🕵️‍♀️ The tests says that the text shouldn't render

mimse: Good catch, fixed now

dzhavat (approved):

PR #3844, opened 2024-12-18
EO: Add license executor to run after generating translation keys

frontend-specialisten:
## Description
 - Add license executor to run after generating translation keys

## References

- #0000


MJ1227UB (approved):
Neat ! :) No more reverting the top of the translation file! 😋

PR #3846, opened 2024-12-18
feat(dh): adjustments to settlement report

dzhavat:
### DataHub
- move period selection to the top 
- fetch grid areas based on the selected period

## References

- #0000


mimse (approved):

mimse (approved):
Suggestion: I would not add the grid area query to shared. I think it was a mistake to move the queries away from where they are used. Better to have more specific query closer to the there are used, IMO.

dzhavat:
Placing the query under Settlement reports now makes sense but it's also going to be used in Calculations very soon. In this case "Settlement reports" is not the best place anymore. Maybe "wholesale/data-access-graphql" makes more sense.

mimse:
I would have two query with the same content, but with different names. ex. GetGridAreasForSettlementReports and GetGridAreasForCalculations. Then dont the road if new fields a are added to only one of them, it do not over fetch data in the other because they are separated

dzhavat:
Interesting. Will need to think about it because the query right now is only used in a util function that maps the result to dropdown options. The function can of course be modified to take the query document as an input.

PR #3848, opened 2024-12-18
Load msw before app in mock mode

mimse:
To avoid async hacks to check if msw is loaded, we now loaded before the app is initialized


apps/dh/app-dh/src/main.ts
```diff
@@ -55,7 +56,24 @@ if (environment.authDisabled) {
...
+    serviceWorkerPromise = setupServiceWorker(dhLocalApiEnvironment.apiBase, mocks);
+  });
+}
+
+Promise.all([
+  loadDhApiEnvironment(),
+  loadDhB2CEnvironment(),
+  loadDhAppEnvironment(),
+  serviceWorkerPromise,
+])
```
dzhavat, line 76: issue: It feels like there's a race condition here. While the `Promise.all` inside the `if` executes, the code will continue to the next `Promise.all`. If the first `Promise.all` completes by that time, `setupServiceWorker` will execute and the mocks will be set up correctly. Otherwise the setup will run after the app has bootstrapped which can make the app unstable.

mimse: Could be fixed now :-)

dzhavat (approved):
praise: great work. The mock mode is going to be more stable now 🚀

PR #3854, opened 2024-12-19
chore: monorepo = monoencoding

ManBearTM:
Having multiple forms of encoding in the same repo is weird and causes issues with editors that are not owned by M$ 😄 The `utf8-bom` rule is outdated and basically all editors default to utf8 (without bom). Yes, even Visual Studio. Also, the aspnetcore repository doesn't even use `utf8-bom` in their [`.editorconfig`](https://github.com/dotnet/aspnetcore/blob/49df2e6c7207f6845b671b29e17d58d91edd6e93/.editorconfig#L18).

Also [this](https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-with-bom/2223926#2223926).

And [this](https://www.rfc-editor.org/rfc/rfc7159#section-8.1).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -19,8 +19,5 @@
   },
   "prettier.prettierPath": "./node_modules/prettier/index.cjs",
   "eslint.validate": ["json", "graphql"],
-  "dotnet.defaultSolution": "apps/dh/api-dh/DataHub.WebApi.sln",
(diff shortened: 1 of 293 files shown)


mimse:
dotnet will not compile without bom :-(

mimse (approved):

PR #3857, opened 2024-12-19
chore(dh): update hot-toast dependency

dzhavat:
### Workspace
- remove custom mock since "hot-toast" can now run in Jest

## References

- #0000


mimse (approved):
Praise: Nice one less hack

PR #3880, opened 2025-01-13
Added simple view for electricity market

mimse:


dzhavat (dismissed):
praise: looks good.

suggestion: if the `libs/dh/electricity-market` is intended to be a domain (wrapping) folder  that will contain multiple pages in the future, it will be hard to expand in its current format. That is because the folder itself is an Nx library. A future PR can move the Nx library to a sub-folder in order to make `electricity-market` truly a domain/wrapping folder.

dzhavat (approved):

PR #3887, opened 2025-01-16
Chore/use vite in storybook

mimse:


libs/watt/project.json
```diff
@@ -36,17 +30,10 @@
...
-        "outputDir": "dist/designsystem",
-        "configDir": "libs/watt/.storybook",
-        "browserTarget": "watt:build-storybook",
-        "compodoc": false,
-        "stylePreprocessorOptions": {
-          "includePaths": ["libs/watt/src/lib/styles"]
-        },
-        "styles": ["libs/watt/.storybook/styles.scss"]
+        "cwd": "libs/watt",
+        "command": "storybook build --output-dir ../../dist/storybook/my-app"
```
dzhavat, line 36: question: wonder whether the change to output directory from `designsystem` to `storybook/my-app` causes any deployment issues with Chromatic (after the PR is merged)?

libs/watt/.storybook/main.ts
```diff
@@ -1,26 +1,63 @@
...
+import type { UserConfig } from 'vite';
+import { nxViteTsPaths } from '@nx/vite/plugins/nx-tsconfig-paths.plugin';
 
-const config: StorybookConfig = {
-  staticDirs: [
-    {
-      from: '../src/assets',
-      to: 'assets/watt',
-    },
-  ],
```
dzhavat, line 9: issue: this will need to be kept somehow because the custom icons as well as the visual spacing reference image (under Spacing) do not load.

mimse: Good catch, will have a look

libs/watt/project.json
mimse: Think you are right, will looking to setting the same output path

libs/watt/src/lib/components/stepper/watt-stepper.component.ts
```diff
@@ -60,9 +60,7 @@ export class WattStepperComponent extends MatStepper implements AfterViewInit {
   @Input() isCompleting = false;
 
   @ContentChildren(WattStepperStepComponent, { descendants: true })
-  override _steps!: QueryList<WattStepperStepComponent>;
-
-  override readonly steps!: QueryList<WattStepperStepComponent>;
+  declare _steps: QueryList<WattStepperStepComponent>;
```
dzhavat, line 63: nitpick: TIL that this can be used too. Don't think I've seen it in this context 🤔

mimse: If i dont do this, storybook will not build.. I seems not to have any side effects

mimse: Or i could be wrong because now apps dont build..

mimse: Ah used the wrong collection

dzhavat (approved):

PR #3901, opened 2025-01-21
fixerino cappucino

sahma19:
## Description

Transfer tests are flaky. This is a hacky fix.


MJ1227UB (commented):
But.. I'm not sure I agree that we should go after the smaller screen size quite yet. For now we have made the page responsive, but have not agreed whether or not, or even how we should handle mobile users on our webapp. I would prefer we kept on assuming that we're making a desktop app, therefore we should run our tests on a screen size that also reflects that. At least on Energy Track and Trace :) Hope it makes sense

PR #3907, opened 2025-01-22
Feature/query time

mimse:


dzhavat (approved):

apps/dh/api-dh/source/DataHub.WebApi/Registration/GraphQLRegistrationExtensions.cs
```diff
@@ -27,6 +27,7 @@ public static IRequestExecutorBuilder AddGraphQLServices(this IServiceCollection
     {
         return services
             .AddGraphQLServer(disableDefaultSecurity: true)
+            .InitializeOnStartup()
```
ManBearTM, line 30: Is this to speed up the first query or? 🙂

mimse: Yes

ManBearTM (approved):
Not loving the console spam, but I'm open to try it 😄

PR #3912, opened 2025-01-24
feat(dh): customer card to metering point overview

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-overview/src/lib/dh-customer-overview.component.ts
```diff
@@ -0,0 +1,117 @@
...
+          </button>
+        </watt-description-list-item>
+      </watt-description-list>
+
+      <h4 class="watt-space-stack-s">Kunde 2</h4>
+
+      <watt-description-list variant="stack" [itemSeparators]="false">
+        <watt-description-list-item [label]="t('nameLabel')" [value]="null | dhEmDashFallback" />
+        <watt-description-list-item [label]="t('cprLabel')">
+          <button type="button" class="show-cpr-button">
```
mimse, line 109: Suggestion: move show cpr button into its own component

libs/dh/metering-point/feature-overview/src/lib/dh-customer-overview.component.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+    }
+  `,
+  template: `
+    <watt-card *transloco="let t; read: 'meteringPoint.overview.customer'" class="customer">
+      <watt-card-title>
+        <h3>{{ t('title') }}</h3>
+      </watt-card-title>
+
+      <div
```
mimse, line 82: Comment: @PernilleChristiansen said we should just use badge

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.ts
```diff
@@ -19,27 +19,59 @@
...
-      }
+    .page-content {
+      margin: var(--watt-space-ml);
     }
   `,
   template: `
-    <div *transloco="let t; read: 'meteringPoint.overview'" class="page-header">
-      <h2>{{ meteringPointId() }}</h2>
-    </div>
+    <ng-container *transloco="let t; read: 'meteringPoint.overview'">
```
mimse, line 54: Suggestion: Can we move *transloco to the div, and get rid of the ng-container?

mimse (changes requested):
Looks good, a few suggestions

libs/dh/metering-point/feature-overview/src/lib/dh-customer-overview.component.ts
dzhavat: It looks a bit ugly compared to Figma, but that's what we got :) 
One limitation is that the badge does not support icons. Projecting an icon doesn't look good. So we should only stick to text

image

dzhavat: Let's continue with the custom component.

mimse (approved):

PR #3913, opened 2025-01-24
Refactor/esett data table

mimse:


libs/dh/esett/data-access-outgoing-messages/src/index.ts
```diff
@@ -17,5 +17,4 @@
  */
 //#endregion
 export { DhEsettDataAccessApiStore } from './lib/dh-esett-data-access-api.store';
-export { DhOutgoingMessagesSignalStore } from './lib/dh-outgoing-messages.store';
```
dzhavat, line 20: nitpick: the actual file can be deleted as well

dzhavat (commented):
praise: looks waaay better. Great work 💪 
Only one small nitpick. Can be a follow up PR together with removing the `signal` package.

dzhavat (approved):

libs/dh/esett/data-access-outgoing-messages/src/index.ts
mimse: You are right, will remove in followup PR

PR #3925, opened 2025-01-27
refac(dh): settlement reports module

ManBearTM:
There are changes to the schema snapshot, but only concerning ordering (and a single type name, which was updated to match in the frontend). There should be no impact otherwise.

I also had to enable `EmitCompilerGeneratedFiles` since otherwise tests that would run fine locally would fail in the CI. I suspect this is happening due to the weird CI setup, where the files are compiled on a windows machine, then copied to a linux machine where the tests are executed. By setting `EmitCompilerGeneratedFiles` to true, the source generated files are emitted in a `generated` folder under `Release` or `Debug`, meaning they are now copied along with the other files while still being excluded from source control. I don't see any downsides to this approach.


mimse (approved):

mimse:
Praise: nice work

PR #3929, opened 2025-01-28
Refactor to use the new module way

mimse:


apps/dh/api-dh/source/DataHub.WebApi/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+                var baseUrls = provider.GetRequiredService<IOptions<SubSystemBaseUrls>>();
+                var baseUrl = getBaseUrl(baseUrls.Value);
+                var client = provider.GetRequiredService<AuthorizedHttpClientFactory>().CreateClient(GetBaseUri(baseUrl));
+                return createClient(baseUrl, client);
+            });
+    }
+
+    private static Uri GetBaseUri(string baseUrl)
+    {
+        return Uri.TryCreate(baseUrl, UriKind.Absolute, out var url)
```
defectiveAi, line 37: Throw instead?

defectiveAi (approved):

PR #3931, opened 2025-01-29
feat(dh): status and highlights badges in metering point

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.ts
```diff
@@ -89,7 +98,10 @@ import { DhMeteringPointDetailsComponent } from './dh-metering-point-details.com
   `,
   template: `
     <div *transloco="let t; read: 'meteringPoint.overview'" class="page-header">
-      <h2 class="watt-space-stack-s">{{ meteringPointId() }}</h2>
+      <h2 vater-stack direction="row" gap="m" class="watt-space-stack-s">
```
mimse, line 101: Question: Should both have the directive stack and the class stack?

mimse (approved):

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.ts
dzhavat: Yeah, the `watt-space-stack-s` add `margin` to the element whereas `vater-stack` sets its layout (`display: flex`)

PR #3943, opened 2025-01-30
fix(dh): implement GetRelevantGridAreas for create calculation

ManBearTM:


Sejruppen (approved):
Neat! 🙂👍

libs/dh/wholesale/feature-calculations/src/lib/grid-areas/dropdown.component.ts
```diff
@@ -48,30 +47,36 @@ import { TranslocoDirective } from '@ngneat/transloco';
...
   control = input.required<FormControl<string[] | null>>();
-  period = input.required<Range<string> | null>();
-  disabled = input(false);
+  period = input<WattRange<Date> | null>();
+
+  gridAreasQuery = lazyQuery(GetRelevantGridAreasDocument, { fetchPolicy: 'network-only' });
+  isResolved = computed(() => this.gridAreasQuery.status() === QueryStatus.Resolved);
 
-  gridAreasQuery = query(GetGridAreasDocument);
+  fetchGridAreas = effect(() => {
```
mimse, line 67: Suggestion: Move effect to a constructor to avoid create a unused variable

libs/dh/wholesale/feature-calculations/src/lib/grid-areas/dropdown.component.ts
```diff
@@ -48,30 +47,36 @@ import { TranslocoDirective } from '@ngneat/transloco';
...
+    const period = this.period();
+    if (!period) this.gridAreasQuery.reset();
+    else this.gridAreasQuery.refetch({ period });
+  });
 
   gridAreas = computed(() => {
-    const gridAreas = this.gridAreasQuery.data()?.gridAreas ?? [];
+    const gridAreas = this.gridAreasQuery.data()?.relevantGridAreas ?? [];
 
     // HACK: This is a temporary solution to filter out grid areas that has no data
```
mimse, line 76: Question: Is it still a hack when only relevantGridArea is fetched?

mimse (commented):
Just a few comments :-)

libs/dh/wholesale/feature-calculations/src/lib/grid-areas/dropdown.component.ts
ManBearTM: Yes, that comment refers to the `gridAreas.filter((g) => ['803', '804', '533', '543', '584', '950']` line, which removes grid areas in the test environment that has no real data (and thus would cause calculations to fail). I haven't heard anything about this not being the case anymore, but I'll check. Would be nice to no longer need that hack 👍

libs/dh/wholesale/feature-calculations/src/lib/grid-areas/dropdown.component.ts
ManBearTM: But I like the variable, it allows me to name the effect 😄

mimse: Up to you, I just think it its a bit weird :P

libs/dh/wholesale/feature-calculations/src/lib/grid-areas/dropdown.component.ts
mimse: 👍🏻

ManBearTM: Confirmed to still be required unfortunately

libs/dh/wholesale/feature-calculations/src/lib/grid-areas/dropdown.component.ts
ManBearTM: agreed, but signals are weird in general 😄

PR #3958, opened 2025-02-05
Feature/dev example with process detail page

mimse:


libs/dh/dev-examples/feature-processes/graphql/get-processes.graphql
```diff
@@ -23,6 +23,28 @@ query GetProcesses(
...
+        meteringPointType
+        calculationType
+        period
+      }
+      ... on Calculation {
+        __typename
+        gridAreas {
+          id
+          displayName
+        }
```
ManBearTM, line 43: suggestion: for the sake of the example, you could consider not fetching gridAreas for the table view (since it is an additional request) and instead only fetch them when opening the details view

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
```diff
@@ -0,0 +1,212 @@
...
+  `,
+})
+export class DhProcessDetailsComponent {
+  private navigation = inject(DhNavigationService);
+  private processQuery = lazyQuery(GetProcessDocument);
+
+  loading = this.processQuery.loading;
+  hasError = this.processQuery.hasError;
+  result = computed(() => this.processQuery.data()?.processById);
+  startedAtOrScheduledAt = computed(() => this.result()?.scheduledAt ?? this.result()?.scheduledAt);
```
dzhavat, line 178: suggestion: `scheduledAt` is used both as value and fallback

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
```diff
@@ -0,0 +1,212 @@
...
+  );
+
+  energyTimeSeriesRequestDetails = computed(() =>
+    this.result() && this.result()?.__typename === 'RequestCalculatedEnergyTimeSeriesResult'
+      ? (this.result() as DhProcessEnergyTimeSeriesRequest)
+      : null
+  );
+
+  wholesaleRequestDetails = computed(() =>
+    this.result()?.__typename === 'RequestCalculatedWholesaleServicesResult'
```
dzhavat, line 193: suggestion: shouldn't this also start with `this.result() &&` like the previous two

```suggestion
    this.result() && this.result()?.__typename === 'RequestCalculatedWholesaleServicesResult'
```

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
```diff
@@ -0,0 +1,212 @@
...
+      ? (this.result() as DhProcessWholesaleRequest)
+      : null
+  );
```
ManBearTM, line 196: suggestion: since this is supposed to be about best practices it doesn't feel good to use `as` in our examples (that is usually bad practice). you should be able to rewrite this code without using these unsafe type casts:

```ts
  calculationDetails = computed(() => {
    const result = this.result();
    return result?.__typename === 'Calculation' ? result : null;
  });
```

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
ManBearTM, line 178: issue: the code is identical on either side of `??`

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    dh-calculation-details-grid-areas {
+      margin: 0 var(--watt-space-ml);
+    }
+
+    watt-progress-tracker {
+      margin: 0 var(--watt-space-ml);
+    }
+  `,
```
ManBearTM, line 84: question: do we want to advocate for styles like this? it would be better IMO if we could solve this with `vater`.

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
```diff
@@ -0,0 +1,212 @@
...
+  template: `
+    @let process = result();
+    @let calculation = calculationDetails();
+    @let energyTimeSeriesRequest = energyTimeSeriesRequestDetails();
+    @let wholesaleRequest = wholesaleRequestDetails();
+
+    <watt-drawer *transloco="let t; read: 'devExamples.processes'" (closed)="closed()">
```
ManBearTM, line 91: suggestion: maybe inline the contents of `closed()`?

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
```diff
@@ -0,0 +1,212 @@
...
+      <watt-drawer-topbar>
+        @if (process) {
+          <dh-process-state-badge [status]="process.state">
+            {{ 'shared.states.' + process.state | transloco }}
```
ManBearTM, line 95: comment: from the perspective of someone inexperienced with the codebase, it could be confusing why we are using the transloco pipe here while also using the structural directive. I know we use it to "reach" global translations, but it always seemed a bit "quick-fixy" to me 😄 

also, I think `read` is deprecated and it is `prefix` now (at least if we are using the latest version).

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  // Param value
+  id = input.required<string>();
+  drawer = viewChild(WattDrawerComponent);
+
+  constructor() {
+    afterRenderEffect(() => {
+      this.processQuery.query({ variables: { id: this.id() } });
+      this.drawer()?.open();
+    });
```
ManBearTM, line 206: suggestion: is there a reason to query here in `afterRenderEffect` as opposed to just using a regular `query`? and if there is, I think it would be a good idea to add a comment here.

it could also be cool to have an "autoOpen" input or similar on the drawer instead of having to do these shenanigans

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
```diff
@@ -0,0 +1,70 @@
...
+        [hideColumnHeaders]="true"
+        sortBy="displayName"
+        sortDirection="asc"
+      />
+    </watt-data-table>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DhCalculationsDetailsGridAreasComponent {
+  sort = viewChild.required(MatSort);
```
ManBearTM, line 54: question: is this used?

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import { WATT_CARD } from '@energinet-datahub/watt/card';
+import { WattDataTableComponent } from '@energinet-datahub/watt/data';
+import { WattTableDataSource, WattTableColumnDef, WATT_TABLE } from '@energinet-datahub/watt/table';
+
+import { DhProcessCalculationGridArea } from '../types';
+
+@Component({
+  imports: [MatSortModule, TranslocoDirective, WATT_TABLE, WattDataTableComponent, WATT_CARD],
+  selector: 'dh-calculation-details-grid-areas',
```
ManBearTM, line 31: issue: this selector doesn't look right

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
```diff
@@ -0,0 +1,70 @@
...
+        [suppressRowHoverHighlight]="true"
+        [hideColumnHeaders]="true"
+        sortBy="displayName"
+        sortDirection="asc"
+      />
+    </watt-data-table>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DhCalculationsDetailsGridAreasComponent {
```
ManBearTM, line 53: issue: this component name should also be updated

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  sort = viewChild.required(MatSort);
+
+  gridAreas = input.required<DhProcessCalculationGridArea[]>();
+
+  datasource: WattTableDataSource<DhProcessCalculationGridArea> = new WattTableDataSource(
+    undefined
+  );
+  columns: WattTableColumnDef<DhProcessCalculationGridArea> = {
```
ManBearTM, line 61:
```suggestion

  columns: WattTableColumnDef<DhProcessCalculationGridArea> = {
```

libs/dh/dev-examples/feature-processes/src/table.component.ts
```diff
@@ -54,7 +58,12 @@ import { DhProcessStateBadge } from '@energinet-datahub/dh/wholesale/shared';
     >
       <h3>{{ t('headline') }}</h3>
 
-      <watt-table [dataSource]="dataSource" [columns]="columns" [loading]="dataSource.loading">
+      <watt-table
+        [dataSource]="dataSource"
+        [columns]="columns"
+        [loading]="dataSource.loading"
+        (rowClick)="navigate($event.id)"
+        >)>
```
ManBearTM, line 66: `>)>` what? 😄

libs/dh/dev-examples/feature-processes/src/table.component.ts
```diff
@@ -54,7 +58,12 @@ import { DhProcessStateBadge } from '@energinet-datahub/dh/wholesale/shared';
     >
       <h3>{{ t('headline') }}</h3>
 
-      <watt-table [dataSource]="dataSource" [columns]="columns" [loading]="dataSource.loading">
+      <watt-table
+        [dataSource]="dataSource"
+        [columns]="columns"
+        [loading]="dataSource.loading"
+        (rowClick)="navigate($event.id)"
```
ManBearTM, line 65: suggestion: we could inline here as well (if we think that is best practice, not sure)

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -76,6 +76,36 @@
...
+      },
+      "2": {
+        "pending": "Afventer beskeddannelse",
+        "running": "Danner beskeder",
+        "failed": "Beskeddannelse fejlet",
+        "skipped": "Beskeddannelse sprunget over",
+        "canceled": "Beskeddannelse annulleret",
+        "succeeded": "Beskeder dannet"
+      }
+    },
```
ManBearTM, line 108: suggestion: these are specific to calculations and should not be considered shared

libs/dh/shared/navigation/src/navigation.service.ts
```diff
@@ -73,7 +73,7 @@ export class DhNavigationService {
     }
 
     if (path === 'details') {
-      this.router.navigate([path, id], {
+      this.router.navigate([path, id, ...args], {
```
dzhavat, line 76: question: the `args` doesn't seem to be used. Is it something coming later?

ManBearTM (changes requested):

tsconfig.base.json
```diff
@@ -47,6 +47,10 @@
       ],
       "@energinet-datahub/dh/core/routing": ["libs/dh/core/routing/src/index.ts"],
       "@energinet-datahub/dh/core/shell": ["libs/dh/core/shell/src/index.ts"],
+      "@energinet-datahub/dh/dev-examples": ["libs/dh/dev-examples/src/index.ts"],
+      "@energinet-datahub/dh/dev-examples/feature-processes": [
+        "libs/dh/dev-examples/feature-processes/src/index.ts"
+      ],
```
dzhavat, line 53: issue: this way of nesting one library inside another diverges slightly from the rest of the codebase. I'm mostly concerned about config files where a file from the "parent" lib "leaks" into the nested lib. Why not separate the two libs since the "boilerplate" files are already there?

libs/dh/dev-examples/feature-processes/graphql/get-process.graphql
```diff
@@ -0,0 +1,36 @@
+query GetProcess($id: UUID!) {
```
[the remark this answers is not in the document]
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/815cbe7ff94f1ca00bbed70b98008ceafe1126e8

libs/dh/dev-examples/feature-processes/graphql/get-processes.graphql
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/131eb0512bd099ad65160cd7c597c7c77edb6a4b

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/54fe4cfaea99a5abf8b886ce8152362924be46dd

libs/dh/shared/navigation/src/navigation.service.ts
mimse: We are now able to add as many values to the query string as we want. I had a use case for it, but refactored it to do it in another way, but decided to keep the functionality.

tsconfig.base.json
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/7b43583eda02f31a3dcc965e51f79f73dcbe2ddb

libs/dh/dev-examples/feature-processes/src/table.component.ts
mimse: Fixed in 2ecf51ac6

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/commit/54fe4cfaea99a5abf8b886ce8152362924be46dd

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
mimse: I agree a single line is good to inline fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/20e4154dee826ad501630d27de7f5d1c09a04d6e

libs/dh/dev-examples/feature-processes/src/table.component.ts
mimse: I agree a single line is good to inline fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/20e4154dee826ad501630d27de7f5d1c09a04d6e

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/0cacdd16f0f13212375bf7121ef435315f6a0114

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
mimse: The query is in an effect because its needs the id to be set. As to why it is a afterRenderEffect instead of an normal effect is because it saw issues with values changing after render and we get that angular error https://angular.dev/errors/NG0100

Think its a good idea with auto open

mimse: Drawer has some odd logic around reusing the existing drawer if a new one is opened. The auto open input, breaks this logic.
Might do a PR later to see if it can be implemented.

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
mimse: Why not? It is grid areas only for calculations?

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
mimse: Think its okay

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
mimse: Oooh, do we have some more generic naming for a process then?

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/4b4c74497dcf37d4de07d7f4b1fa0ae934c49286

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/1eafadc5be272e88e463060d0315a9266cc8500f

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
mimse: If you have a better solution I´m all ears :-) prefix doesnt work. Guess we are not one the latest version

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
ManBearTM: I didn't realize this was only for calculations, so I see your point. That said, do we want this component to use the same selector/component "prefix" as components in `wholesale/feature-calculations`? Again, normally I probably wouldn't care that much, but since this is about best practices, maybe we should settle on how we prefix components?

For example, there is the super-overly-long-but-technically-correct-way:
`dh-dev-examples-feature-processes-calculation-details-grid-areas`

Or the one that skips domain and the feature part (fx. we use this in calculations and actors):
`dh-processes-calculations-details-grid-areas`

Or the one like yours that just don't care because it is never used outside the feature (which is a valid argument):
`dh-calculations-details-grid-areas`

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
ManBearTM: See comment above 🙂

libs/dh/dev-examples/feature-processes/src/details/gridareas.component.ts
mimse: What was my thought. That it was okay because it was never exported and only used as a feature component, so it can not be referenced in other domain / feature. What say you @dzhavat

dzhavat: We have started naming components with shorter names when they are internal for a library. From that point of view it doesn't matter much that the selector is not technically correct.

Looking more long-term, the Angular team is looking at **selectorless** components, which will land at some point this year. If **selectorless** means what we think it does, the selector itself will become less important over time.

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
ManBearTM: No, steps are always specific to individual "processes", meaning the number of steps can be different between processes and fx. step 1 is not always a "calculation" step, but could be anything. So this translation only applies to calculation processes and should not be used for requests. I don't actually know what the steps for a request are, so maybe we should just only show steps for calculations (for now)? And then move these translations back into the "wholesale.calculations" area.

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
```diff
@@ -0,0 +1,193 @@
...
+  `,
+})
+export class DhProcessDetailsComponent {
+  private processQuery = lazyQuery(GetProcessByIdDocument);
+
+  navigation = inject(DhNavigationService);
+  loading = this.processQuery.loading;
+  hasError = this.processQuery.hasError;
+  result = computed(() => this.processQuery.data()?.processById);
+  startedAtOrScheduledAt = computed(() => this.result()?.scheduledAt);
```
ManBearTM, line 166: suggestion: now the naming implies that the value is either `startedAt` or `scheduledAt`, but it is always `scheduledAt`. I would add `startedAt` so it works like in calculations.
```suggestion
  startedAtOrScheduledAt = computed(() => this.result()?.startedAt ?? this.result()?.scheduledAt);
```

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
ManBearTM: I really need to update `util-apollo` to accept a variables function just like `request` in the Angular resource API - that should prevent this exact issue

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
ManBearTM: It's fine for now 🙂 I would like to move towards [translation scopes](https://jsverse.gitbook.io/transloco/advanced-topics/lazy-load/scope-configuration) and then just always using `t()`. That way we would save some initial loading and we could flatten the translation object by using multiple scopes instead. Something like this:

```html
<watt-drawer *transloco="let t;">
  <dh-process-state-badge [status]="process.state">
    {{ t('shared.states.' + process.state) }}

    ...

  <watt-description-list-item
    [label]="t('details.startedBy')"
```

In this setup, `shared` would be its own translation file, as could `details`.

libs/dh/dev-examples/feature-processes/src/details/details.component.ts
mimse: Ofc, my bad :-)

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
mimse: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/3958/commits/6611bd5f0e0156e5eafdd2c035af715ce39b8087

ManBearTM (approved):

PR #3959, opened 2025-02-05
fix pipeline by adding checkout to update acorn job

sahma19:
## References

- #0000


mimse (approved):

Karl-koder:
The fix has been applied in the acorn action, So this is not necessary and can be closed.

PR #3963, opened 2025-02-06
feat(dh): electrical heating calculations

ManBearTM:


libs/dh/wholesale/feature-settlement-reports/src/lib/modal/dh-request-settlement-report-modal.component.ts
```diff
@@ -320,7 +323,7 @@ export class DhRequestSettlementReportModalComponent extends WattTypedModal<Sett
             includeMonthlySums: includeMonthlySum,
             gridAreasWithCalculations: this.getGridAreasWithCalculations(
               gridAreas,
-              calculationType == CalculationType.BalanceFixing
+              calculationType == WholesaleAndEnergyCalculationType.BalanceFixing
```
dzhavat, line 326: nitpick:

```suggestion
              calculationType === WholesaleAndEnergyCalculationType.BalanceFixing
```

libs/dh/wholesale/feature-calculations/src/lib/table/table.component.ts
```diff
@@ -80,7 +81,10 @@ export class DhCalculationsTableComponent {
 
   columns: WattTableColumnDef<Calculation> = {
     calculationType: { accessor: 'calculationType' },
-    period: { accessor: 'period', size: 'minmax(max-content, auto)' },
+    period: {
+      accessor: (r) => (r.__typename == 'WholesaleAndEnergyCalculation' ? r.period : null),
```
dzhavat, line 85: nitpick:

```suggestion
      accessor: (r) => (r.__typename === 'WholesaleAndEnergyCalculation' ? r.period : null),
```

libs/dh/wholesale/domain/src/calculation.ts
```diff
@@ -16,29 +16,32 @@
  * limitations under the License.
  */
 //#endregion
+import dayjs from 'dayjs';
```
dzhavat, line 19: suggestion: let's use `dayjs` from Watt

```suggestion
import { dayjs } from '@energinet-datahub/watt/date';
```

libs/dh/wholesale/feature-calculations/src/lib/details/details.component.html
```diff
@@ -71,10 +71,12 @@ <h3 class="headline watt-headline-2">{{ t("calculationTypes." + type()) }}</h3>
         [label]="t('details.executionTime')"
         [value]="startedAtOrScheduledAt() | wattDate: 'long' | dhEmDashFallback"
       />
-      <watt-description-list-item
-        [label]="t('details.period')"
-        [value]="calculation?.period | wattDate: 'short' | dhEmDashFallback"
-      />
+      @if (wholesaleAndEnergyCalculation?.period || query.loading()) {
```
dzhavat, line 74: question: is the `query.loading()` intentional here? None of the other fields have it.

libs/dh/wholesale/domain/src/calculation.ts
ManBearTM: b23ef7570c03e07229b82a70b315886349fe0099

libs/dh/wholesale/feature-calculations/src/lib/details/details.component.html
ManBearTM: it is intentional, yes 🙂 it sticks out I agree, but it is because `period` makes no sense for electrical heating calculations, so in that case they want to just hide the field entirely. but since I didn't want the field to "pop in" for normal calculations, I included the loading check, so it is only hidden once it knows for sure if it is an electrical heating calculation or not.

but I can see if I can make the logic a bit clearer on that

dzhavat: I see. Well, if it's intentional then it's ok to leave as is. I thought it looked like a leftover from testing something, that's why I pointed it out.

libs/dh/shared/data-access-graphql/src/lib/dh-graphql.providers.ts
```diff
@@ -75,6 +76,7 @@ export const graphQLProviders = makeEnvironmentProviders([
           },
         },
         cache: new InMemoryCache({
+          possibleTypes: introspection.possibleTypes,
```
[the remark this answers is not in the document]
ManBearTM: this is the `possibleTypes` configuration generated automatically from the schema. it helps apollo to understand relationships between types via interfaces and unions.

I needed it because the cache lookup was failing for calculations after it was updated to use graphql interfaces. i don't know much about the inner workings of this, but with this configuration I was able to add cache configuration for the `Calculation` type (which is an interface) and have it automatically apply to all subtypes (e.g. `WholesaleAndEnergyCalculation`, `ElectricalHeatingCalculation`). This means that the `calculationById` query could successfully look in the cache for a matching calculation. before it would fail to find anything, meaning opening the drawer would always require an extra fetch.

codegen.ts
```diff
@@ -30,6 +30,9 @@ const config: CodegenConfig = {
         },
       },
     },
+    'libs/dh/shared/domain/src/lib/generated/graphql/introspection.ts': {
+      plugins: ['fragment-matcher'],
```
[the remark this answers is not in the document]
ManBearTM: this plugin generates the `possibleTypes` configuration from the schema (see comment below)

apps/dh/api-dh/source/DataHub.WebApi.Tests/Integration/GraphQL/Calculation/CalculationGridAreasQueryTests.cs
```diff
@@ -32,7 +32,7 @@ public class CalculationGridAreasQueryTests
     $$"""
       query {
         calculationById(id: "{{OrchestrationInstanceFactory.Id}}") {
-          ... on Calculation {
+          ... on WholesaleAndEnergyCalculation {
```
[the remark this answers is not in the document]
ManBearTM: Thanks, I struggled a lot with the naming in this one, so nice for it to be appreciated 😄

package.json
```diff
@@ -80,6 +80,7 @@
     "@apollo/client": "3.8.1",
     "@azure/msal-angular": "3.0.16",
     "@azure/msal-browser": "3.13.0",
+    "@graphql-codegen/fragment-matcher": "5.0.2",
```
dzhavat, line 83: question: does this need to be under "dependencies" and not "devDependencies"?

ManBearTM: 4c4fd025c49fc48d19f614bd4db40b6516e7d15b

apps/dh/api-dh/source/DataHub.WebApi/Modules/Common/Types/WholesaleAndEnergyCalculationTypeEnum.cs
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Energinet.DataHub.WebApi.Modules.Common;
+
+public class WholesaleAndEnergyCalculationTypeEnum : EnumType<WholesaleAndEnergyCalculationType>
+{
+    protected override void Configure(IEnumTypeDescriptor<WholesaleAndEnergyCalculationType> descriptor)
+    {
+        descriptor.Name("WholesaleAndEnergyCalculationType");
```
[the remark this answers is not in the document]
ManBearTM: I think it is a good idea as long as we are pretty certain the enum is not going to change/diverge. Having had to refactor CalculationType related code sooo many times because of our effort to unify it in the GraphQL API, I have kinda walked back on this a little.

Something I have been thinking about is to "split" input and output enums. For example, with CalculationType's, we now have an extra "ElectricalHeating" type that is (currently) only used in calculations. But this enum could still be fine to return for e.g. when querying settlement reports - maybe it can't be an electrical heating calculation, but that usually doesn't matter and as an added bonus, the UI is now a little more future proof since it could handle this type (if needed).

But this enum cannot be used for "input" types. With that, I mean enums used to render e.g. filters/dropdowns for which to initiate searches or mutations. When requesting settlement reports, we do not want "Electrical Heating" to show up all of a sudden. This is what I mean by "splitting" the enums. For example, the schema could look something like this:

```
enum CalculationType {
   Aggregation
   ...
   ElectricalHeating
}

enum RequestSettlementReportCalculationType {
  Aggregation
  ...
}
```

We could then have many specific calculation types used for inputs, but only one (if possible) for output. Anyway, just some ramblings from my mind, not sure if it is feasible or not. But would be nice to avoid constant refactoring

libs/dh/wholesale/feature-calculations/src/lib/details/details.component.html
```diff
@@ -16,6 +16,8 @@
 -->
 @let guid = id();
 @let calculation = result();
+@let wholesaleAndEnergyCalculation =
```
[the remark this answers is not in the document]
ManBearTM: I think it made it worse, since then I have to come up with an additional variable name. So if it is okay, I would like to skip this part for now - but I have some ideas on automatic type narrowing that I want try, so I'll revisit this and see if I can make it better!

apps/dh/api-dh/source/DataHub.WebApi/Modules/Common/Types/WholesaleAndEnergyCalculationTypeEnum.cs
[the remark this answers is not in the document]
mimse: Its good ramblings :-) Think we need a company wide effort to align enums :-) maybe in 2026 👯

libs/dh/shared/data-access-graphql/src/lib/dh-graphql.providers.ts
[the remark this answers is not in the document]
mimse: Okay, thank you for the explanation ❤️

mimse (approved):

PR #3964, opened 2025-02-07
feat(dh): empty state for non-existing metering point

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/metering-point/feature-search/src/lib/dh-can-activate-metering-point-overview.ts
+++ b/libs/dh/metering-point/feature-search/src/lib/dh-can-activate-metering-point-overview.ts
@@ -20,6 +20,8 @@ import { ActivatedRouteSnapshot, CanActivateFn, Router } from '@angular/router';
 import { inject } from '@angular/core';
 
 import { BasePaths, getPath, MeteringPointSubPaths } from '@energinet-datahub/dh/core/routing';
+import { DoesMeteringPointExistDocument } from '@energinet-datahub/dh/shared/domain/graphql';
+import { query } from '@energinet-datahub/dh/shared/util-apollo';
 
 import { dhIsValidMeteringPointId } from './dh-metering-point-utils';
 import { dhMeteringPointIdParam } from './dh-metering-point-id-param';
@@ -28,12 +30,30 @@ export const dhCanActivateMeteringPointOverview: CanActivateFn = (
   route: ActivatedRouteSnapshot
 ) => {
   const meteringPointId: string = route.params[dhMeteringPointIdParam];
+  const isValidMP = dhIsValidMeteringPointId(meteringPointId);
+  const router = inject(Router);
 
-  return (
-    dhIsValidMeteringPointId(meteringPointId) ||
-    inject(Router).createUrlTree([
-      getPath<BasePaths>('metering-point'),
-      getPath<MeteringPointSubPaths>('search'),
-    ])
-  );
+  if (isValidMP) {
+    return query(DoesMeteringPointExistDocument, { variables: { meteringPointId } })
+      .result()
+      .then((result) => {
+        if (result.data === undefined) {
+          return router.createUrlTree(
+            [getPath<BasePaths>('metering-point'), getPath<MeteringPointSubPaths>('search')],
+            {
+              queryParams: {
+                [dhMeteringPointIdParam]: meteringPointId,
+              },
+            }
+          );
+        }
+
+        return true;
+      });
+  }
+
+  return router.createUrlTree([
+    getPath<BasePaths>('metering-point'),
+    getPath<MeteringPointSubPaths>('search'),
+  ]);
 };
--- a/libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
+++ b/libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
@@ -16,23 +16,26 @@
  * limitations under the License.
  */
 //#endregion
-import { Router } from '@angular/router';
-import { Component, inject } from '@angular/core';
+import { Component, inject, signal } from '@angular/core';
+import { ActivatedRoute, Router } from '@angular/router';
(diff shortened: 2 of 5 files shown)


libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
```diff
@@ -73,28 +77,56 @@ import { dhMeteringPointIdValidator } from './dh-metering-point.validator';
...
-  searchControl = new FormControl('', [dhMeteringPointIdValidator()]);
+  searchControl = new FormControl('', {
+    validators: [dhMeteringPointIdValidator()],
+    nonNullable: true,
+  });
+
+  mpNotFound = signal(false);
+
+  constructor() {
+    this.activatedRoute.queryParamMap.pipe(takeUntilDestroyed()).subscribe((params) => {
```
mimse, line 100: Suggestion: use a input to get the param value, like we do on other details pages.

libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
```diff
@@ -73,28 +77,56 @@ import { dhMeteringPointIdValidator } from './dh-metering-point.validator';
...
+      if (maybeMeteringPointId) {
+        this.mpNotFound.set(true);
+        this.searchControl.setValue(maybeMeteringPointId);
+      } else {
+        this.mpNotFound.set(false);
+        this.searchControl.reset();
+      }
+    });
+
+    this.searchControl.valueChanges
```
mimse, line 112: Suggestion: Use to signal to get the changed value

libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
```diff
@@ -73,28 +77,56 @@ import { dhMeteringPointIdValidator } from './dh-metering-point.validator';
...
 
     const meteringPointId = this.searchControl.getRawValue();
-
-    if (!meteringPointId) return;
-
     const result = await this.doesMeteringPointExist.query({ variables: { meteringPointId } });
 
-    if (!result.data?.meteringPoint.meteringPointId) return;
+    if (!result.data) {
+      this.mpNotFound.set(true);
```
mimse, line 126: Suggestion: use computed to compute if the meteringpoint was found or not

libs/dh/metering-point/feature-search/src/lib/dh-can-activate-metering-point-overview.ts
```diff
@@ -28,12 +30,30 @@ export const dhCanActivateMeteringPointOverview: CanActivateFn = (
...
-      getPath<BasePaths>('metering-point'),
-      getPath<MeteringPointSubPaths>('search'),
-    ])
-  );
+  if (isValidMP) {
+    return query(DoesMeteringPointExistDocument, { variables: { meteringPointId } })
+      .result()
+      .then((result) => {
+        if (result.data === undefined) {
+          return router.createUrlTree(
```
mimse, line 41: Issue: Looks like the function is returning either a string or a boolean?

mimse (changes requested):
A few comments :-)

libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
dzhavat: Good idea. Didn't know query params are also supported as `input`s but they do. Done now.

libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
dzhavat: This is a bit tricky because there are a multiple cases where `mpNotFound` can be set. I couldn't figure it out a good solution. Have kept it the same for now. Need to experiment some more. Also welcome to suggestions.

libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
```diff
@@ -43,10 +46,11 @@ import { dhMeteringPointIdValidator } from './dh-metering-point.validator';
     WattTextFieldComponent,
     WattFieldErrorComponent,
     WattButtonComponent,
+    WattEmptyStateComponent,
   ],
   styles: `
     .search-wrapper {
-      margin-top: 15rem;
+      margin: 15rem 0 6rem;
```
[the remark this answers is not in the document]
dzhavat: Not sure it's possible here because the `margin` values are custom and specific for this page only

libs/dh/metering-point/feature-search/src/lib/dh-can-activate-metering-point-overview.ts
dzhavat: Have added a return type to the function.

libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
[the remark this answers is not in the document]
mimse: If the margins do not line up with an known spacing from watt, maybe its a mistake, or maybe they should be added to our watt?

libs/dh/metering-point/feature-search/src/lib/dh-search.component.ts
```diff
@@ -73,28 +75,60 @@ import { dhMeteringPointIdValidator } from './dh-metering-point.validator';
...
+  private seachControlChange = toSignal(this.searchControl.valueChanges);
+
+  meteringPointId = input<string>();
+  meteringPointNotFound = signal(false);
+
+  constructor() {
+    effect(() => {
+      const maybeMeteringPointId = this.meteringPointId();
+
+      if (maybeMeteringPointId) {
```
mimse, line 103: Issue: if maybeMeteringPointId is something, meteringPointNotFound is set to true, is it not the other way around?

mimse (approved):

PR #3970, opened 2025-02-11
chore: upgrade Nx and Angular to latest stable

dzhavat:
### Workspace 
- bump Node.js to latest 20.x
- bump Cypress to latest stable
- bump Nx to latest stable
- bump Angular to latest stable
- bring back binary lock file instead of text-based
- add `bunfig.toml` file to configure install behaviour
- adjust Husky to support binary lock file

Note: Nx does NOT yet support text-based lock file. While I managed to upgrade the dependencies with text-based lock file, the `affected` command did not work properly afterwards. Some apps/libs were not detected as affected even though they were. That's why I decided to temporarily bring the binary lock file back until Nx adds support for text-based.

Bun 1.2 generates text-based lock file by default. The `buckfig.toml` file configures Bun to generate binary file instead. This way we can upgrade Bun and Nx until Nx adds support for text-based lock file.

### DataHub
- add CommonJS packages to `allowedCommonJsDependencies` config
- bump production budget
- fix error with mocks which prevented the app from building after bumping Node.js

## References

- #0000


mimse (dismissed):
Praise: Great work!

mimse (approved):

PR #3971, opened 2025-02-11
Added customer info

mimse:


libs/dh/metering-point/feature-overview/src/lib/dh-customer-overview.component.ts
```diff
@@ -67,46 +69,56 @@ import { DhCustomerContactDetailsComponent } from './dh-customer-contact-details
...
         @for (customer of customers(); track customer.id) {
           <div vater-flex gap="s" basis="0" class="customer">
+            <div
+              vater-stack
+              direction="row"
+              gap="s"
+              class="watt-space-inset-squish-s watt-space-stack-m"
+              [class.protected-address]="customer.isProtectedName"
+            >
+              @if (customer.isProtectedName) {
```
dzhavat, line 82: suggestion: let's move the `if` around the `div` instead. Otherwise it can end up with an empty element

libs/dh/metering-point/feature-overview/src/lib/dh-customer-overview.component.ts
```diff
@@ -67,46 +69,56 @@ import { DhCustomerContactDetailsComponent } from './dh-customer-contact-details
...
-  modalService = inject(WattModalService);
+  private modalService = inject(WattModalService);
+
+  meteringPoint = input.required<MeteringPointDetails | undefined>();
+
+  customers = computed(() =>
+    this.meteringPoint()?.currentCommercialRelation?.currentEnergySupplierPeriod?.contacts.filter(
+      (x) =>
+        x.relationType === CustomerRelation.Primary || x.relationType === CustomerRelation.Secondary
+    )
```
dzhavat, line 111: question: shouldn't there be a `[]` fallback in case of `undefined`? Strange that TS does not flag that.

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-details.component.ts
```diff
@@ -63,10 +63,6 @@ import { DhActualAddressComponent } from './dh-actual-address.component';
   `,
   template: `
     <watt-card *transloco="let t; read: 'meteringPoint.overview.details'">
-      <watt-card-title>
-        <h3>{{ t('title') }}</h3>
```
dzhavat, line 67: suggestion: let's remove the translation if it's not used anywhere else

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-status.component.ts
```diff
@@ -47,5 +47,6 @@ import { WattBadgeComponent } from '@energinet-datahub/watt/badge';
   `,
 })
 export class DhMeteringPointStatusComponent {
-  status = input.required<'CLOSED' | 'CONNECTED' | 'DISCONNECTED'>();
+  // 'CLOSED' | 'CONNECTED' | 'DISCONNECTED'
+  status = input.required<string>();
```
dzhavat, line 51: suggestion: looks like "UNKNOWN" status is also sent in. Let's handle that too.

libs/dh/metering-point/feature-overview/src/lib/dh-customer-overview.component.ts
mimse: I put it there, to avoid the text jumping up if customer name is not protected

libs/dh/metering-point/feature-overview/src/lib/dh-customer-overview.component.ts
mimse: Good idea

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-status.component.ts
mimse: This will become a enum at some point, will do it there :-)

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-details.component.ts
mimse: Removed

libs/dh/metering-point/feature-overview/src/lib/dh-customer-overview.component.ts
dzhavat: Can we keep the element more badge-looking (not taking the full width caused by the `flexbox` parent)
`self-align: flex-start` should do it.

image

dzhavat (approved):

PR #3989, opened 2025-02-14
feat(dh): reactive query options for util-apollo

ManBearTM:
I have noticed an increasing amount of cases where we use a `lazyQuery` (or a `query` with `skip: true`), only to call it immediately in a reactive context (such as an effect) - usually because the variables depend on other reactive variables, such as input signals. Being inspired by the new `resource` function from Angular and how it deals with reactivity, I decided to extend the functionality of `query` (and `lazyQuery`, although not sure how useful it is there) to also accept a function as the second parameter (options). Just like `request` in `resource`, this function tracks dependencies, so you can use all kinds of signals (including inputs) and the options object should automatically update and, if not skipped, also automatically trigger a refetch. I hope this should reduce the amount of times we have to call `refetch` manually. Example:

```ts
id = input.required<string>(); 
calculationQuery = query(GetCalculationByIdDocument, () => ({
  variables: {
    id: this.id()
  }
}));
```

I initially tried to fully embrace the `resource` API to see if I could utilize that in the `query` function itself, but due to several limitations in that API, I decided to not go forward with that. For example, the `resource` function only supports emitting "once" per "request", but we often emit multiple times from the same query with the same variables (refetch queries, subscriptions, cache updates). If Angular delivers something cool related to resources and router hoisting, I might revisit this and perhaps create a lightweight "queryResource" utility with only a subset of the features of query.

You may also notice that I removed `queryTime` from the `query` implementation. I think it added unnecessary complexity that could instead be moved to a separate helper function. That is why I have created the `queryTime` utility function and used it in the two places that show query time in tables. But this means that the global query time logging is now gone. If you REALLY REALLY miss it, then I will look for an alternative solution. Otherwise, with this new helper I think it is very easy to add it to queries on a case-by-case basis:

```ts
calculationsQuery = query(GetCalculationsDocument);
calculationsQueryTime = queryTime(this.calculationsQuery);
logQueryTimeEffect = effect(() => {
  console.log(this.calculationsQueryTime());
});
```

Or, you know, use the devtools in the browser :P


mimse (commented):
Praise: Reactive query very nice!
Praise: Querytime as opt in very nice!
Comment: I dont know why, but referencing the query directly in the template seems more "noisy" then creating variables for loading and error

libs/dh/shared/util-apollo/src/lib/performance/queryTime.ts
```diff
@@ -0,0 +1,66 @@
...
+    source: query.status,
+    computation: (_, prev) => {
+      query.status(); // tracked dependency
+      const startMarks = performance.getEntriesByName(startMark);
+      const endMarks = performance.getEntriesByName(endMark);
+      return startMarks.length > 0 && startMarks.length === endMarks.length
+        ? performance.measure(measureName, startMark, endMark).duration
+        : prev?.value;
+    },
+  }).asReadonly();
```
dzhavat, line 65: praise: beautiful 🥰

ManBearTM:
> Praise: Reactive query very nice! Praise: Querytime as opt in very nice! Comment: I dont know why, but referencing the query directly in the template seems more "noisy" then creating variables for loading and error

Great to hear, thanks! Valid input, I don't have strong feelings about it, just personal preference. But maybe we should settle on some standard here 🙂

dzhavat (dismissed):

mimse (approved):

dzhavat (approved):

PR #3994, opened 2025-02-18
feat(dh): adjustments to metering point overview

dzhavat:
### DataHub
- change translations based on fedback
- minor adjustment to displaying address in card
- show address next to metering point id

## References

- #0000


libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-overview.component.ts
```diff
@@ -96,7 +98,8 @@ import { GetMeteringPointByIdDocument } from '@energinet-datahub/dh/shared/domai
     <dh-result [hasError]="hasError()" [loading]="loading()">
       <div *transloco="let t; read: 'meteringPoint.overview'" class="page-header">
         <h2 vater-stack direction="row" gap="m" class="watt-space-stack-s">
-          {{ meteringPointId() }}
+          {{ meteringPointId() }} •
+          <ng-container *ngTemplateOutlet="addressTmpl" />
```
mimse, line 102: Question: Why the templateOutlet? It does not seem to be reused anywhere?

dzhavat: No other reason but to move the logic away from the `h2`. It would've gotten unnecessarily complex otherwise.

dzhavat: That is also an options. Unfortunately it can't be reused in the `dh-metering-point-details.component.ts` because the `postCode` + `cityName` are on new line there, whereas these two are on the same line here.

But a new component is still an option even though it's going to be used only once.

mimse: Keep the outlet, I´m just not a big fan :D

mimse (approved):

PR #3997, opened 2025-02-19
Sender cannot be empty

Sejruppen:
Sets the "showResetOption" to false, to remove the empty choice
image


sahma19 (approved):
godt arbejde

PR #4001, opened 2025-02-19
Chore/add mock data to dev example

mimse:


libs/dh/shared/data-access-mocks/src/lib/process/data.ts
```diff
@@ -0,0 +1,126 @@
...
+    calculationType: CalculationType.Aggregation,
+    createdAt: new Date('2022-01-01T00:00:00Z'),
+    state: ProcessState.Running,
+    steps: [
+      {
+        __typename: 'OrchestrationInstanceStep' as const,
+        isCurrent: true,
+        state: ProcessStepState.Running,
+      },
+    ],
```
dzhavat, line 43: suggestion: the `as const` casting looks a bit off. If we want to keep the `export default` pattern, then it's better to assign the array to a typed constant and export it afterwards. Or was there another reason for using `as const`?

```ts
const data: ElectricalHeatingCalculation[] = [ ... ];

export default data;
```

libs/dh/dev-examples/feature-processes/tsconfig.lib.json
```diff
@@ -7,6 +7,11 @@
...
   },
-  "exclude": ["src/**/*.spec.ts", "src/test-setup.ts", "jest.config.ts", "src/**/*.test.ts"],
-  "include": ["src/**/*.ts"]
+  "exclude": ["tests/**/*.spec.ts", "tests/test-setup.ts", "jest.config.ts", "tests/**/*.test.ts"],
+  "include": [
+    "src/**/*.ts",
+    "index.ts",
+    "mocks/**/*.ts",
+    "../../shared/data-access-mocks/src/lib/process/index.ts"
+  ]
```
dzhavat, line 16: issue: revert changes here

libs/dh/dev-examples/feature-processes/src/components/details.component.ts
```diff
@@ -94,7 +94,7 @@ import { DhCalculationsDetailsGridAreasComponent } from './gridareas.component';
           @if (loading()) {
             {{ t('loading') }}
           } @else if (calculationType()) {
-            {{ 'calculationTypes.' + calculationType() | transloco }}
+            {{ 'shared.calculationTypes.' + calculationType() | transloco }}
```
dzhavat, line 97: question: is there still a need for `devExamples.processes.calculationTypes` if these are looked up in `shared` now.

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -1980,7 +1999,9 @@
         "calculationId": "Beregnings-ID",
         "executionType": "Afviklingstype",
         "executionTime": "Afviklingstidspunkt",
-        "gridAreas": "Netområder"
+        "gridAreas": "Netområder",
+        "meteringPointType": "{{ meteringPoint.overview.shared.meteringPointType }}",
```
dzhavat, line 2003: question: is this necessary? It's found here but not in the English translation file.

libs/dh/shared/data-access-mocks/src/lib/process/data.ts
mimse: It is to make typescript happy. If __typename is not a const. Typescript fails to infer the right type after the data is imported.

libs/dh/globalization/assets-localization/src/assets/i18n/da.json
mimse: Added it to the english versino as well

libs/dh/shared/data-access-mocks/src/lib/process/data.ts
mimse: Fixed

PR #4003, opened 2025-02-20
fix(watt): make drawer great again

ManBearTM:


libs/watt/src/lib/components/drawer/watt-drawer.component.ts
```diff
@@ -57,28 +57,48 @@ export type WattDrawerSize = 'small' | 'normal' | 'large';
...
+  autoOpen = input(false, { transform: booleanAttribute });
 
-  /** Whether the drawer should show a loading state.  */
+  /**
+   * Used to track the current drawer when reusing the same drawer instance to
+   * render different content. This is required when interactions outside the
+   * drawer should result in updating the drawer's content instead of closing it.
+   */
+  // Transform input to a string (omitting input should not be the same as passing undefined)
+  key = input<unknown>();
```
dzhavat, line 85: question: is there supposed to be a `transform` here? The comment sounds as if there is.

libs/watt/src/lib/components/drawer/watt-drawer.component.ts
```diff
@@ -110,51 +130,40 @@ export class WattDrawerComponent implements OnDestroy {
...
-    // Disable click outside check until the current event loop is finished.
-    // This might seem hackish, but the order of execution is stable here.
-    this.bypassClickCheck = true;
-    setTimeout(() => {
-      this.bypassClickCheck = false;
-    }, 0);
-
-    if (!this.isOpen()) {
+    // Prevent tracking drawer signals when open is triggered in a reactive context
+    untracked(() => {
```
dzhavat, line 136: suggestion: let's make the comment more explicit by adding some examples of "reactive context". I can only think of `effect` and `afterRenderEffect`. I hope noone calles `open` from a `computed`.

libs/watt/src/lib/components/drawer/watt-drawer.component.ts
```diff
@@ -57,28 +57,48 @@ export type WattDrawerSize = 'small' | 'normal' | 'large';
...
   /** Used to adjust drawer size to best fit the content. */
   size = input<WattDrawerSize>('normal');
 
-  /** Whether the drawer is open.  */
-  isOpen = model(false);
+  /**
+   * Whether the drawer should open automatically. If `key` is provided and
+   * `autoOpen` is true, the drawer will open every time the key changes.
+   */
+  autoOpen = input(false, { transform: booleanAttribute });
```
dzhavat, line 77: question: the `input` is boolean. Is it still necessary to have the `booleanAttribute` then?

libs/watt/src/lib/components/drawer/watt-drawer.component.ts
dzhavat, line 1: issue: looks unrelated to this PR (exists on main as well) but can we fix the console error in Storybook?

image

dzhavat, line 1: issue: can we adjust the stories so `isOpen` is called? 

"storybook-drawer-loading.component.ts"
image

"watt-drawer.stories.ts" (multiple places)

image

dzhavat (commented):
praise: nice work adding support for auto open 🚀 
It will save us a lot of manual setup.

libs/watt/src/lib/components/drawer/watt-drawer.component.ts
ManBearTM: Oh yeah good catch, I removed that. Turns out it wasn't needed.

da81af762588ea1f30f10a255623f7fb6c4bc454

libs/watt/src/lib/components/drawer/watt-drawer.component.ts
ManBearTM: Not strictly necessary, no, but I like that you can omit the boolean (`<watt-drawer autoOpen />` instead of `<watt-drawer [autoOpen]="true" />`). It also coerces any value to boolean, so you can do fancy stuff (perhaps too fancy 😄) like this:

```html
<watt-drawer [autoOpen]="id()" />
```

Resulting in a drawer that only "auto opens" if the id is truthy

libs/watt/src/lib/components/drawer/watt-drawer.component.ts
ManBearTM: f8838c06c4701cbcdbc4ed36fbf724bf031818b7

libs/watt/src/lib/components/drawer/watt-drawer.component.ts
ManBearTM: Improved it a bit: c5b5b1002a4fcb18989946f65308e547fdb291a0

dzhavat (approved):

PR #4004, opened 2025-02-20
fix(dh): requests v2 sorting issues

ManBearTM:


apps/dh/api-dh/source/DataHub.WebApi/Modules/ProcessManager/Requests/Types/ActorRequestQueryResultType.cs
```diff
@@ -30,80 +28,10 @@ protected override void Configure(
 
         descriptor
             .Field("calculationType")
-            .Resolve(c =>
```
mimse, line 33: Worry: We had issues with debugging logic moved to an extension method before, because it hides the complexcity. But it does look better.. not sure about the trade offs here.

mimse (approved):
Have a comment, but will approve :-)

apps/dh/api-dh/source/DataHub.WebApi/Modules/ProcessManager/Requests/Types/ActorRequestQueryResultType.cs
ManBearTM: oh yeah, that was the client extension methods right? we have moved away from that at least, with the new modules structure enabling custom clients that wrap other clients, instead of using extension methods. but i don't disagree that it can hide complexity, so we should use it with caution. the reason i opted to use it here was that i needed to reuse the code in my sort input, so it was either an extension method or a static method in a class somewhere. but you cannot use regular method calls in these "lambda" expressions, but for some reason extension methods are fine 🤷

Issue #4010, opened 2025-02-20
e2e tests should catch problems with the bff


ManBearTM:
This issue has been created in light of a recent incident caused by a bad infrastructure PR. This PR caused all requests to the BFF to fail, but because the frontend was deployed successfully and the user could still change language (which there is an E2E test for), the pipeline thought everything was fine and proceeded to deploy to production. It would be nice if there was a test that would fail if the BFF is misbehaving.

PR #4021, opened 2025-02-24
fix(e2e-dh): check market participant name in test

dzhavat:
### DataHub
- check for existence of market participant name in test
- update dependencies
- set `viewportWidth` to 1280 in config

## References

This PR introduces a quick solution to #4010 until a better approach is agreed upon


ManBearTM (approved):
🥇

PR #4058, opened 2025-03-06
refac: split .gitignore

ManBearTM:
Had some issues with C# related ignores interfering with "frontend" related stuff, so I decided to move all the C# specific stuff out of the global .gitignore file and into a separate .gitignore that is scoped to `api-dh`.


.gitignore
```diff
@@ -12,375 +12,44 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-## Ignore Visual Studio temporary files, build results, and
-## files generated by popular Visual Studio add-ons.
-##
-## Get latest from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
+# Poor mans monorepo
+backend/*
```
mimse, line 16: Issue: This is not scoped in api-dh but the root
image

mimse (changes requested):
Good idea, just one issue i see :-)

.gitignore
ManBearTM: but this is the global .gitignore file, so it is fine right? 🙂

mimse (approved):

PR #4063, opened 2025-03-06
refac(watt): prepare for npm

ManBearTM:
This PR only moves things around within watt (plus some minor adjustments to names/folders). The change is needed since `ng-packagr` expects a certain folder structure (at least if you want a nice output). So in order to meet the requirements, the contents of the watt library has been "flattened" under a single folder called "package". Each sub folder here corresponds to a sub entry point in the coming watt npm package. As a result, we no longer have the separation of "foundations", "components", "utils" etc, but instead most of "foundations" has been moved into "package/core" and most of "utils" has been moved into "package/utils".


dzhavat (approved):
praise: great, great success! 💯💯

image

mimse (approved):
Praise: Great work

Sejruppen (approved):

PR #4070, opened 2025-03-10
refac(watt): inline assets

### LINKED ISSUE #3130: refactor icon to use signals
Refactor ./watt/src/lib/foundations/icon/icon.component.ts to use the new signal syntax

ManBearTM:
In an effort to avoid requiring apps to serve assets from `node_modules` I have created this PR that "inlines" the few custom SVG assets from Watt. Here is a list of changes:

- `custom-map-marker` removed (it wasn't referenced anywhere)
- `custom-primary-info` removed (it wasn't referenced anywhere)
- `custom-meter` removed (it wasn't referenced anywhere)
- `watt-icon` refactored, removing the icon service and instead supporting (svg) components as `ng-content`
- custom flag icons converted to components (will become a sub package, "watt/icons/flags")
- "empty state" icons (`explore`, `no-results`, `power`) converted to components and moved under `empty-state`
- `account` icon removed, since it was the same reference as `user`
- `custom-assignment-add` moved into EO, since it seemed very specific to EO

Closes #3130


mimse (approved):
Praise: nice refactor

Sejruppen (approved):

PR #4077, opened 2025-03-12
refac: move watt-code highlight worker to app

ManBearTM:
Yet another refactoring because of "npm watt". This time, I realized it was kinda hard to bundle a service worker with a library. Things I considered:

- Inline web worker using blob url
  -  Cool trick, but got complicated when I had to include the `highlight.js` dependency
- Removing the worker altogether
  - I didn't like the idea of removing a feature just so that we can publish a package
- Try to bundle it anyway using a bunch of tricks like compiling and copying the worker file
  - It felt very complicated and fragile and not like a solution I'd like to maintain 

So in the end I settled on this solution, where I use an injection token to (optionally) provide a worker factory to the WattCode component. If none is provided, the code is simply not highlighted. This also has the added benefit that we can customize which languages to include from `highlight.js` without worrying about if other consumers of watt needs different packages.


dzhavat (approved):
praise: super nice solution with the `InjectionToken`. Good job 💯🏆

dzhavat (approved):

PR #4088, opened 2025-03-18
Watt ett stepper editable

Sejruppen:
## Description
This PR enables a bit more control of the Stepper, to enable blocking going back through the stepper in a couple of ways:
- Disable Ripple (Stepper); can be used for removing the effect, if one prefers the way it looks and feels. I  use it in ETT to make the steps appear not to be clicked.
- Editable (Step); Custom boolean for setting whether or not a step can be returned to. FYI this is just a thing mat-stepper already supports.
- Disable Previous Button; As name indicates, it is a copy of disable next button. A way to set whether or not the button is disabled.

I have implemented new stories in storybook to reflect these 3 additions and I have implemented them in ETT.

tanoback


mimse (dismissed):
Praise: nice addition 👍🏻

dzhavat (approved):

dzhavat (approved):

PR #4090, opened 2025-03-18
feat(dh): change permission that is allowed to access metering point

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-search/src/lib/dh-metering-point.routes.ts
```diff
@@ -31,7 +31,7 @@ export const dhMeteringPointRoutes: Routes = [
   {
     path: '',
     canActivate: [
-      PermissionGuard(['fas']),
+      PermissionGuard(['metering-point:search']),
```
mimse, line 34: Issue: Should it be both fas and metering-point:search that should have access?

dzhavat: Good question. Let's discuss it tomorrow :)

mimse (approved):

PR #4118, opened 2025-03-21
Update Process Manager package and fix breaking change for migrated calculations

ebbeknudsen:
## Description

Models for migrated Wholesale calculations has been removed from the Process Manager, but we still need it to be able to get the correct calculation id from migrated calculations.


dstenroejl (approved):
LGTM

PR #4124, opened 2025-03-25
refac: remove importProvidersFrom for MatDialog and MatSnackBar

ManBearTM:
https://dev.to/this-is-angular/you-dont-need-importprovidersfrom-with-angular-material-3nih


dzhavat (approved):
praise: nice refactor. One/two things less to worry about :)

PR #4127, opened 2025-03-25
refac: replace angular devkit with nx angular

ManBearTM:


mimse (commented):
Issue: looks like it uses webpack out of the box and not esbuild, that we are using now

PR #4131, opened 2025-03-26
feat(dh): hide energy supplier card based on type

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/metering-point/feature-overview/src/lib/dh-can-see.directive.ts
+++ b/libs/dh/metering-point/feature-overview/src/lib/dh-can-see.directive.ts
@@ -1,0 +1,99 @@
+//#region License
+/**
+ * @license
+ * Copyright 2020 Energinet DataHub A/S
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License2");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


libs/dh/metering-point/feature-overview/src/lib/dh-can-see.directive.ts
```diff
@@ -0,0 +1,99 @@
...
+const dhWhoCanSeeWhatMap: {
+  [k in PropertyName]: {
+    marketRoles: EicFunction[];
+    meteringPointTypes: ElectricityMarketMeteringPointType[];
+  };
+} = {
+  'energy-supplier-card': {
+    marketRoles: [EicFunction.DataHubAdministrator],
+    meteringPointTypes: [
+      ElectricityMarketMeteringPointType.Consumption,
```
mimse, line 95: Suggestion: If all types are should see the field, maybe we should allow an empty list to see the field as well.
Then we dont need add very possible options

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point-master-data.component.ts
```diff
@@ -94,10 +91,7 @@ import { EnergySupplier } from './types';
         <dh-customer-overview [meteringPointDetails]="meteringPointDetails()" />
 
         <dh-energy-supplier
-          *dhCanSeeValue="
-            [EicFunction.DataHubAdministrator];
-            isResponsible: isEnergySupplierResponsible()
-          "
+          *dhCanSee="'energy-supplier-card'; meteringPointDetails: meteringPointDetails()"
```
mimse, line 94: Praise: Much better 💯

mimse (changes requested):

libs/dh/metering-point/feature-overview/src/lib/dh-can-see.directive.ts
dzhavat: Good idea. I'd like to postpone implementing it until there's a use case for it.

mimse (approved):

PR #4134, opened 2025-03-27
feat(dh): adjustments to metering point master data

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/metering-point/feature-overview/src/lib/dh-can-see.directive.spec.ts
+++ b/libs/dh/metering-point/feature-overview/src/lib/dh-can-see.directive.spec.ts
@@ -1,0 +1,136 @@
+//#region License
+/**
+ * @license
+ * Copyright 2020 Energinet DataHub A/S
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License2");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


libs/dh/metering-point/feature-overview/src/lib/dh-can-see.directive.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+function getPropertyName(propertyName: PropertyName) {
+  return propertyName;
+}
+
+type Params = {
+  canSeeProperty: PropertyName;
+  mpType: ElectricityMarketMeteringPointType;
+};
+
+describe(DhCanSeeDirective, () => {
```
mimse, line 40: Praise: nice with tests here

mimse (approved):
Praise: Like the simplicity of the cansee directive you have made.

libs/dh/metering-point/feature-overview/src/lib/dh-can-see.directive.spec.ts
dzhavat: It's not much but at least a start. More tests can be added if we find a bug where something that shouldn't be shown is displayed. We have a starting point now.

PR #4135, opened 2025-03-28
feat(watt): drawer appear animation

ManBearTM:


libs/dh/shared/data-access-graphql/src/lib/error-handler.ts
```diff
@@ -19,7 +19,6 @@
 import { onError } from '@apollo/client/link/error';
 
 import { DhApplicationInsights } from '@energinet-datahub/dh/shared/util-application-insights';
-import { translate } from '@jsverse/transloco';
```
mimse, line 22: Strange, that it didnt get caught by the lint

mimse (commented):
Praise: nice refactor

dzhavat (approved):
praise: works like a charm ... and it's now way more easier to animate drawer across features/products. Nice work!

libs/dh/shared/data-access-graphql/src/lib/error-handler.ts
dzhavat: Apparently it did ... but a little too late.

On "main" 👇

image

PR #4162, opened 2025-04-04
Clear session storage if login fails

mimse:
Sometimes, the login would fail and you would not be redirected to the login page. This fixes that issue.


mimse:
E2E not happy, closing for now

PR #4167, opened 2025-04-07
fix(dh): persist breadcrumb in matering point page

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/metering-point/feature-overview/src/lib/dh-metering-point.component.ts
+++ b/libs/dh/metering-point/feature-overview/src/lib/dh-metering-point.component.ts
@@ -16,11 +16,9 @@
  * limitations under the License.
  */
 //#endregion
-import { NavigationEnd, Router } from '@angular/router';
+import { Router } from '@angular/router';
 import { Component, computed, effect, inject, input } from '@angular/core';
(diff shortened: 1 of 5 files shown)


libs/dh/metering-point/feature-overview/src/lib/dh-metering-point.component.ts
```diff
@@ -142,9 +144,14 @@ export class DhMeteringPointComponent {
   isEnergySupplierResponsible = computed(() => this.meteringPointDetails()?.isEnergySupplier);
 
   breadcrumbLabel = translateSignal('meteringPoint.breadcrumb');
+  onNavigationEnd_EffectTrigger = toSignal(
```
mimse, line 147: Suggestion: if its possible, maybe use the navigation ended signal from the breadcrumb service?

mimse (changes requested):

libs/dh/metering-point/feature-overview/src/lib/dh-metering-point.component.ts
dzhavat: Looks like it's possible 👍

mimse (approved):

PR #4169, opened 2025-04-08
Feature/date format

mimse:


libs/dh/metering-point/feature-overview/src/components/measurement/dh-measurements.component_v2.ts
```diff
@@ -0,0 +1,154 @@
...
+        *transloco="let resolveHeader; read: 'meteringPoint.measurements.columns'"
+        [resolveHeader]="resolveHeader"
+        [columns]="columns()"
+        [dataSource]="dataSource"
+        [loading]="query.loading()"
+        sortDirection="desc"
+        [sortClear]="false"
+      >
+        <ng-container *wattTableCell="columns().observationTime; let element">
+          {{ this.formatObservationTime(element.observationTime, element.current.resolution) }}
```
dzhavat, line 71: suggestion: we should consider creating a pipe for this because the function is called quite a lot - on initial load as well as on every change detection cycle. Can be a separate PR

mimse: Good idea, will have a look at it in the next PR 👍🏻

dzhavat (approved):

PR #4179, opened 2025-04-09
feat(dh): add current metering point in list of relateed metering points, add hover state

dzhavat:
## References

- #0000


mimse (approved):
Praise: looks great 💯

PR #4184, opened 2025-04-11
feat(watt): YearMonthField

ManBearTM:
This PR adds a new WattYearMonthField component using the same approach as WattDateTimeField. There is probably some duplication, but I'd rather have some duplication over the wrong abstraction here (like the current pickers). So, for now, there is no "base" picker logic or anything like that. 

Notice: This uses the popover API + anchor positioning, which is currently not fully supported in Firefox (the positioning of the popover may be weird when scrolling the page).

yearmonth-field


Sejruppen (approved):
praise: Overall really nice ! 😃 - I especially like the way you have made the "YearMonth" class to wrap some of the date logic

libs/watt/package/yearmonth-field/watt-yearmonth-field.component.ts
```diff
@@ -0,0 +1,178 @@
...
+  label = input('');
+
+  /** The minimum selectable date. */
+  min = input<Date>();
+
+  /** The maximum selectable date. */
+  max = input<Date>();
+
+  /** Emits when the selected month has changed. */
+  monthChange = outputFromObservable(this.valueChanges);
```
mimse, line 145: Question: What use case would you need an event on month changed? Would you not use the formcontrols valuechanged event?

mimse (approved):
Praise: Nice Fing work! 💯

libs/watt/package/yearmonth-field/watt-yearmonth-field.component.ts
ManBearTM: Yes usually, this just makes it possible to use the component without a `FormControl` for simpler use cases. Also handy if you need to trigger something directly from the template in response to the change event. It is just one line of code that adds some convenience and I see the same pattern in some of the Angular Material components, which is why I included it.

ManBearTM:
> praise: Overall really nice ! 😃 - I especially like the way you have made the "YearMonth" class to wrap some of the date logic

Appreciated! I struggled a bit with how to write the date logic, since there was a lot of back and forth between the different formats and types. So when I encounter these challenges I try to take a more "data-driven" approach which usually results in some pretty clean encapsulation IMHO.

Sejruppen (approved):

PR #4186, opened 2025-04-11
use year month field

mimse:


libs/dh/metering-point/feature-overview/src/components/measurement/dh-measurements-month.component.ts
```diff
@@ -103,8 +103,8 @@ export class DhMeasurementsMonthComponent {
   private sum = computed(
     () => `${this.formatNumber(this.measurements().reduce((acc, x) => acc + x.quantity, 0))}`
   );
-  maxDate = dayjs().subtract(1, 'days').toDate();
-  date = this.fb.control<Date>(this.maxDate);
+  maxDate = dayjs().subtract(1, 'days');
+  yearMonth = this.fb.control<string>(this.maxDate.format('YYYY-MM'));
```
ManBearTM, line 107: suggestion: maybe in another PR, but perhaps we should consider exporting the "model format" from the "yearmonth-field", so we can use that instead of "hardcoding" the string here? 🙂

ManBearTM (approved):
just a comment, feel free to resolve it, we can do it later

libs/dh/metering-point/feature-overview/src/components/measurement/dh-measurements-month.component.ts
mimse: Good idea !

ManBearTM (approved):

PR #4190, opened 2025-04-15
feat(dh): YearMonth scalar

ManBearTM:
This PR makes it possible to use `NodaTime.YearMonth` directly in GraphQL types. It will serialize to a string in the format of `YYYY-MM` (e.g. `2025-03`). The same goes for input types, so if used as input in a query or mutation, just use a string in the `YYYY-MM` format. This conveniently matches the format of the new `WattYearMonthField`.


mimse (approved):
Praise: looks good 👍

PR #4196, opened 2025-04-22
feat(dh): initial day view drawer

dzhavat:
## References

- #0000


mimse (commented):
Should we have a route for the drawer?

dzhavat:
> Should we have a route for the drawer?

For now, I don't think so. If we need to do that, then the selected day and position will need to be part of the url. e.g. 

`.../metering-point/222222222222222222/measurements/day/2025-04-22/4`

or some other variation. Let's wait until there's a need for it.

mimse (approved):

PR #4199, opened 2025-04-22
Feature/update measurement client and update data model

mimse:


libs/dh/metering-point/feature-search/src/routes.ts
```diff
@@ -68,7 +66,6 @@ export const dhMeteringPointRoutes: Routes = [
               import('@energinet-datahub/dh/metering-point/feature-overview').then(
                 (m) => m.DhMeasurementsNavigationComponent
               ),
-            canMatch: [() => inject(DhFeatureFlagsService).isEnabled('measurements-v2')],
```
dzhavat, line 71: suggestion: a `canActivate` guard is still needed to prevent manual route navigation.

dzhavat (approved):

PR #4201, opened 2025-04-23
feat(dh): adjustments to historic measurements

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-overview/src/utils/dh-format-number.ts
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+//#endregion
+import { formatNumber } from '@angular/common';
+
+import { WattSupportedLocales } from '@energinet-datahub/watt/date';
+
+export function dhFormatNumber(value: number | string, locale: WattSupportedLocales): string {
```
mimse, line 23: Suggestion: So its not used for anything else, maybe we should call it formatMeasurementNumber

mimse (changes requested):

libs/dh/metering-point/feature-overview/src/utils/dh-format-number.ts
dzhavat: Good point. I've pushed a new commit.

mimse (approved):

PR #4211, opened 2025-04-24
An optional icon can now be displayed next to the modal title.

Supelana:


libs/watt/package/modal/watt-modal.component.html
```diff
@@ -26,7 +26,12 @@
       />
     }
     @if (title && loading === false) {
-      <h2 class="watt-modal-title">{{ title }}</h2>
+      <h2 class="watt-modal-title">
+        @if (titleIcon) {
+          <watt-icon [name]="titleIcon" [size]="'l'" />
+        }
+        {{ title }}
+      </h2>
```
dzhavat, line 34: issue: the title's styling needs to be adjusted as well because the icon has `display: flex` which causes the text to break on a new line.

image

PR #4226, opened 2025-04-25
fix(watt): only use label for watt-field when it is not in chipMode

ManBearTM:
`watt-chip` has its own `<label>` and label elements are not allowed to be nested inside one another. I found this when I encountered a bug with a `watt-field` containing two choice chips. When I would click on the label text for the field or anywhere inside the field container, the selection would change to the first radio button. This is not the desired behavior, as any misclick could cause the selection to change.


dzhavat (approved):
praise: nice find! ... and fix :)

dzhavat (approved):

PR #4238, opened 2025-04-29
feat(dh): Relax QueryOptions for skipped queries

ManBearTM:
There are cases when you want to supply some default variables to a query that has `skip: true`, but due to the way the types were set up, you could not provide any variables without also providing the "required" variables (as defined in the schema). This PR relaxes the QueryOptions type in case you explicitly provide `skip: true`.


mimse (approved):
Magic!

PR #4282, opened 2025-05-13
Chore/update client v6

mimse:
Added all year view as well


libs/dh/metering-point/feature-overview/src/components/measurement/dh-measurements-all-years.component.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  columns: WattTableColumnDef<AggregatedMeasurementsForAllYears> = {
+    month: {
+      accessor: 'year',
+      size: 'min-content',
+      footer: { value: signal(this.transloco.translate('meteringPoint.measurements.sum')) },
+    },
+    currentQuantity: {
+      accessor: 'quantity',
+      align: 'right',
```
dzhavat, line 103: question: wouldn't this push the text all the way to the right? Maybe there needs to be a "spacer" column.

dzhavat (dismissed):

libs/dh/metering-point/feature-overview/src/components/measurement/dh-measurements-all-years.component.ts
mimse: You are right, fixed now

dzhavat (approved):

PR #4292, opened 2025-05-14
feat(dh): initial measurement reports overview

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/reports/feature-measurement-reports/src/components/measurement-reports.component.ts
+++ b/libs/dh/reports/feature-measurement-reports/src/components/measurement-reports.component.ts
@@ -16,9 +16,9 @@
  * limitations under the License.
  */
 //#endregion
-import { Component, inject } from '@angular/core';
+import { Component, computed, inject } from '@angular/core';
 import { TranslocoDirective } from '@jsverse/transloco';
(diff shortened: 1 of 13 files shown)


libs/dh/reports/feature-measurement-reports/src/components/measurement-reports.component.ts
```diff
@@ -16,27 +16,125 @@
...
+    numberOfGridAreasInReport: { accessor: 'numberOfGridAreasInReport' },
+    status: { accessor: 'statusType' },
+  };
+
+  displayedColumns = Object.keys(this.columns);
+
+  dataSource = new WattTableDataSource([]);
+
+  constructor() {
+    this.permissionService
```
mimse, line 131: Suggestion: Could we use toSignal here?

mimse (dismissed):

libs/dh/reports/feature-measurement-reports/src/components/measurement-reports.component.ts
dzhavat: Yes. A fix was just pushed.

mimse (approved):

PR #4312, opened 2025-05-19
feat(dh): empty state for measurements reports and basic overview

dzhavat:
## References

- #0000


libs/dh/reports/feature-measurement-reports/src/components/measurement-reports.component.ts
```diff
@@ -16,111 +16,94 @@
...
-        >
-          --
-        </ng-container>
+              <vater-spacer />
 
-        <ng-container
-          *wattTableCell="columns['gridAreas']; header: t('columns.gridAreas'); let entry"
-        >
-          @let gridAreas = entry.gridAreas;
+              <dh-new-report-request />
```
mimse, line 89: Suggestion: Should this not be moved into the data table as an action. Data table has a count as well

libs/dh/reports/feature-measurement-reports/src/components/measurement-reports.component.ts
```diff
@@ -16,111 +16,94 @@
...
+    }
+  `,
+  template: `
+    <watt-card vater inset="ml" *transloco="let t; read: 'reports.measurementReports'">
+      @if (isLoading()) {
+        <vater-stack fill="vertical" justify="center">
+          <watt-spinner />
+        </vater-stack>
+      } @else {
+        @if (totalCount() === 0) {
```
mimse, line 69: Suggestion: Maybe we could extend dhresult with a the possibility to render a action component if it has an empty state?

libs/dh/reports/feature-measurement-reports/src/components/measurement-reports.component.ts
```diff
@@ -16,111 +16,94 @@
...
-
-  dataSource = new WattTableDataSource([]);
-
-  displayedColumns = computed(() => {
-    const tableColumns = Object.keys(this.columns);
-    const isFas = this.isFas();
-
-    return isFas ? tableColumns : tableColumns.filter((column) => column !== 'actorName');
+  private readonly measurementsReportsQuery = query(GetMeasurementsReportsDocument, {
+    fetchPolicy: 'network-only',
```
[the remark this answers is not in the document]
dzhavat: We have the same under settlement reports. I think it's because we always want to show the latest reports state and not cached when someone comes back to the page after being in another page.

libs/dh/shared/domain/src/lib/dh-measurements-report.ts
```diff
@@ -0,0 +1,27 @@
+//#region License
+/**
+ * @license
+ * Copyright 2020 Energinet DataHub A/S
+ *
```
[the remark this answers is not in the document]
dzhavat: Done

libs/dh/reports/feature-measurement-reports/src/components/measurement-reports.component.ts
dzhavat: The PR moves away from Watt data table because Pernille specifically wanted the empty state to look like the one in Settlement reports.

mimse: Maybe i could extend data table with the empty state -> action button pattern? if that something we are going to use in multiple places

dzhavat: I think it'll require a lot of effort to make the `watt-data-table` support a completely clean empty view similar to Settlement reports. That component displays a few things by default that will need to be toggled, which will increase its complexity.

mimse (approved):

PR #4313, opened 2025-05-20
Release toggles v1

frontend-specialisten:
## Description

- Add MVP of release toggle implementation


libs/dh/shared/release-toggle/src/lib/dh-release-toggle.directive.spec.ts
```diff
@@ -0,0 +1,448 @@
...
+      mockReleaseToggleService[SERVICE_METHODS.IS_ENABLED].mockReturnValue(false);
+      mockReleaseToggleService[SERVICE_METHODS.HAS_ALL_ENABLED].mockReturnValue(false);
+
+      const { rerender } = await renderComponent();
+      expect(screen.queryByTestId(TEST_IDS.SINGLE_TOGGLE)).not.toBeInTheDocument();
+
+      // Act - Enable toggle and force component to re-evaluate
+      const updatedToggleName = `${TOGGLE_NAMES.TOGGLE1}-updated`;
+      mockReleaseToggleService[SERVICE_METHODS.IS_ENABLED].mockReturnValue(true);
+      rerender({ componentProperties: { singleToggle: updatedToggleName } });
```
dzhavat, line 158: suggestion: this part - and the same logic in the other tests - is testing whether the content is visible based on a new toggle name. What about keeping the same toggle name but only returning `true`/`false` for the `SERVICE_METHODS.IS_ENABLED` method? This is more like what is happening in reality, doesn't it?

dzhavat (commented):
praise: really nice initial implementation and tests. You've clearly thought a lot about the details 💯

There are a few comments, which can hopefully help improve things a bit.

libs/dh/core/shell/src/lib/dh-core-shell.component.ts
```diff
@@ -53,11 +54,13 @@ import { DhPrimaryNavigationComponent } from './dh-primary-navigation.component'
...
 export class DhCoreShellComponent {
   private readonly dhTopBarStore = inject(DhTopBarStore);
+  private releaseToggleService = inject(DhReleaseToggleService);
   breadcrumbService = inject(DhBreadcrumbService);
 
   titleTranslationKey = toSignal(this.dhTopBarStore.titleTranslationKey$);
 
   constructor(inactivityDetection: DhInactivityDetectionService) {
     inactivityDetection.trackInactivity();
+    this.releaseToggleService.refetch();
```
ManBearTM, line 64: question: does this constructor ever run multiple times? if not, is the refetch necessary?

libs/dh/shared/release-toggle/README.md
```diff
@@ -0,0 +1,273 @@
...
+constructor(private toggleService: DhReleaseToggleService) {}
+
+checkFeature() {
+  if (this.toggleService.isEnabled('premium-features')) {
+    // Enable premium functionality
+  }
+}
+```
+
+##### `hasAllEnabled(names: string[]): boolean`
```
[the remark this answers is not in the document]
ManBearTM: if renaming, then it should probably also be `areAnyEnabled`

libs/dh/shared/release-toggle/README.md
```diff
@@ -0,0 +1,273 @@
...
+  </div>
+  
+  <ng-template #legacyDashboard>
+    <app-legacy-dashboard></app-legacy-dashboard>
+  </ng-template>
+  
+  <!-- Error states -->
+  <div *ngIf="toggleService.hasError()" class="error-banner">
+    <p>Some features may be unavailable. <button (click)="retryToggles()">Retry</button></p>
+  </div>
```
ManBearTM, line 271: suggestion: promote the usage of new control flow syntax (`@if`)

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.service.ts
```diff
@@ -0,0 +1,245 @@
...
+  private handleRefetchError(error: unknown): void {
+    this.incrementFailureCount();
+    console.warn(`Manual refetch failed (${this.failureCount}/${this.maxRetries}):`, error);
+  }
+
+  /**
+   * Log when polling stops
+   */
+  private logPollingStop(): void {
+    console.error(`Release toggle polling stopped after ${this.maxRetries} consecutive failures`);
```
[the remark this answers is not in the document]
frontend-specialisten: Very good point, I will just add it now :)

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.directive.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.viewContainer.clear();
+    this.embeddedViewRef = null;
+  }
+
+  /**
+   * Cleanup when directive is destroyed.
+   */
+  ngOnDestroy(): void {
+    this.viewUpdateEffect.destroy();
+  }
```
ManBearTM, line 145: suggestion: remove this, i don't believe it is necessary to destroy the effect manually

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.directive.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.toggleExpression.set(expression);
+  }
+
+  /**
+   * Computed signal that evaluates the toggle expression.
+   */
+  private readonly shouldShowContent = computed((): boolean => {
+    const expression = this.toggleExpression();
+    return this.evaluateToggleExpression(expression);
+  });
```
ManBearTM, line 77: suggestion: since this signal is only used in the below effect, I think it would be simpler and easier to reason about if you just move this logic into the effect itself (and then remove this computed signal).

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.service.ts
```diff
@@ -0,0 +1,245 @@
...
+  private readonly maxRetries = DhReleaseToggleService.MAX_CONSECUTIVE_RETRIES;
+
+  // Internal state
+  private failureCount = 0;
+
+  // Apollo query setup
+  private readonly togglesQuery = query(GetReleaseTogglesDocument, {
+    fetchPolicy: 'network-only',
+    errorPolicy: 'all', // Continue even if there are errors
+  });
```
ManBearTM, line 47: suggestion: unless there are some requirements that I'm not aware of, I think this service is a bit "over-engineered". The `query` options has a `pollInterval` config that can be used to set up polling. I suggest to use that instead and then remove all this manual polling and retry logic. I'd rather have less code to maintain. We would probably lose the "max retries" functionality, but I don't think that is a big deal for something that polls every minute for an app with relatively few users 🙂

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.service.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  // Computed signal that maps toggle data to a record for easy lookup
+  private readonly togglesMap = computed(() => {
+    const data = this.togglesQuery.data();
+    if (!data?.releaseToggles) {
+      return {};
+    }
+
+    // Assume releaseToggles is an array with objects that have 'name' and 'enabled' properties
```
ManBearTM, line 56: nitpick: sounds like an AI comment, the releaseToggles is typed and is indeed an array with objects that have 'name' and 'enabled' properties. i would just remove it

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.service.ts
```diff
@@ -0,0 +1,245 @@
...
+    const togglesRecord: Record<string, boolean> = {};
+    data.releaseToggles.forEach((toggle: { name: string; enabled: boolean }) => {
```
ManBearTM, line 58:
```suggestion
    data.releaseToggles.forEach((toggle) => {
```

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.service.ts
```diff
@@ -0,0 +1,245 @@
...
+      togglesRecord[toggle.name] = toggle.enabled;
+    });
+
+    return togglesRecord;
+  });
```
ManBearTM, line 63: suggestion: you could also simplify this function a bit:

```ts
  // Computed signal that maps toggle data to a record for easy lookup
  private readonly togglesMap = computed<Record<string, boolean>>(() => {
    const toggles = this.togglesQuery.data()?.releaseToggles ?? [];
    return toggles.reduce((acc, toggle) => {
      return { ...acc, [toggle.name]: toggle.enabled };
    }, {});
  });

```

libs/dh/shared/release-toggle/README.md
```diff
@@ -0,0 +1,273 @@
...
+  }
+}
+```
+
+### Template Error Handling
+
+```html
+<div class="app-content">
+  <!-- Main content with fallbacks -->
+  <div *dhReleaseToggle="'new-dashboard'; else legacyDashboard">
```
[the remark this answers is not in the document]
frontend-specialisten: I actually thought it worked, but I've adjusted the directive now so it's supported :D

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.directive.ts
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * The toggle expression to evaluate.
+   * Single toggle name, array of toggle names, or inverse toggle (prefixed with '!')
+   */
+  dhReleaseToggle = input<ToggleExpression>('');
+
+  /**
+   * The template to show when the toggle condition is false.
+   */
+  @Input() dhReleaseToggleElse: TemplateRef<unknown> | null = null;
```
ManBearTM, line 70: use `input()` signal 😉

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.service.ts
```diff
@@ -0,0 +1,91 @@
...
+import { Injectable, computed, DestroyRef, inject } from '@angular/core';
+import { query } from '@energinet-datahub/dh/shared/util-apollo';
+
+import { GetReleaseTogglesDocument } from '@energinet-datahub/dh/shared/domain/graphql';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class DhReleaseToggleService {
+  private readonly destroyRef = inject(DestroyRef);
```
dzhavat, line 28: nitpick: looks unused

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.guard.ts
```diff
@@ -17,17 +17,34 @@
...
  */
 export function dhReleaseToggleGuard(toggleName: string): CanActivateFn {
   return () => {
     const releaseToggleService = inject(DhReleaseToggleService);
-    return releaseToggleService.isEnabled(toggleName);
+    const router = inject(Router);
+
+    const checkToggle = () => {
+      return releaseToggleService.isEnabled(toggleName) ? true : router.parseUrl('/');
+    };
```
dzhavat, line 38: suggestion:

```suggestion
    const checkToggle = () => releaseToggleService.isEnabled(toggleName) || router.parseUrl('/');
```

libs/dh/shared/release-toggle/src/lib/dh-release-toggle.service.ts
```diff
@@ -42,7 +40,9 @@ export class DhReleaseToggleService {
 
   // Public API - Read-only signals
   readonly toggles = computed(() => Array.from(this.togglesSet()));
-  readonly loading = this.togglesQuery.loading;
+  readonly loading = computed(() =>
+    !this.togglesQuery.called() ? true : this.togglesQuery.loading()
+  );
```
dzhavat, line 45:
```suggestion
  readonly loading = computed(() => !this.togglesQuery.called() || this.togglesQuery.loading());
```

ManBearTM (dismissed):

mimse (approved):

dzhavat (approved):

PR #4321, opened 2025-05-21
feat(dh): request measurements report

dzhavat:
## References

- #0000

### DIFF
--- a/apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/MeasurementsReports/MeasurementsReportsClient.cs
+++ b/apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/MeasurementsReports/MeasurementsReportsClient.cs
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Text;
 using Energinet.DataHub.WebApi.Clients.Wholesale.MeasurementsReports.Dto;
+using Newtonsoft.Json;
 
 namespace Energinet.DataHub.WebApi.Clients.Wholesale.MeasurementsReports;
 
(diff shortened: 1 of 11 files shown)


apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/MeasurementsReports/MeasurementsReportsClient.cs
```diff
@@ -44,4 +60,12 @@ public async Task<IEnumerable<RequestedMeasurementsReportDto>> GetAsync(Cancella
 
         return responseApiContent.OrderByDescending(x => x.CreatedDateTime);
     }
+
+    private static bool IsPeriodAcrossMonths(MeasurementsReportRequestFilterDto measurementsReportRequestFilter)
+    {
+        var startDate = TimeZoneInfo.ConvertTimeBySystemTimeZoneId(measurementsReportRequestFilter.PeriodStart, "Romance Standard Time");
```
mimse, line 66: Suggestion: I think nodetime can help here. So we dont need timezone

libs/dh/reports/feature-measurement-reports/src/components/request-report/request-report-modal.component.ts
```diff
@@ -89,16 +104,97 @@ export class DhRequestReportModal extends WattTypedModal<MeasurementsReportReque
...
+    const { period, gridAreas } = this.form.getRawValue();
+
+    if (period == null || gridAreas == null) {
+      return;
+    }
+
+    this.submitInProgress.set(true);
+
+    this.apollo
+      .mutate({
```
mimse, line 152: Suggestion: Use import { mutation } from '@energinet-datahub/dh/shared/util-apollo';

libs/dh/reports/feature-measurement-reports/src/components/request-report/request-report-modal.component.ts
```diff
@@ -89,16 +104,97 @@ export class DhRequestReportModal extends WattTypedModal<MeasurementsReportReque
...
+      return;
+    }
+
+    const { period, gridAreas } = this.form.getRawValue();
+
+    if (period == null || gridAreas == null) {
+      return;
+    }
+
+    this.submitInProgress.set(true);
```
mimse, line 149: Suggestion: Use state from the mutation help

mimse (changes requested):

apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/MeasurementsReports/MeasurementsReportsClient.cs
ManBearTM: I think it is hard to avoid the use of timezone here TBH 🤔 In the next gen of range date pickers, I think we should seriously consider dropping the time part and let the BFF convert it to the proper zone. Then we can have DateInterval and LocalDate to work with in the BFF, which is much nicer for these sort of things. But that is a discussion for another time 😄

libs/dh/reports/feature-measurement-reports/src/components/request-report/request-report-modal.component.ts
dzhavat: The use of `mutation` should be done now. I'll wait with the `injectToast` in a later PR. 🙂

libs/dh/reports/feature-measurement-reports/src/components/request-report/request-report-modal.component.ts
dzhavat: Done

apps/dh/api-dh/source/DataHub.WebApi/Clients/Wholesale/MeasurementsReports/MeasurementsReportsClient.cs
dzhavat: If we decide to change it, then we can look at both this one and Settlement reports because the same method is also used there. Can that be a separate PR?

mimse (approved):

PR #4327, opened 2025-05-23
1201 impl create signature in bff

Kramerrm:
GetMeteringPointAsync now request a signature first and after that do the request for masterdata to the new protected endpoint.


apps/dh/api-dh/source/DataHub.WebApi/AuthorizedHttpClientFactory.cs
```diff
@@ -39,6 +45,15 @@ public HttpClient CreateClient(Uri baseUrl)
         return client;
     }
 
+    public IElectricityMarketClient_V1 CreateClientWithSignature(string signature)
+    {
+        var client = _httpClientFactory.CreateClient();
+        client.DefaultRequestHeaders.Add("Signature", signature);
+        client.BaseAddress = new(_baseUrls.Value.ElectricityMarketBaseUrl);
+        var electricityMarketClient = new ElectricityMarketClient_V1(_baseUrls.Value.ElectricityMarketBaseUrl, client);
```
mimse, line 53: Nitpick: No need for a var, just rerturn new Elec.

apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/ElectricityMarketDebug.cs
```diff
@@ -35,7 +35,7 @@ public static async Task<IEnumerable<MeteringPointsGroupByPackageNumber>> GetMet
         CancellationToken ct,
         [Service] IElectricityMarketClient_V1 electricityMarketClient)
     {
-        var response = await electricityMarketClient.MeteringPointDebugAsync(gridAreaCode, ct).ConfigureAwait(false);
+        var response = await electricityMarketClient.MeteringPointDebugAsync(gridAreaCode, null, ct).ConfigureAwait(false);
```
mimse, line 38: Suggestion: Looks like this script default-api-version-param.ps1 did not run after the swagger was generated. It removed the need to provide the version param.

apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPointNode.cs
```diff
@@ -64,13 +77,38 @@ public static async Task<RelatedMeteringPointsDto> GetRelatedMeteringPointsAsync
...
+    {
+        if (httpContextAccessor.HttpContext == null)
+        {
+            throw new InvalidOperationException("Http context is not available.");
+        }
+
+        var user = httpContextAccessor.HttpContext.User;
+
+        var actorNumber = user.GetActorNumber();
+        var marketRole = Enum.Parse<EicFunctionAuth>(user.GetActorMarketRole());
```
mimse, line 99: Suggestion: Make a extension method like this one apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/Extensions/MeteringPointDtoExtensions.cs to map EicFunction from one subsystem to another. So we dont need up with an exception because they are out of sync

apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPointNode.cs
```diff
@@ -64,13 +77,38 @@ public static async Task<RelatedMeteringPointsDto> GetRelatedMeteringPointsAsync
...
+        var accessValidationRequest = new MeteringPointMasterDataAccessValidationRequest
+        {
+            MeteringPointId = meteringPointId,
+            ActorNumber = actorNumber,
+            MarketRole = marketRole,
+        };
+        var signature = await requestAuthorization.RequestSignatureAsync(accessValidationRequest);
+        var signatureJson = JsonSerializer.Serialize(signature);
+        var signatureByteArray = System.Text.Encoding.UTF8.GetBytes(signatureJson);
```
mimse, line 108: Suggestion: Could this be wrap up in a helper function. I´m guessing this is going to be done on every request?

mimse (changes requested):

apps/dh/api-dh/source/DataHub.WebApi/AuthorizedHttpClientFactory.cs
Kramerrm: Changed according comment.

apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/ElectricityMarketDebug.cs
Kramerrm: Did run the script now manual. I ques the reason was that I was not allowed to exectue powershell scripts on my laptop.
Also removed the not needed null at a couple of places.

apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPointNode.cs
Kramerrm: Hmmz. The conversion from a Signature object to a Base64 String is indeed needed for every request. Because it is set as Base64 in the header.
That raises the question why the RequestSignatureAsync operation is also not just returning it in Base64 format.
Something to discuss in team.

Kramerrm: Comment Jesper:
The reason we aren't returning a base64 string, is because quite soon the user of the create signature might get a return value where parameters needs to be modified, so we need the entire object at that point.

I agree So possible these many conversions are just needed.
Not sure if a helper function is a good option here, as it looks like this part is not always exactly the same. Sometimes specific parameter modifications are needed. Also the accessValidationRequest can differ case by case.

PR #4365, opened 2025-06-06
feat(watt): WattDropZone

ManBearTM:
dropzone
_GIF showing the component used with FormControl + custom validator_

This is a "first edition" of a dropzone component aka file selector/drag and drop. We need this component for reading a file directly in the browser, so I have not tried actually uploading files with it (we will cross that bridge when we need to). I have also not implemented any way to show the currently chosen files (other than in the story "FileList" as an example), because we have no need for that currently. The way I imagine we could implement the file list is as a separate component that binds to the same form control. That way, you could remove files from the list within the file list component (fx. by clicking on an "X"). The template could look something like this:

```html
<watt-dropzone [formControl]="files" />
<watt-file-list [formControl]="files" />
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,9 @@ build/terraform/**/*.tfvars
 !.vscode/launch.json
 !.vscode/extensions.json
 
+# Visual Studio 2015/2017 cache/options directory
+.vs/
+
 # misc
 /.angular/cache
 /.sass-cache
--- a/apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/Types/CommercialRelationDtoType.cs
+++ b/apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/Types/CommercialRelationDtoType.cs
@@ -29,7 +29,10 @@ public static partial class CommercialRelationDtoType
         commercialRelation.ActiveElectricalHeatingPeriods != null;
 
     public static DateTimeOffset? ElectricalHeatingStartDate([Parent] CommercialRelationDto commercialRelation) =>
-        commercialRelation.ElectricalHeatingPeriods.OrderBy(x => x.ValidFrom).FirstOrDefault()?.ValidFrom;
+        commercialRelation.ElectricalHeatingPeriods
+            .Where(x => x.ValidFrom < DateTimeOffset.UtcNow && x.ValidTo == DateTimeOffset.MaxValue)
+            .FirstOrDefault()
+            ?.ValidFrom;
 
     public static async Task<ActorNameDto?> GetEnergySupplierNameAsync(
         [Parent] CommercialRelationDto commercialRelation,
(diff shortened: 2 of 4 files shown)


libs/watt/package/dropzone/watt-dropzone.ts
```diff
@@ -0,0 +1,175 @@
...
+  signal,
+} from '@angular/core';
+import { ControlValueAccessor, NG_VALUE_ACCESSOR, ReactiveFormsModule } from '@angular/forms';
+import { WattButtonComponent } from '@energinet/watt/button';
+import { WattFieldComponent } from '@energinet/watt/field';
+import { VaterStackComponent, VaterUtilityDirective } from '@energinet/watt/vater';
+import { WattDropZoneIntlService } from './watt-dropzone-intl';
+
+// Slightly better typing than just raw string
+type MimeType = `${string}/${string}`;
```
mimse, line 35: Suggestion: Maybe this could be used https://www.npmjs.com/package/@types/mime-types ?

libs/watt/package/dropzone/watt-dropzone.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  handleDragOver(event: DragEvent) {
+    if (!event.dataTransfer) return;
+
+    const items = Array.from(event.dataTransfer.items);
+    const accept = this.accept();
+
+    // Ignore non-file items such as strings
+    if (items.some((i) => i.kind !== 'file')) return;
```
mimse, line 155: Question: It seems wrong, that just because one item is not a file, all are rejected? But its properly right :-)

frontend-specialisten (commented):
Nicely done! I have a few suggestions regarding validation and user feedback.

- File type validation feedback - Currently, if a user attempts to upload an unsupported file type, there's no clear indication of what went wrong. I think the component should provide some sort of feedback.
- File size constraints - Consider adding a maxFileSize input to prevent users from uploading overly large files. This would be beneficial for both UX and backend performance.
- Form validation integration - The component implements ControlValueAccessor but doesn't integrate with Angular's validation system. Adding the Validator interface would enable proper form validation and error state management.

Suggested Enhancements:
- Add validation errors for build-in validations to provide some feedback for the user
- Consider adding the `Validator` to integrate it nicely to reactive forms 
- Consider adding file size limits 
- Consider adding a maxFiles limit for multiple file uploads

libs/watt/package/dropzone/watt-dropzone.ts
ManBearTM: If you look at the `index.d.ts` file of that package (https://www.npmjs.com/package/@types/mime-types?activeTab=code), it just looks to be types for the `mime-types` package. We could of course use the `mime-types` package to map extensions to MIME types during runtime (so that we can use extensions for accept input), but that is super not worth it, since the package pulls in `mime-db`, which is a 200KB+ JSON file.

Perhaps you were thinking something more like this (https://github.com/samhuk/mime-types-typescript/blob/master/dist/latestUpdate.ts)? That is not a package though and I'm not sure it is worth it either, since that list has to be maintained as new MIME types are added.

mimse: I just wanted it for type completion when coding, should not be a part of the runtime

ManBearTM: I'm all for type completion, but I think in this case the benefit from the type completion does not outweight the cost of maintenance. I could only find this repo (https://github.com/samhuk/mime-types-typescript), and that requires us to manually rebuild the list occasionally (I don't know how often that is though).

frontend-specialisten (approved):

PR #4398, opened 2025-06-16
refac(watt): vater vatest

ManBearTM:
## Motivation
I have been seeing more and more workarounds for `vater` components, especially `vater-flex`. One of the annoying problems was that `vater-flex` was not "flexible" enough, but would instead often use its intrinsic size over the size dictated from the containing layout (e.g. another `vater-flex` component). This was especially problematic with deeply nested components (fx. a table in a router outlet), where you would have to resort to workarounds like absolute positioning or non-percentage based sizes.

To make this easier, I have changed `vater-flex` to always be fully flexible by ignoring its intrinsic size. This means that it will now always shrink if the containing layout dictates it (no more `scrollable` needed). Sibling flex items will also now have their size equally distributed, unless one of the items has `fill` set (or if using `autoSize`, more on that later). In general, what I hope this change means is that `vater-flex` should feel more like it _just works™_. The downside to this change is that if `vater-flex` does not have an explicit size (from a containing layout fx), then it will not take into account the intrinsic size and will instead just shrink as much as possible (overflow other than "visible" = minimum size of zero, so beware of that). Usually you don't want to use `vater-flex` without an explicit size, but it is still possible to revert to the old sizing by setting `autoSize` on `vater-flex` (essentially `flex-basis: auto;`).

I also saw some (justifiably) use of `vater-flex` which should have been `vater-stack` instead. Sometimes it was to `wrap` items (which `vater-stack` can now also do!), other times I think the `grow`, `shrink` and `basis` inputs were too powerful, meaning you could simply make a stack layout from `vater-flex`. With the addition of the `autoSize` and `wrap` inputs, there was no longer any need for `grow`, `shrink` and `basis`, so I removed them. I hope these changes makes the components more distinct from each other and makes it clearer on which component to use.

## Breaking change for `watt`
If using `watt-data-table` or `vater` components in general, then this PR may break your layout. You can use `autoSize` on `watt-data-table` and `vater-flex` to get the old layout back. That should generally be enough.

## Summary

- vater: changed `vater-flex` to use `flex-basis: 0` by default
- vater: added `autoSize` input to `vater-flex` to get the old behavior back (`flex-basis: auto`)
- vater: removed `grow`, `shrink` and `basis` inputs from `vater-flex`
- vater: added `wrap` input to `vater-stack`, removing the need for using `vater-flex` for wrapping flex items
- vater: added `align` input to `vater-flex` (of course it broke something in dh, so fixed that as well)
- vater: `flex-start` and `flex-end` for `align` and `justify` is now just `start` and `end`
- vater: made `scrollable` part of VaterUtilityDirective, making it available on all vater components
- vater: moved all inline styles on vater components to classes on the `vater` css layer
- vater: added VaterLayoutDirective, that is only used internally for better code reuse
- watt: tweaked `watt-data-table` and added an `autoSize` input for intrinsic sizing (used mostly in EO)
- dh: changed many `vater-flex` with `wrap` to `vater-stack` with newly added `wrap` input
- dh: added `autoSize` to a select few `vater-flex` instances where the old behavior was required (for now)
- dh: made many table filters horizontally scrollable
- dh: minor fixes to layout in dev-examples
- eo: streamlined usages of `vater-stack` a few places
- eo: added `autoSize` to all `watt-data-tables`, as they should all be intrinsically sized


Sejruppen (dismissed):
praise: Nice work! I have checked all components on our end, and everything looks good 🙂

libs/watt/package/vater/vater-utility.directive.ts
```diff
@@ -16,34 +16,33 @@
...
+  scrollable = input(false, { transform: booleanAttribute });
 
-  @HostBinding('class.vater-center')
-  get _center() {
-    return this.center === '';
-  }
+  // Computed class names
+  protected fillClass = computed(() => this.fill() && `vater-fill-${this.fill()}`);
+  protected insetClass = computed(() => this.inset() && `vater-inset-${this.inset()}`);
+  protected class = computed(() => [this.fillClass(), this.insetClass()].filter(Boolean));
```
dzhavat, line 47: question: the class property in the layout directive ends with `.join(' ')`. Should we add it here as well or remove it there?

dzhavat: Let's remove it. One thing less to keep in mind.

ManBearTM: 8ba0eafc9ad1d4901c02a0f9dd74b2723fed4c51

PR #4411, opened 2025-06-19
Ett 4650 report confirmation

Sejruppen:
## Description

This PR alters the design and UX of the report generation button. I also added some unit tests and some e2e tests of the new report feature :)

https://github.com/user-attachments/assets/bdaf972c-3e15-44f3-a2e6-60fcb0972b2b


libs/eo/reports/data-access-api/eo-reports.service.ts
```diff
@@ -41,7 +41,7 @@ export class EoReportsService implements OnDestroy {
   readonly loading = computed(() => this.#loading());
   readonly error = computed(() => this.#error());
 
-  private readonly POLLING_INTERVAL = 5000; // 5 seconds
+  private readonly POLLING_INTERVAL = 5000; // 10 seconds
```
ckr123, line 44: isnt it 10000 for 10 sec?

Sejruppen: We decided to alter it to 5 seconds, I have updated the comment :)

PR #4445, opened 2025-06-26
Fixed date, and renamed stuff

mimse:


libs/dh/metering-point/feature-overview/src/graphql/get-metering-point-by-id.graphql
```diff
@@ -33,6 +33,7 @@ query GetMeteringPointById(
     meteringPointId
     isEnergySupplier(energySupplierActorGln: $actorGln)
     isGridAccessProvider(gridAccessProviderActorGln: $actorGln)
+    electricalHeatingStartDate
```
dzhavat, line 36: suggestion: let's add it to the mock data to get rid of a console error

PR #4450, opened 2025-06-29
1306 update bff to use new create signature for relevant measurement requests

janhped:
Issues:

**1)** The name `EicFunction` was already registered by another type
  - This error occurs when GraphQLTestService is executed. 

This GraphQL name collision must handled some how or maybe ActorNumber and EicFunction shouldn't be parameters to the GetByDayQuery record but instead come from IHttpContextAccessor

For info, a Work-around for "The name `EicFunction` was already registered..." to work was to add descriptor.Name("ActorEicFunctionType") that way I could adjust the unittest "Get_measurements_has_quantity_or_quality_changed" for using Measurements.Client 8.10.0
 
protected override void Configure(IEnumTypeDescriptor<EicFunction> descriptor)
{
    **descriptor.Name("ActorEicFunctionType");**
    descriptor.AsIsCase();
}

**2)** Frontend - ActorNumber and EicFunction needs to be added to the GetByDayQuery.


apps/dh/api-dh/source/DataHub.WebApi.Tests/TestServices/GraphQLTestService.cs
```diff
@@ -13,26 +13,36 @@
...
 using System.Threading.Tasks;
+using Energinet.DataHub.Core.FunctionApp.TestCommon.Databricks;
 using Energinet.DataHub.Edi.B2CWebApp.Clients.v1;
 using Energinet.DataHub.Edi.B2CWebApp.Clients.v3;
+using Energinet.DataHub.MarketParticipant.Authorization.Model;
+using Energinet.DataHub.MarketParticipant.Authorization.Services;
+using Energinet.DataHub.Measurements.Abstractions.Api.Queries;
 using Energinet.DataHub.Measurements.Client;
+using Energinet.DataHub.Measurements.Client.Authorization;
+using Energinet.DataHub.Measurements.Client.ResponseParsers;
```
jonasdmoeller, line 27: Renamed to `Energinet.DataHub.Measurements.Client.Mappers`

apps/dh/api-dh/source/DataHub.WebApi.Tests/TestServices/GraphQLTestService.cs
```diff
@@ -121,6 +144,12 @@ public GraphQLTestService()
 
     public Mock<IHttpContextAccessor> HttpContextAccessorMock { get; set; }
 
+    public Mock<IMeasurementsApiHttpClientFactory> MeasurementsApiHttpClientFactoryMock { get; set; }
+
+    public Mock<IRequestSignatureFactory> RequestSignatureFactoryMock { get; set; }
+
+    public Mock<IMeasurementsDtoResponseParser> MeasurementsDtoResponseParserMock { get; set; }
```
jonasdmoeller, line 151: Renamed to `MeasurementsResponseMapper`

apps/dh/api-dh/source/DataHub.WebApi.Tests/TestServices/GraphQLTestService.cs
mimse: Fixed

dzhavat (approved):
praise: nice changes.

suggestion: as a future task we should consider expanding the `dh-actor-storage.ts` test suite. There are some methods that are not covered and this service is quite central.

PR #4485, opened 2025-07-04
refac(dh): simplify CSV parser

ManBearTM:
This PR includes the following changes:

- Moves all code related to upload measurements into its own feature (`feature-upload-measurements`)
- Rewrites the CSV parser to reduce dependencies, simplify flow and improve performance
  - Remove the need for handling encoding manually, taking advantage of the FileReader API instead
  - Centralizes the "parse result" state and aggregates the result row-by-row
    - This simplifies the logic since we only have to look at one function to understand how the data is collected
    - When positions are missing or unexpected, the error can now report the actual row index 
    - Also avoids the browser freezing up at the end, since it no longer has to run through all measurements twice
- Added snapshot tests for the CSV parser

### DIFF
--- a/libs/watt/package/package.json
+++ b/libs/watt/package/package.json
@@ -1,7 +1,6 @@
 {
   "type": "module",
   "name": "@energinet/watt",
-  "version": "2.1.15",
   "version": "3.0.1",
   "license": "Apache-2.0",
   "exports": {


libs/dh/metering-point/feature-upload-measurements/src/upload-page.ts
```diff
@@ -211,10 +198,9 @@ import { assertIsDefined } from '@energinet-datahub/dh/shared/util-assert';
     </watt-card>
   `,
 })
-export class DhMeasurementsUploadComponent {
+export class DhUploadMeasurementsComponent {
   private navigate = injectRelativeNavigate();
-  protected measurements = inject(DhMeasurementsUploadDataService);
-  private readonly csvParser: CsvParseService = inject(CsvParseService);
+  protected measurements = inject(DhUploadMeasurementsService);
```
mimse, line 203: Nitpick: Why not private, protected is for inheritance?

libs/dh/metering-point/feature-upload-measurements/src/upload-service.ts
```diff
@@ -0,0 +1,97 @@
...
+} from '@energinet-datahub/dh/shared/domain/graphql';
+import { MeasureDataResult } from './models/measure-data-result';
+import { injectToast } from '@energinet-datahub/dh/shared/ui-util';
+import { mutation } from '@energinet-datahub/dh/shared/util-apollo';
+import { assert, assertIsDefined } from '@energinet-datahub/dh/shared/util-assert';
+import { parseMeasurements } from './parse-measurements';
+
+@Injectable({ providedIn: 'root' })
+export class DhUploadMeasurementsService {
+  private toast = injectToast('meteringPoint.measurements.upload.toast');
```
mimse, line 34: Nitpick: If i understand i correctly when according to https://angular.dev/style-guide#use-readonly-on-properties-that-are-initialized-by-angular it should be private readonly? :-)

libs/dh/metering-point/feature-upload-measurements/src/upload-service.ts
```diff
@@ -0,0 +1,97 @@
...
+        throw new Error(`Unsupported metering point type: ${type}`);
+    }
+  };
+
+  /** Parses a CSV file of measurement data, streaming the result. */
+  parseFile = (file: File, resolution: string) =>
+    parseMeasurements(file, this.mapResolution(resolution));
+
+  /** Sends measurements to the server. */
+  send = (
```
mimse, line 75: Question: No need for await, and error handling here?

mimse (approved):
Praise: very nice refactor, just a few comments.

libs/dh/metering-point/feature-upload-measurements/src/upload-page.ts
ManBearTM: Good eye, I think it was protected because it used to be accessed from the template. But that is no longer the case, so private it is 🙂 

https://github.com/Energinet-DataHub/greenforce-frontend/pull/4485/commits/bab065b7003f8a9b32dc7d1aaeb3544c03790f08

libs/dh/metering-point/feature-upload-measurements/src/upload-page.ts
```diff
@@ -235,11 +221,23 @@ export class DhMeasurementsUploadComponent {
...
+    return qualities.size > 1 ? 'MIXED' : first;
+  });
+
   private validate = async (): Promise<ValidationErrors | null> => {
     if (!this.file.value) return null;
     const [file] = this.file.value;
-    await this.csvParser.parseFile(file).forEach(this.csv.set);
+    const resolution = this.resolution();
+    assertIsDefined(resolution);
+    await this.measurements.parseFile(file, resolution).forEach(this.csv.set);
```
dzhavat, line 240: praise: I was about to write a comment how `forEach` won't work here, but then TIL it works! because `forEach` here is from RxJS, and not the native JS method. Nicely done 💯

libs/dh/metering-point/feature-upload-measurements/src/upload-page.ts
```diff
@@ -235,11 +221,23 @@ export class DhMeasurementsUploadComponent {
     }
   });
 
-  csv = signal<CsvParseResult | null>(null);
+  csv = signal<MeasureDataResult | null>(null, { equal: () => false });
+  totalSum = computed(() => this.csv()?.sum ?? 0);
+  totalPositions = computed(() => this.csv()?.measurements.length ?? 0);
+  progress = computed(() => (this.csv()?.isFatal ? undefined : this.csv()?.progress));
```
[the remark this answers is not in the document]
ManBearTM: That was actually intentional, due to `undefined` basically meaning "do not show the progress bar". But I admit that it was pretty confusing so by adding an extra input to dropzone, I was able to make the behavior more explicit:

https://github.com/Energinet-DataHub/greenforce-frontend/pull/4485/commits/e0d928a796a6b9fd2b3961111928862a6ad2d1a3

PR #4498, opened 2025-07-10
Feat: Add revision log to measurements report and settlement report

johevemi:
## References

- #0000


apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SettlementReportRevisionLogTests.cancelSettlementReport.verified.json
```diff
@@ -0,0 +1,9 @@
+﻿{
+  "activity": "cancelSettlementReport",
+  "payload": {
+    "query": "mutation { cancelSettlementReport(input: { requestId: { id: \u002212345678-1234-1234-1234-123456789012\u0022 } }) { boolean } }",
+    "variables": {}
+  },
+  "affectedEntityType": "SettlementReport",
+  "affectedEntityKey": null
```
ManBearTM, line 8: question: same as above, should the requestId be included here?

apps/dh/api-dh/source/DataHub.WebApi.Tests/TestServices/GraphQLTestService.cs
```diff
@@ -60,6 +61,7 @@ public GraphQLTestService()
         MeasurementsResponseMapperMock = new Mock<IMeasurementsResponseMapper>();
         RequestSignatureFactoryMock = new Mock<IRequestSignatureFactory>();
         HttpClientFactoryMock = new Mock<IHttpClientFactory>();
+        SettlementReportsClientMock = new Mock<ISettlementReportsClient>();
```
ManBearTM, line 64: suggestion: delete the other client named "SettlementReportClientMock", since we now mock the wrapper client instead.

ManBearTM (changes requested):
Good start, but I have a few comments 🙂

ManBearTM (approved):
GJ ME!

PR #4536, opened 2025-08-05
INC0462912 - Use watt-date-range-chip instead of two individual watt-chip

frontend-specialisten:
## Description

Changes to `message-archive` to solve inconsistent behaviour when starting and filtering dates: 
- Use `watt-date-range-chip` instead of two individual `watt-chip`
- Use a single `watt-datepicker` with `[range]="true"` in `start.component.ts` instead of having two individual `watt-datepicker`

## References

- #0000

### DIFF
--- a/libs/dh/globalization/assets-localization/src/assets/i18n/da.json
+++ b/libs/dh/globalization/assets-localization/src/assets/i18n/da.json
@@ -596,8 +596,7 @@
       "businessReason": "Forretningsårsag",
       "sender": "Afsender",
       "receiver": "Modtager",
-      "start": "Starttidspunkt",
-      "end": "Sluttidspunkt",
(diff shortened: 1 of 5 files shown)


libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -600,8 +600,7 @@
       "businessReason": "Forretningsårsag",
       "sender": "Afsender",
       "receiver": "Modtager",
-      "start": "Starttidspunkt",
-      "end": "Sluttidspunkt",
+      "dateRange": "Datointerval",
```
ManBearTM, line 603: question: should this be "reverted"?

ManBearTM (approved):

PR #4549, opened 2025-08-08
feat(dh): move upload measurements to actions menu

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-measurements/src/components/upload.component.ts
```diff
@@ -81,6 +82,7 @@ import { assertIsDefined } from '@energinet-datahub/dh/shared/util-assert';
     WATT_CARD,
     DhEmDashFallbackPipe,
   ],
+  providers: [DhMeasurementsUploadDataService],
```
ManBearTM, line 85: I think this was "provided" in the navigation component, as it has to persist after submitting the measurement data. This is because when you click "Godkend", the page navigates back and thus destroys the upload component while the "send measurements" is still in flight. Perhaps we can just make the upload service "providedIn: root" instead?

ManBearTM (dismissed):
🏅

PR #4560, opened 2025-08-12
feat(dh): allow large text files in measurements reports

dzhavat:
## References

- #0000


libs/dh/reports/feature-measurements-reports/src/components/request-report/request-report-modal.component.ts
```diff
@@ -217,6 +227,7 @@ export class DhRequestReportModal extends WattTypedModal<MeasurementsReportReque
             end: period.end ? period.end : null,
           },
           gridAreaCodes: gridAreas,
+          preventLargeTextFiles: !allowLargeTextFiles,
```
mimse, line 230: Question: Why is one called prevent and the other one allow. Seems odd?

dzhavat: Because it's made exactly the same way as in settlement reports to keep the same pattern. A rename might follow for both reports to make the naming/param more clear.

PR #4569, opened 2025-08-13
Refactor/remove actor part 8

mimse:


libs/watt/package/modal/watt-modal.service.ts
```diff
@@ -112,9 +112,3 @@ export class WattModalService {
     });
   }
 }
-
-@NgModule({
-  imports: [MatDialogModule],
-  providers: [WattModalService],
-})
-export class WattModalModule {}
```
dzhavat, line 120: issue: CI is failing with some errors related to this. Also the change might cause issues in many places where a modal is used, so we'll need to check the different ways a modal is shown (instantiated via a service, via a route, used directly in a template, etc.) to make sure it all still works

libs/dh/market-participant/feature-market-participant/src/components/create/create.component.ts
```diff
@@ -298,15 +292,5 @@ export class DhActorsCreateActorModalComponent extends WattTypedModal {
...
-
-    this.isCompleting.set(false);
-  }
-
-  private newOrganizationNameToActorName(): void {
-    this.newOrganizationForm.controls.companyName.valueChanges
-      .pipe(takeUntilDestroyed())
-      .subscribe((value) => {
-        this.newActorForm.controls.name.setValue(value);
-      });
```
dzhavat, line 310: question: this method made sure that the organization's name was also set as the market participant's name. Does that still happen or is it not necessary anymore?

libs/watt/package/modal/watt-modal.service.ts
mimse: They are fixed now. I think we are okay. I can try to deploy to dev02, and click more around?

dzhavat (approved):

PR #4581, opened 2025-08-15
feat(dh): show metering point ids in table

dzhavat:
## References

- #0000

### DIFF
--- a/libs/dh/reports/feature-measurements-reports/src/components/overview/metering-point-cell.component.ts
+++ b/libs/dh/reports/feature-measurements-reports/src/components/overview/metering-point-cell.component.ts
@@ -1,0 +1,50 @@
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
+import { TranslocoDirective, TranslocoPipe } from '@jsverse/transloco';
+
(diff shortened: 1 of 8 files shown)


libs/dh/reports/feature-measurements-reports/src/components/overview/metering-point-cell.component.ts
```diff
@@ -0,0 +1,68 @@
...
+          {{ meteringPointIds.join(', ') }}
+        } @else {
+          {{
+            t('itemsAndCount', {
+              items: meteringPointIds[0] + ', ' + meteringPointIds[1],
+              remainingCount: meteringPointIds.length - 2,
+            })
+          }}
+        }
+      } @else if (meteringPointTypes.length < 4) {
```
mimse, line 44: Question: Your if checks for meteringPointIds but the if else checks on meteringPointTypes, is this correct?

mimse (approved):

libs/dh/reports/feature-measurements-reports/src/components/overview/metering-point-cell.component.ts
dzhavat: Yes. If the ids are there, it should display them, otherwise check for types

PR #4597, opened 2025-08-20
fix(dh): handle redirect for newly registered users

dzhavat:
### DataHub
- newly registered users were redirected back to the app to `<app-url>?code=<value>` URL by B2C. This caused issues with login where users were ended up again on the login page every time they tried to login. The fix here checks for `?code=` part and sets the `dhRedirectTo` param to `/` (home path). 

## References

- #0000


apps/dh/app-dh/src/app/datahub-app.component.ts
```diff
@@ -103,6 +103,11 @@ export class DataHubAppComponent implements OnInit {
       return '/';
     }
 
+    // Handle case where a new users is redirected to the app after sign up
+    if (path.startsWith('?code=')) {
```
frontend-specialisten, line 107: Consideration: Are we sure the `path` will always `startsWith` or should we use something like "includes" instead?

frontend-specialisten (dismissed):

PR #4623, opened 2025-08-26
fix(watt): auto-expand menu groups

dzhavat:
### Watt Design System
- fix auto-expand of the menu group that contains an active router link when the active item is added after initialization based on some logic e.g. `[at]if` or structural directive.

## References

- #0000

### DIFF
--- a/libs/watt/package/package.json
+++ b/libs/watt/package/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "@energinet/watt",
-  "version": "3.0.11",
+  "version": "3.0.12",
   "license": "Apache-2.0",
   "exports": {
     ".": {


libs/watt/package/shell/nav-list/watt-expand-on-active-link.directive.ts
```diff
@@ -16,32 +16,37 @@
...
 @Directive({
   selector: '[wattExpandOnActiveLink]',
   exportAs: 'wattExpandOnActiveLink',
 })
-export class WattExpandOnActiveLinkDirective implements AfterViewInit {
+export class WattExpandOnActiveLinkDirective {
   private panel = inject(MatExpansionPanel);
 
-  wattNavListItemComponents = input<readonly WattNavListItemComponent[]>([]);
+  private subs: OutputRefSubscription[] = [];
```
ManBearTM, line 31: suggestion: make this a local variable in `navListItemsEffect`, since it doesn't look like it is being used anywhere else.

dzhavat: Fixed https://github.com/Energinet-DataHub/greenforce-frontend/pull/4623/commits/0576560a9cc6277b0f7cd1600857c1c54da37db5

ManBearTM (dismissed):

ManBearTM (approved):

PR #965, opened 2022-09-27
feat: add charges webapi

ShaBloze:
## Description

Update Charges nuget package.
Add new Charges WebApi and call it from the new page.

For now we get all charges.

## References

- #1656
https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/energinet-datahub/geh-charges/1656

PR #4633, opened 2025-09-01
feat(api-dh): Add DocumentFormatter for JSON/XML formatting

frontend-specialisten:
## Description

This pull request introduces a new utility for formatting document content (JSON and XML) before returning it from the API, and adds comprehensive unit tests to ensure correct formatting behavior. The main change is that documents returned by the `MessageArchiveController` are now automatically formatted for readability if they are valid JSON or XML and not already formatted, with safeguards for very large documents.

**Document formatting feature:**

* Added a new static utility class `DocumentFormatter` in `DocumentFormatter.cs` that detects and formats JSON and XML documents, preserving already formatted content and handling edge cases like empty, null, or oversized documents.
* Updated `MessageArchiveController` so that documents returned by `GetDocumentByIdAsync` are passed through `DocumentFormatter.FormatDocumentIfNeeded` before being returned, improving readability for clients. Also added validation for the document ID format. [[1]](diffhunk://#diff-d126084c81791b5684fc50dc4823523abd81a9e29e6bfcd00a4e37c1b308a2f3R17) [[2]](diffhunk://#diff-d126084c81791b5684fc50dc4823523abd81a9e29e6bfcd00a4e37c1b308a2f3R38-R50)

**Testing and validation:**

* Added a comprehensive set of unit tests in `DocumentFormatterTests.cs` to cover various scenarios, including valid/invalid JSON/XML, empty/null/whitespace input, already formatted content, XML declarations, documents with leading whitespace, very large documents, and edge cases.

## References

- [#965](https://github.com/Energinet-DataHub/team-mosaic/issues/964)
- [#965](https://github.com/Energinet-DataHub/team-mosaic/issues/965)

### DIFF
--- a/apps/dh/api-dh/source/DataHub.WebApi/Utilities/DocumentFormatter.cs
+++ b/apps/dh/api-dh/source/DataHub.WebApi/Utilities/DocumentFormatter.cs
@@ -1,0 +1,120 @@
+// Copyright 2020 Energinet DataHub A/S
+//
+// Licensed under the Apache License, Version 2.0 (the "License2");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Text.Json;
+using System.Xml.Linq;
+
+namespace Energinet.DataHub.WebApi.Utilities;
+
+public static class DocumentFormatter
+{
+    private const int MaxDocumentSizeForFormatting = 100 * 1024 * 1024; // 100 MB - safely handles 50MB documents
+
+    private static readonly JsonSerializerOptions _jsonOptions = new()
+    {
+        WriteIndented = true,
+        Encoder = System.Text.Encodings.Web.JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
+        DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.Never,
+        PropertyNamingPolicy = null,
+    };
+
+    public static string FormatDocumentIfNeeded(string document)
+    {
+        if (string.IsNullOrWhiteSpace(document))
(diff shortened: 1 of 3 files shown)


mimse:
Normally the json and xml files I have seen, is formatted, and then highlighted with watt-code component in the client. What is the use case here?

apps/dh/api-dh/source/DataHub.WebApi/Utilities/DocumentFormatter.cs
```diff
@@ -0,0 +1,158 @@
...
+// limitations under the License.
+
+using System.Text.Json;
+using System.Xml.Linq;
+
+namespace Energinet.DataHub.WebApi.Utilities;
+
+public static class DocumentFormatter
+{
+    private const int MaxDocumentSizeForFormatting = 100 * 1024 * 1024; // 100 MB - safely handles 50MB documents
```
ManBearTM, line 22: question: is this important to enforce?

apps/dh/api-dh/source/DataHub.WebApi/Utilities/DocumentFormatter.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        if (TryFormatXml(document, out var formattedXml))
+        {
+            return formattedXml;
+        }
+
+        return document;
+    }
+
+    private static bool IsAlreadyFormatted(string input)
```
ManBearTM, line 57: question: is this necessary? i assume it is to save time/processing, but in the near future, it will only be true for old messages. perhaps it is better to trade some CPU cycles for reduced complexity 🙂

ManBearTM:
> Normally the json and xml files I have seen, is formatted, and then highlighted with watt-code component in the client. What is the use case here?

They want to remove the formatting in the backend to save space in the database.

apps/dh/api-dh/source/DataHub.WebApi/Utilities/DocumentFormatter.cs
frontend-specialisten: Should be removed now

apps/dh/api-dh/source/DataHub.WebApi/Utilities/DocumentFormatter.cs
frontend-specialisten: Should be removed now

ManBearTM (approved):

PR #4637, opened 2025-09-02
chore: add permissions to pipelines

dzhavat:
## References

- #0000


.github/workflows/ci-orchestrator.yml
```diff
@@ -11,9 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 name: CI orchestrator
 
+permissions:
```
AdmirMuric, line 16: This is a very broad set of permissions that you are granting. Are you sure that you need write on all of these different ones? 

Try starting out with

```
permissions:
  contents: read
```

And se, if any errors arise. Also, if any of the underlying jobs need say write, then use it down in the specific job, so that you limit what has write access vs read access.

dzhavat: I've used these permissions because the `dh_ci_dotnet` job calls https://github.com/Energinet-DataHub/.github/blob/v14/.github/workflows/dotnet-postbuild-test.yml#L142-L147 internally, which requires these permissions. The `contents` is set to `read` in that workflow, but the pre-release job in another workflow (https://github.com/Energinet-DataHub/.github/blob/v14/.github/workflows/dotnet-build-prerelease.yml#L163) requires `contents: write`. I can try moving them to the relevant jobs instead of keeping them here.

AdmirMuric (approved):

PR #4679, opened 2025-09-15
Call current period api when no historical values are to be shown

mimse:


apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeasurementOperations.cs
```diff
@@ -90,11 +91,21 @@ public static async Task<IEnumerable<MeasurementAggregationByYearDto>> GetAggreg
     [Authorize(Roles = new[] { "metering-point:search" })]
     public static async Task<MeasurementDto> GetMeasurementsAsync(
         bool showOnlyChangedValues,
+        bool showHistoricalValues,
         GetByDayQuery query,
         CancellationToken ct,
         [Service] IMeasurementsClient client)
     {
-        var maybeMeasurements = await client.GetByDayAsync(query, ct);
+        var maybeMeasurements = showHistoricalValues ?
```
HenrikSommer, line 99: Not sure but I think this should take `showOnlyChangedValues` into account as well

```suggestion
        var maybeMeasurements = showOnlyChangedValues || showHistoricalValues ?
```

PR #4680, opened 2025-09-15
fix: support different measurement units for send measurements

MadsDue:
## References

- #0000


libs/dh/metering-point/feature-upload-measurements/src/upload-page.ts
```diff
@@ -230,9 +232,27 @@ export class DhUploadMeasurementsPage {
...
+    switch (measureUnit) {
+      case MeteringPointMeasureUnit.KvArh:
+        return SendMeasurementsMeasurementUnit.KiloVoltAmpereReactiveHour;
+      case MeteringPointMeasureUnit.KWh:
+        return SendMeasurementsMeasurementUnit.KilowattHour;
+      default:
+        // Default to kWh for any other unit types
+        return SendMeasurementsMeasurementUnit.KilowattHour;
+    }
+  }
```
ManBearTM, line 254: suggestion: move this into the `upload-service` instead along with other "mapping" functions.

ManBearTM: fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/4680/commits/5a9e8e5b18b7662dd4b4f7d587e6c88811756cae

ManBearTM (approved):

PR #4686, opened 2025-09-16
feat(dh): show unit in month and year view for measurements

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-measurements/src/components/month.component.ts
```diff
@@ -145,14 +145,21 @@ export class DhMeasurementsMonthComponent {
...
+  private sum = computed(
+    () =>
+      `${this.formatNumber(
+        this.measurements()
+          .map((x) => x.quantity)
+          .filter((quantity) => quantity !== null && quantity !== undefined)
+          .reduce((acc, quantity) => acc + Number(quantity), 0)
+      )} ${this.unit()}`
   );
+  private unit = computed(() => {
```
mimse, line 157: Suggestion: We could add Unit to apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/Types/MeasurementPositionDtoType.cs but its up to you.

mimse (approved):

libs/dh/metering-point/feature-measurements/src/components/month.component.ts
dzhavat: Yes, but that will only have an effect on the day view. The month and year views use different types 🤔

PR #4691, opened 2025-09-17
fix(dh): move-in customer details

dzhavat:
## References

- #0000


mimse (approved):
Two things:
1) I would remove dh- from the file names
2) I´m thinking this might be a big thing, maybe it would have its own feature under metering-point?

PR #4713, opened 2025-09-22
fix(dh): adjustments to move-in

dzhavat:
### DataHub
- enable/disable CPR 2 field if Name 2 contains a value

## References

- #0000

### DIFF
--- a/libs/dh/metering-point/feature-move-in/src/components/move-in.component.ts
+++ b/libs/dh/metering-point/feature-move-in/src/components/move-in.component.ts
@@ -19,7 +19,6 @@
 import { Component, effect, inject } from '@angular/core';
 import { TranslocoDirective } from '@jsverse/transloco';
 import { NonNullableFormBuilder, ReactiveFormsModule, Validators } from '@angular/forms';
-import { Subscription } from 'rxjs';
 import { toSignal } from '@angular/core/rxjs-interop';
 
 import { WATT_MODAL, WattTypedModal } from '@energinet-datahub/watt/modal';
@@ -84,6 +83,13 @@ export class DhMoveInComponent extends WattTypedModal {
     isProtectedAddress: this.fb.control<boolean>(false),
   });
 
+  private privateCustomerForm = this.fb.group({
+    name1: this.fb.control<string>('', Validators.required),
+    cpr1: this.fb.control<string>('', Validators.required),
+    name2: this.fb.control<string>(''),
+    cpr2: this.fb.control<string>({ value: '', disabled: true }, Validators.required),
+  });
+
   contactDetailsForm = this.fb.group({
     // Define form controls and validation here
   });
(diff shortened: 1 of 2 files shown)


libs/dh/metering-point/feature-move-in/src/components/move-in.component.ts
```diff
@@ -95,25 +91,45 @@ export class DhMoveInComponent extends WattTypedModal {
...
-          cpr2: this.fb.control<string | undefined>({ value: undefined, disabled: true }),
+          name2: this.fb.control<string>(''),
+          cpr2: this.fb.control<string>({ value: '', disabled: true }, Validators.required),
         })
       );
 
+      const name2Control = this.customerDetailsForm.controls.privateCustomer?.controls.name2;
+      const cpr2Control = this.customerDetailsForm.controls.privateCustomer?.controls.cpr2;
+
+      this.name2Subscription = name2Control?.valueChanges.subscribe({
```
mimse, line 122: Suggestion: Can we use toSignal and an effect to do the same?

dzhavat: Not really. That is because the `name2` control is unknown at init time. So its type is `undefined`. This doesn't play well with the `toSignal` function. 🤔

image

ManBearTM: What if you created the controls at init time, but separate from the form. Then just add the controls conditionally. Something like:

```ts
  customerDetailsForm = this.fb.group<MoveInCustomerDetailsFormType>({
    transactionId: this.fb.control<string>({ value: '', disabled: true }, Validators.required),
    cutOffDate: this.fb.control({ value: new Date(), disabled: true }, Validators.required),
    moveInType: this.fb.control<string>('', Validators.required),
    customerType: this.fb.control('private'),
    customerType: this.fb.control(this.customerTypeInitialValue),
    isProtectedAddress: this.fb.control<boolean>(false),
  });
  
  privateCustomerForm = this.fb.group({
    name1: this.fb.control<string>('', Validators.required),
    cpr1: this.fb.control<string>('', Validators.required),
    name2: this.fb.control<string | undefined>(undefined),
    cpr2: this.fb.control<string | undefined>({ value: undefined, disabled: true }),
  });

  // then later
  if (customerType === 'private') {
      this.customerDetailsForm.addControl('privateCustomer', this.privateCustomerForm);
  }
```

Then you could possibly create a signal for the "name2" FormControl.

libs/dh/metering-point/feature-move-in/src/components/move-in.component.ts
```diff
@@ -73,18 +74,13 @@ import { MoveInCustomerDetailsFormType, MoveInType } from '../types';
...
     transactionId: this.fb.control<string>({ value: '', disabled: true }, Validators.required),
     cutOffDate: this.fb.control({ value: new Date(), disabled: true }, Validators.required),
     moveInType: this.fb.control<string>('', Validators.required),
-    customerType: this.fb.control('private'),
-    privateCustomer: this.fb.group({
-      name1: this.fb.control<string>('', Validators.required),
-      cpr1: this.fb.control<string>('', Validators.required),
-      name2: this.fb.control<string | undefined>(undefined),
-      cpr2: this.fb.control<string | undefined>({ value: undefined, disabled: true }),
-    }),
```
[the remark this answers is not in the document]
dzhavat: Great idea 💡
Looks like it works. And the manual subscription can be removed. 💯

mimse (approved):
💯

PR #4714, opened 2025-09-22
Auto open search modal at `message-archive`

frontend-specialisten:
## Description

Auto open search modal at `message-archive`

## References

- [#1067](https://github.com/Energinet-DataHub/team-mosaic/issues/1067)

### DIFF
--- a/libs/dh/message-archive/feature-search/src/lib/start.component.ts
+++ b/libs/dh/message-archive/feature-search/src/lib/start.component.ts
@@ -49,7 +49,7 @@ import { DhDateTimeRangeField } from './datetime-range-field';
       *transloco="let t; read: 'messageArchive.start'"
       size="small"
       [title]="t('title')"
-      [autoOpen]="true"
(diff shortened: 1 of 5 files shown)


libs/dh/message-archive/feature-search/src/lib/start.component.ts
```diff
@@ -119,6 +120,7 @@ export class DhMessageArchiveSearchStartComponent {
   form = inject(DhMessageArchiveSearchFormService);
   searchChanged = output<GetArchivedMessagesQueryVariables>();
   modal = viewChild.required(WattModalComponent);
+  autoOpen = input(false);
```
ManBearTM, line 123: question: does it have to be an input? you can just do this if it should always "auto open":

```html
<watt-modal autoOpen />
```

frontend-specialisten: Should be simplified now :)

PR #4724, opened 2025-09-24
chore: change default landing-page to /metering-point/search

frontend-specialisten:
## Description
- Change default landing-page to /metering-point/search

## References

- [#1078](https://github.com/Energinet-DataHub/team-mosaic/issues/1078)

### DIFF
--- a/libs/dh/core/shell/src/lib/dh-core-shell.routes.ts
+++ b/libs/dh/core/shell/src/lib/dh-core-shell.routes.ts
@@ -22,7 +22,12 @@ import { Routes } from '@angular/router';
 import { DhCoreShellComponent } from './dh-core-shell.component';
 import { DhCoreLoginComponent } from './dh-core-login.component';
 
-import { BasePaths, ReportsSubPaths, getPath } from '@energinet-datahub/dh/core/routing';
+import {
+  BasePaths,
+  ReportsSubPaths,
+  MeteringPointSubPaths,
+  getPath,
+} from '@energinet-datahub/dh/core/routing';
 import { PermissionGuard } from '@energinet-datahub/dh/shared/feature-authorization';
 
 export const dhCoreShellRoutes: Routes = [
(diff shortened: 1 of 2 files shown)


libs/dh/core/shell/src/lib/dh-core-shell.routes.ts
```diff
@@ -32,7 +37,7 @@ export const dhCoreShellRoutes: Routes = [
     children: [
       {
         path: '',
-        redirectTo: getPath<BasePaths>('message-archive'),
+        redirectTo: `${getPath<BasePaths>('metering-point')}/${getPath<MeteringPointSubPaths>('search')}`,
```
mimse, line 40: Suggestion: Use combine path function from dhPaths

libs/dh/metering-point/feature-search/src/routes.ts
```diff
@@ -49,7 +49,9 @@ const marketRolesWithDataAccess = [
 export const dhMeteringPointRoutes: Routes = [
   {
     path: '',
-    canActivate: [PermissionGuard(['metering-point:search'])],
+    canActivate: [
+      PermissionGuard(['metering-point:search'], `/${getPath<BasePaths>('message-archive')}`),
```
mimse, line 53: Suggestion: Maybe we should add / in permission.guard.ts file, so the next developer doesnt forget to do that?

libs/dh/core/shell/src/lib/dh-core-shell.routes.ts
dzhavat, line 40: suggestion: what about creating an async redirect here which can decide where to send the user? That way there will be no need to extend the `PermissionGuard` with the extra param.

mimse: Okay :-)

dzhavat (approved):

PR #4743, opened 2025-09-29
Feature/add permission

mimse:


libs/dh/shared/domain/src/lib/permission.ts
```diff
@@ -46,6 +46,7 @@ export const permissions = [
   'users:manage',
   'users:reactivate',
   'users:view',
+  'charges:view',
```
dzhavat, line 49: nitpick: the permissions are ordered alphabetically to make it easier to see duplicates/missing items. Can we move this new one to its place :)

mimse: Updated

PR #4775, opened 2025-10-06
feat(dh): allow additional market roles to request measurements reports

dzhavat:
### DataHub
- add support for "SystemOperator" and "DanishEnergyAgency" when accessing and requesting measurements reports
- adjust UI to hide unnecessary fields when requesting as these roles
- hide `grid areas` column in overview when active market participant is one of these roles
- allow admin to request on behalf of these roles

## References

- #0000

### DIFF
--- a/libs/dh/reports/feature-measurements-reports/src/components/overview/overview.component.ts
+++ b/libs/dh/reports/feature-measurements-reports/src/components/overview/overview.component.ts
@@ -75,18 +75,15 @@ export class DhOverview {
   dataSource = new WattTableDataSource<DhMeasurementsReport>([]);
 
   displayedColumns = computed(() => {
-    const allColumns = Object.keys(this.columns);
     const isFas = this.isFas();
 
-    let columnsToDisplay = isFas
-      ? allColumns
-      : allColumns.filter((column) => column !== 'actorName');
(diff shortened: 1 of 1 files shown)


libs/dh/reports/feature-measurements-reports/src/components/overview/overview.component.ts
```diff
@@ -70,10 +75,18 @@ export class DhOverview {
...
-    return isFas ? tableColumns : tableColumns.filter((column) => column !== 'actorName');
+    let columnsToDisplay = isFas
+      ? allColumns
+      : allColumns.filter((column) => column !== 'actorName');
+
+    columnsToDisplay = specialMarketRoles.includes(this.actorStorage.getSelectedActor().marketRole)
+      ? columnsToDisplay.filter((column) => column !== 'gridAreas')
+      : columnsToDisplay;
+
+    return columnsToDisplay;
```
ManBearTM: suggestion: you can avoid the `let` by chaining the `filter` calls:
```ts
return Object.keys(this.columns)
  .filter(column => isFas || column !== 'actorName')
  .filter(column => !isSpecialMarketRole || column !== 'gridAreas')
```

PR #4783, opened 2025-10-08
feat(watt, menu): add async menu items support, and fix imports

frontend-specialisten:
## Description

- Add support for async menu items in `watt-menu`
- Optimize imports of `watt-menu`
- Replace `mat-menu` -> `watt-menu` in `dh-profile-avatar.component.ts`
image

## References

- #0000

### DIFF
--- a/libs/watt/package/menu/+storybook/menu-async-story.component.ts
+++ b/libs/watt/package/menu/+storybook/menu-async-story.component.ts
@@ -89,15 +89,15 @@ export class MenuAsyncStoryComponent {
     // Watch for loadDelay changes and restart loading
     effect((onCleanup) => {
       const delay = this.loadDelay();
-      
+
       // Reset loaded state
       this.itemsLoaded.set(false);
-      
+
       // Start new loading timeout
       const timeout = setTimeout(() => {
         this.itemsLoaded.set(true);
       }, delay);
-      
+
       // Cleanup function will be called when effect reruns or component is destroyed
       onCleanup(() => {
         clearTimeout(timeout);
--- a/libs/watt/package/menu/+storybook/menu-async-story.component.ts
+++ b/libs/watt/package/menu/+storybook/menu-async-story.component.ts
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 //#endregion
-import { Component, Input, signal, OnInit, OnDestroy } from '@angular/core';
+import { Component, input, signal, effect, OnDestroy } from '@angular/core';
 import { WattButtonComponent } from '../../button';
 import { WattIconComponent } from '../../icon';
 import { WattMenuComponent } from '../watt-menu.component';
@@ -48,7 +48,7 @@ import { WattMenuTriggerDirective } from '../watt-menu-trigger.directive';
         <watt-menu #asyncMenu>
           @if (itemsLoaded()) {
             <watt-menu-item (click)="onAction('Profile')">
-              @if (showIcons) {
+              @if (showIcons()) {
                 <watt-icon name="user" />
               }
               Profile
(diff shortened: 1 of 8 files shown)


libs/watt/package/menu/watt-menu.component.ts
```diff
@@ -119,7 +116,20 @@ export class WattMenuComponent {
...
+   * CSS classes to apply to the menu panel.
+   * @ignore
+   */
+  panelClasses = computed(() => {
+    const classes = ['watt-menu-panel'];
+    if (this.hasIcons()) {
+      classes.push('watt-menu-panel--has-icons');
+    }
+    return classes;
+  });
```
ManBearTM, line 129: suggestion: i think you can save some code by doing this in the template instead:

```html
<mat-menu
  class="watt-menu-panel"
  [class.watt-menu-panel--has-icons]="hasIcons()"
>
```

libs/watt/package/menu/watt-menu.component.ts
```diff
@@ -64,11 +64,8 @@ import { WattMenuItemComponent } from './watt-menu-item.component';
 @Component({
   selector: 'watt-menu',
   template: `
-    <mat-menu
-      #menu="matMenu"
-      [class]="'watt-menu-panel' + (hasIcons() ? ' watt-menu-panel--has-icons' : '')"
-    >
-      <ng-content select="watt-menu-group, watt-menu-item" />
+    <mat-menu #menu="matMenu" [class]="panelClasses().join(' ')">
+      <ng-content />
```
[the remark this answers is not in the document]
frontend-specialisten: This is needed to allow async patterns

libs/watt/package/menu/watt-menu.component.ts
```diff
@@ -119,7 +116,20 @@ export class WattMenuComponent {
    */
   hasIcons = computed(() => this.menuItems().some((item) => item.hasIcon()));
```
ManBearTM, line 117: question: it looks like all this "has icons" logic only applies to styles. have you considered a CSS-only solution using the `:has` selector?

libs/watt/package/menu/index.ts
```diff
@@ -16,7 +16,22 @@
  * limitations under the License.
  */
 //#endregion
+// Individual exports (kept for backward compatibility)
```
ManBearTM, line 19:
```suggestion
// Individual exports
```

Not really true IMO, you can still require the individual exports (fx. for child queries)

libs/watt/package/menu/+storybook/menu-async-story.component.ts
```diff
@@ -0,0 +1,125 @@
...
+})
+export class MenuAsyncStoryComponent implements OnInit, OnDestroy {
+  @Input() showIcons = true;
+  @Input() set loadDelay(value: number) {
+    this._loadDelay = value;
+    this.resetLoading();
+  }
+  get loadDelay(): number {
+    return this._loadDelay;
+  }
```
ManBearTM, line 89: nitpick: i know it is just a story, but let's not use `@Input` anymore 🙂

libs/watt/package/menu/+storybook/menu-async-story.component.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  itemsLoaded = signal(false);
+  lastAction = signal<string | null>(null);
+
+  constructor() {
+    // Watch for loadDelay changes and restart loading
+    effect(() => {
+      this.loadDelay();
+      this.startLoading();
+    });
```
ManBearTM, line 95: just a little nitpick, you can use the `onCleanup` argument in `effect` to clear the timeout, then you don't need the `ngOnDestroy` and the `if (this.loadingTimeout)` check

ManBearTM (approved):

PR #4790, opened 2025-10-09
UI/UX adjustments

Sejruppen:
**Base branch for all changes made during the "Get-Shit-Done" day. See individual commits for related PRs**

**Updates to the main navigation**
image

**Updates to related metering points**

image

### DIFF
--- a/libs/watt/package/package.json
+++ b/libs/watt/package/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "@energinet/watt",
-  "version": "4.0.21",
+  "version": "4.0.22",
   "license": "Apache-2.0",
   "exports": {
     ".": {


libs/dh/metering-point/feature-overview/src/components/dh-metering-point.component.ts
```diff
@@ -134,15 +134,18 @@ import { DhReleaseToggleDirective } from '@energinet-datahub/dh/shared/release-t
               <span
                 gap="s"
                 direction="row"
```
ManBearTM: nitpick: Unrelated to this PR, so we can fix it later. But these have no effect right? Since it doesn't have a `vater` directive applied.

libs/watt/package/field/watt-field.component.ts
```diff
@@ -128,6 +129,7 @@ export class WattFieldComponent {
   errors = signal<ValidationErrors | null>(null);
   isRequired = signal(false);
   isEmpty = computed(() => this.errors()?.['required'] || this.errors()?.['rangeRequired']);
+  showErrors = input(true);
```
ManBearTM: if we can fix the validator to only run on `submit` as suggested above, then we don't need this input anymore

libs/watt/package/modal/watt-modal.component.scss
```diff
@@ -18,6 +18,11 @@
   --watt-modal-min-height: auto;
   --watt-modal-height: calc(100vh - var(--watt-topbar-height, 0px));
   --watt-modal-content-padding: 0 var(--watt-space-ml);
+  border-radius: var(--watt-radius-l);
+}
+
+:host {
+  border-radius: 20px;
```
ManBearTM: suggestion: we just standardized the `border-radius`, seems a little soon to deviate from it. if "small, medium, large" is too limiting, then we could consider using numbers instead (--watt-radius-100, --watt-radius-200, --watt-radius-300, --watt-radius-400 etc). The numbers are not related to their actual size, but are only for "hierarchy".

libs/watt/package/shell/nav-list/watt-nav-list.component.scss
```diff
@@ -21,19 +21,25 @@ watt-nav-list {
...
+      padding: var(--watt-space-s);
       display: block;
       border-left: var(--watt-space-xs) solid transparent;
+      border-radius: var(--watt-space-s);
       color: var(--watt-color-neutral-white);
       text-decoration: none;
     }
 
+    a:hover {
+      transition: background-color linear 0.05s;
```
ManBearTM: suggestion: should we use the new animation variables?

libs/watt/package/text-field/watt-text-field.component.ts
```diff
@@ -158,6 +166,14 @@ export class WattTextFieldComponent implements ControlValueAccessor, AfterViewIn
...
       this.formControl().setValue(trimmedValue);
     });
+
+    // Programmatically focus the input if autoFocus is true
+    if (this.autoFocus()) {
+      // Use requestAnimationFrame to ensure the element is ready in the DOM
+      requestAnimationFrame(() => {
+        this.inputField().nativeElement.focus();
+      });
+    }
```
ManBearTM: suggestion: this is a nice feature, but it is a shame that it is only available on the watt-text-field - consider moving this logic to watt-field, so that all fields can benefit from it.

libs/watt/package/shell/nav-list/watt-nav-list.component.scss
dzhavat: Fixed in https://github.com/Energinet-DataHub/greenforce-frontend/pull/4790/commits/f91b15a6ba7c6db7b965315a7cba717f9e412d22

libs/dh/metering-point/feature-overview/src/components/dh-metering-point.component.ts
dzhavat: Yeah, you're right. Can see the pattern with other elements in this file. Will fix it in another PR.

libs/dh/metering-point/feature-search/src/components/dh-search.component.ts
```diff
@@ -108,13 +110,14 @@ import { dhMeteringPointIdValidator } from './dh-metering-point.validator';
 export class DhSearchComponent {
   private readonly router = inject(Router);
   private readonly doesMeteringPointExist = lazyQuery(DoesMeteringPointExistDocument);
+  protected submitted = signal(false);
 
   searchControl = new FormControl('', {
     validators: [Validators.required, dhMeteringPointIdValidator()],
     nonNullable: true,
   });
```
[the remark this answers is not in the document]
Sejruppen: That's a great idea - I'll make another PR on that 🙂

PR #4822, opened 2025-10-20
Added groups to dropdown

mimse:
Dropdown now supports groups

image

### DIFF
--- a/libs/watt/package/dropdown/index.ts
+++ b/libs/watt/package/dropdown/index.ts
@@ -17,5 +17,10 @@
  */
 //#endregion
 export { WattDropdownComponent } from './watt-dropdown.component';
-export { type WattDropdownOption, type WattDropdownOptions } from './watt-dropdown-option';
+export {
+  type WattDropdownOption,
+  type WattDropdownOptions,
+  type WattDropdownGroupedOptions,
+  type WattDropdownOptionGroup,
+} from './watt-dropdown-option';
 export { type WattDropdownValue } from './watt-dropdown-value';


libs/watt/package/dropdown/watt-dropdown.component.ts
```diff
@@ -208,54 +179,75 @@ export class WattDropdownComponent implements ControlValueAccessor, OnInit {
...
 
-        if (this.sortDirection()) {
-          optionsCopy = this.sortOptions(optionsCopy);
-        }
+        const hasGroups = options.some(
+          (option) =>
+            'options' in option && Array.isArray((option as WattDropdownOptionGroup).options)
+        );
+
+        this.hasGroups.set(hasGroups);
```
ManBearTM: suggestion: `hasGroups` could be a `computed` signal instead

libs/watt/package/dropdown/watt-dropdown.component.ts
```diff
@@ -167,15 +139,14 @@ export class WattDropdownComponent implements ControlValueAccessor, OnInit {
...
   disableSelectedMode = input(false);
-
   sortDirection = input<'asc' | 'desc'>();
 
   /**
    * Sets the options for the dropdown.
+   * Can be a flat array of options or an array containing both options and option groups.
    */
-  options = model<WattDropdownOptions>([]);
+  options = model<WattDropdownGroupedOptions>([]);
```
ManBearTM: question: should it not be `model<WattDropdownOptions | WattDropdownGroupedOptions>`?

libs/watt/package/dropdown/watt-dropdown.component.ts
mimse: Good idea, fixed

mimse:
> Looks good! Does add a bit of complexity to `watt-dropdown`, but what can you do 🤷 Just a few comments

Two components, but that is no good either i think

ManBearTM (approved):

PR #4838, opened 2025-10-23
feat(dh, metering-point): only allow actors with specific market roles to see measurements upload button

frontend-specialisten:
## Description

- Only allow specific market roles for seeing upload measurements button to avoid errors like this:
image

- Add new mocked actor who is a `GridAccessProvider` 

## References

- #0000


libs/dh/metering-point/feature-overview/src/components/dh-metering-point-actions.component.ts
```diff
@@ -127,10 +134,25 @@ export class DhMeteringPointActionsComponent {
   connectionState = input<ConnectionState | null>();
   installationAddress = input<InstallationAddress | null>();
 
+  protected readonly hasAllowedMarketRolesForSendMeasurementsRequest = toSignal(
+    combineLatest([
+      this.permissionService.hasMarketRole(EicFunction.MeteredDataResponsible),
```
dzhavat: nitpick: `measurements:manage` permission cannot be assigned to roles under `MeteredDataResponsible` and `Delegated` so the permission itself will not allow these market roles to see the button. So these are not necessary.

PR #4849, opened 2025-10-24
feat(metering-point, steps): open raw messages in new tab

frontend-specialisten:
## Description

Open messages in proccess overview steps as raw content in a new tab

## References

- [#1102](https://github.com/Energinet-DataHub/team-mosaic/issues/1102)

### DIFF
--- a/libs/watt/package/package.json
+++ b/libs/watt/package/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "@energinet/watt",
-  "version": "4.0.28",
+  "version": "4.1.0",
   "license": "Apache-2.0",
   "exports": {
     ".": {
--- a/libs/watt/package/portal/index.ts
+++ b/libs/watt/package/portal/index.ts
@@ -1,0 +1,21 @@
+//#region License
+/**
+ * @license
+ * Copyright 2020 Energinet DataHub A/S
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License2");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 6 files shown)


apps/dh/api-dh/source/DataHub.WebApi/Modules/MessageArchive/MeteringPointProcessStepNode.cs
```diff
@@ -28,10 +29,33 @@ public static partial class MeteringPointProcessStepNode
...
+        return result.Edges != null && result.Edges.Count > 0
+            ? result.Edges[0].Node
+            : null;
+    }
+
     static partial void Configure(IObjectTypeDescriptor<MeteringPointProcessStep> descriptor)
     {
         descriptor.Name("MeteringPointProcessStep");
-        descriptor.BindFieldsExplicitly();
+        descriptor.BindFieldsImplicitly();
```
ManBearTM: question: why? 😄

libs/dh/metering-point/feature-process-overview/src/components/steps.ts
```diff
@@ -94,4 +100,46 @@ export class DhMeteringPointProcessOverviewSteps {
...
+    try {
+      const content = await firstValueFrom(this.http.get(documentUrl, { responseType: 'text' }));
+
+      if (!content) {
+        this.toastService.open({
+          type: 'danger',
+          message: this.transloco.translate('messageArchive.document.loadFailed'),
+        });
+        return;
+      }
```
ManBearTM: suggestion: replace this with:

```ts
assertIsDefined(content); // import from @energinet-datahub/dh/shared/util-assert
```

This will update the type if it is defined or throw otherwise. Then you don't need the duplicate toastService call.

PR #4853, opened 2025-10-24
added breadcrumb

mimse:
image

### DIFF
--- a/libs/watt/package/package.json
+++ b/libs/watt/package/package.json
@@ -1,7 +1,7 @@
 {
   "type": "module",
   "name": "@energinet/watt",
-  "version": "4.1.0",
+  "version": "4.1.1",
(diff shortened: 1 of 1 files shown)


libs/dh/metering-point/feature-overview/src/components/dh-metering-point.component.ts
```diff
@@ -233,8 +233,6 @@ export class DhMeteringPointComponent {
 
   constructor() {
     effect(() => {
-      this.breadcrumbService.navigationEnded();
-
```
dzhavat: suggestion: looks like this was intentionally added https://github.com/Energinet-DataHub/greenforce-frontend/pull/4167/files Can you test if everything still works after removing it? Like navigating between parent -> child -> parent metering points.

PR #4869, opened 2025-10-28
Added information cards

mimse:
image

### DIFF
--- a/libs/dh/charges/feature-prices/src/components/information.component.ts
+++ b/libs/dh/charges/feature-prices/src/components/information.component.ts
@@ -20,7 +20,6 @@ import { Component } from '@angular/core';
 import { TranslocoDirective, TranslocoPipe } from '@jsverse/transloco';
 
 import {
-  Maybe,
   VatClassification,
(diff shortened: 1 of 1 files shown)


libs/dh/charges/feature-prices/src/components/information.component.ts
```diff
@@ -17,9 +17,125 @@
...
+      chargeDescription:
+        'En længere beskrivelse, der også kan fylde to linjer, eller måske endda endnu mere end to linjer.',
+      vatClassification: VatClassification.Vat25,
+      transparentInvoicing: false,
+      validFromDateTime: dayjs().subtract(2, 'year').toDate(),
+      validToDateTime: dayjs().subtract(1, 'year').toDate(),
+    },
+  ];
+
+  isCurrent(from?: Date, to?: Maybe<Date>) {
```
dzhavat: nitpick: the `Maybe` type is `Maybe<T> = T | null` so the typing here is a bit too wide if `validToDateTime` can only be `Data | undefined.`

PR #4871, opened 2025-10-29
fix(watt): space between buttons in drawer actions component

dzhavat:
### Watt

**After**
image

## References

- #0000


ManBearTM (approved):
🥇

PR #4873, opened 2025-10-29
feat: get contact cpr - actual call from bff to elmark

michaelmajgaard:


apps/dh/api-dh/source/DataHub.WebApi/Clients/ElectricityMarket/V1/nswag.json
```diff
@@ -3,7 +3,7 @@
   "defaultVariables": null,
   "documentGenerator": {
     "fromDocument": {
-      "url": "https://localhost:6005/swagger/v1/swagger.json",
+      "url": "https://localhost:7085/swagger/v1/swagger.json",
```
mimse: Suggestion: Maybe we should point it at dev02?

michaelmajgaard: Fixed

PR #4886, opened 2025-11-03
fix(styles): exclude watt-drawer-heading from min-width constraint in vater-row

frontend-specialisten:
## Description

- Exclude watt-drawer-heading from min-width constraint in vater-row

Before `dh-message-archive-search-details` had issues with long content in the `watt-description-list` would push the `watt-drawer-actions` out of view:

image

After:
image

## References

- #0000


Sejruppen (approved):
praise: Nice find 🙂

PR #4889, opened 2025-11-03
feat(dh): charge series part two

ManBearTM:


libs/dh/shared/util-apollo/src/lib/query.ts
```diff
@@ -113,6 +113,7 @@ export type QueryResult<TResult, TVariables extends OperationVariables> = {
   getDocument: () => TypedDocumentNode<TResult, TVariables>;
   getOptions: () => QueryOptions<TVariables>;
   setOptions: (options: Partial<QueryOptions<TVariables>>) => Promise<ApolloQueryResult<TResult>>;
+  variables: Signal<Partial<TVariables>>;
```
mimse: Question: What is this for? :-)

PR #4947, opened 2025-11-14
Split charge links into tariff-subscriptions and fees

mimse:


libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -507,8 +507,8 @@
       "processes": {
         "tabLabel": "Procesoverblik"
       },
-      "prices": {
-        "tabLabel": "Priser"
+      "charge-links": {
+        "tabLabel": "Pris tilknytninger"
```
ManBearTM: question: shouldn't it be "Priser"? If not, then it should be in one word: "Pristilknytninger"

libs/dh/metering-point/feature-chargelink/src/components/tariff-subscriptions.ts
```diff
@@ -0,0 +1,132 @@
...
+    WATT_TABLE,
+    WattDatePipe,
+    WattDropdownComponent,
+    WattDataTableComponent,
+    WattDataFiltersComponent,
+  ],
+  providers: [DhNavigationService],
+  template: `<watt-data-table
+      [enableCount]="false"
+      [enableSearch]="false"
```
ManBearTM:
```suggestion
      [header]="false"
```

libs/dh/metering-point/feature-chargelink/src/components/tariff-subscriptions.ts
```diff
@@ -0,0 +1,132 @@
...
+  }));
+  navigation = inject(DhNavigationService);
+  dataSource = computed(
+    () =>
+      new WattTableDataSource(
+        (this.query.data()?.chargeLinksByMeteringPointId ?? []).filter(
+          (chargeLink) => chargeLink.type != ChargeType.Fee
+        )
+      )
+  );
```
ManBearTM: suggestion: use reactive `dataSource` util here as well

PR #4957, opened 2025-11-19
fix(dh): replace metering point id with internal id in url

dzhavat:
### DataHub
- replace "metering point id" with "internal id" in url
- adjust related metering points card to use "internal id" in links
- adjust breadcrumbs to use "internal id" in links
- adjust "debug view" page to NOT include "metering point id" in url

### Solution overview
- "metering point id" in search field
- check whether metering point exists
  - if NOT -> show error / + debug view button
- use "internal id" to navigate to metering point page
- activate guard checks again whether metering point with "internal id" exists (useful in case of deep link)
  - if NOT -> redirect back to Search page
- resolver uses the "internal id" from URL param and maps it to a "metering point id" (used as `input`s in components)

Note: both `meteringPointId` and `internalMeteringPointId` are then available as `inputs` in the components

## References

- #0000

### DIFF
--- a/apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
+++ b/apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
@@ -855,11 +855,6 @@ type MeteringGridAreaImbalanceSearchResult {
   outgoingImbalancePerDay: [MeteringGridAreaImbalancePerDayDto!]!
 }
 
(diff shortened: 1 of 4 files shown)


ManBearTM (approved):
praise: nice solution!

Looks like metering point search got a little slower since it now has to look up the metering point twice. I think that treating the internal ID as the primary ID would be faster in terms of cache lookup and fewer requests. The data model could then be something like:

```gql
meteringPoint {
  id #internalMeteringPointId
  sensitive_id # meteringPointId
}
```

That would also make it clearer in code that you should be careful with the `sensitive_id`. 

But all this for another time I guess 🙂

PR #4958, opened 2025-11-19
Feature/UI to attact charge to metering point

mimse:


libs/dh/metering-point/feature-chargelink/src/components/create.ts
```diff
@@ -0,0 +1,194 @@
...
+    WattTextFieldComponent,
+    WattDatepickerComponent,
+
+    VaterStackComponent,
+  ],
+  encapsulation: ViewEncapsulation.None,
+  styles: `
+    dh-metering-point-create-charge-link {
+      watt-button .mdc-button.mat-mdc-button {
+        width: 100%;
```
ManBearTM: suggestion: Let's not rely on an implementation detail of `watt-button`. Move this override to `watt-button` instead. I'd argue it is a good default for all button variants, but maybe just limit it to "selection" type buttons for now.

libs/dh/metering-point/feature-chargelink/src/components/create.ts
```diff
@@ -0,0 +1,194 @@
...
+      size="small"
+      *transloco="let t; prefix: 'meteringPoint.createChargeLink'"
+      [title]="t('title')"
+    >
+      @let type = selectedType();
+      @if (type === null) {
+        <vater-stack align="stretch" gap="ml">
+          @for (chargeType of ChargeTypes; track chargeType) {
+            <watt-button
+              alignText="start"
```
ManBearTM: question: this was removed right?

libs/dh/metering-point/feature-chargelink/src/components/create.ts
```diff
@@ -0,0 +1,194 @@
...
+    startDate: this.fb.control<Date | null>(null, Validators.required),
+  });
+
+  selectedType = signal<ChargeType | null>(null);
+
+  chargeOptions = computed<WattDropdownOptions>(() =>
+    (this.chargesQuery.data()?.chargesByTypeAndOwner ?? []).map((charge) => ({
+      value: charge.id,
+      displayValue: charge.displayName,
+    }))
```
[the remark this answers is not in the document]
mimse: Good idea

PR #4983, opened 2025-11-21
Refactor/charge links

mimse:


libs/dh/metering-point/feature-chargelink/src/components/create.ts
```diff
@@ -133,7 +120,7 @@ export class DhMeteringPointCreateChargeLink extends WattTypedModal {
     startDate: this.fb.control<Date | null>(null, Validators.required),
   });
 
-  selectedType = signal<ChargeType | null>(null);
+  selectedType = model<ChargeType | null>(null);
```
ManBearTM: nit: this does not need to be a model as far as I understand. it can still be a regular signal 🙂

mimse: I´ll play around with it and see :-)

PR #4996, opened 2025-11-26
Signals in Modal!

Sejruppen:


libs/watt/package/modal/+storybook/watt-modal-from-class.component.ts
```diff
@@ -37,18 +38,20 @@ import { FormControl, ReactiveFormsModule } from '@angular/forms';
...
-      this.isLoading = false;
-    }, 300);
+  constructor() {
+    super();
+    effect(() => {
+      this.isLoading.set(true);
+      setTimeout(() => {
+        this.isLoading.set(false);
+      }, 300);
+    });
```
dzhavat: nitpick: there's no need to put the signal update in an `effect`. The previous solution was just as fine :)

PR #4999, opened 2025-11-26
refac(dh): add tariff-tax charge type

ManBearTM:


apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql
```diff
@@ -1716,6 +1716,12 @@ input ChargeSortInput {
   code: SortDirection
 }
 
+input ChargeTypeSortInput {
+  name: SortEnumType
+  type: SortEnumType
+  isFee: SortEnumType
+}
```
ManBearTM: this doesn't look right, but I'm thinking we fix this in a later PR once we implement in-memory pagination

mimse (dismissed):
Looks good

PR #5001, opened 2025-11-26
fix(dh): adjustments to create metering point

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-search/src/components/dh-create-metering-point.component.ts
```diff
@@ -104,15 +119,21 @@ export class DhCreateMeteringPoint {
       validityDate: dhMakeFormControl<Date | null>(this.today, Validators.required),
       status: dhMakeFormControl({ value: 'NEWLY_CREATED', disabled: true }),
       meteringPointId: dhMakeFormControl('', Validators.required),
```
Sejruppen: nitpick: Could you set up some basic validation regarding the length and/or type of input?

dzhavat: Good idea. I have it locally and didn't want to make the PR bigger. So it come as a follow-up PR.

Sejruppen (approved):

PR #5009, opened 2025-11-27
fix(dh): adjustments to create metering point

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-search/src/components/dh-create-metering-point.component.html
```diff
@@ -44,20 +44,30 @@ <h4>{{ t("details.title") }}</h4>
 
         <vater-stack align="start" class="watt-space-stack-m">
           <label>{{ t("details.status") }}</label>
-          <watt-radio group="status" [formControl]="detailsControls.status" value="NEWLY_CREATED">
-            {{ t("details.newlyCreated") }}
-          </watt-radio>
+          <small>{{ t("details.newlyCreated") }}</small>
```
mimse: Question: Small that was a new one. How come? :-)

PR #1170, opened 2022-11-28
feat: 1836 Create charge price page adjustments

ShaBloze:
## References

- #1836
https://app.zenhub.com/workspaces/team-volt-60a61059028fd7000e49956e/issues/energinet-datahub/geh-charges/1836

PR #5050, opened 2025-12-10
Support for multiple processes in metering-point process-overview

frontend-specialisten:
## Description

- Add generic support for multiple processes in metering-point process-overview + docs

## References

[- #1170](https://github.com/Energinet-DataHub/team-mosaic/issues/1170)

### DIFF
--- a/libs/dh/charges/feature-charges/src/components/actions/create.ts
+++ b/libs/dh/charges/feature-charges/src/components/actions/create.ts
@@ -20,16 +20,15 @@ import { Component, computed, signal } from '@angular/core';
 import { FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
 import { TranslocoDirective } from '@jsverse/transloco';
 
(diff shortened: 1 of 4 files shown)


apps/dh/api-dh/source/DataHub.WebApi/Modules/MessageArchive/Models/ProcessStepType.cs
Sejruppen: So this file and "apps/dh/api-dh/source/DataHub.WebApi.Tests/Snapshots/SchemaTests.ChangeTest.verified.graphql" needs to be in sync always?

frontend-specialisten: This is done auto with the graphql snapshot testing, when new types are added the schema -> snapshot will change -> need verification otherwise the test will fail

PR #5052, opened 2025-12-10
Movein and Update Customer Data

Sejruppen:
This PR set out to update the modals of the "Start move in" and "Update customer data" modals, but ended up updating the design as well as moving the "Update customer data" into its own page instead, because of the size of the modal.

image
image

### DIFF
--- a/libs/dh/metering-point/feature-move-in/src/components/dh-business-customer-details-form.component.ts
+++ b/libs/dh/metering-point/feature-move-in/src/components/dh-business-customer-details-form.component.ts
@@ -24,19 +24,18 @@ import { WattTextFieldComponent } from '@energinet/watt/text-field';
 
 @Component({
   selector: 'dh-business-customer-details',
-  imports: [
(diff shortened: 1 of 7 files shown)


libs/dh/metering-point/feature-move-in/src/components/dh-address-details.component.ts
```diff
@@ -0,0 +1,205 @@
...
+
+          <watt-text-field
+            [formControl]="groupControls.darReference"
+            [label]="t('darReference')"
+            data-testid="dar-reference"
+          />
+
+          <watt-checkbox
+            [formControl]="formGroup.controls.nameAddressProtection"
+            class="checkbox-margin-bottom"
```
dzhavat: suggestion: let's use the `watt-` util classes to add bottom margin. Remember to remove the custom class from `styles`.


```suggestion
            class="watt-space-stack-l"
```

PR #5066, opened 2025-12-15
Feature/use edi endpoint for charge link actions

mimse:


apps/dh/api-dh/source/DataHub.WebApi/Modules/Charges/Client/ChargesClient.cs
```diff
@@ -89,22 +87,15 @@ public async Task<IEnumerable<Charge>> GetChargesByTypeAsync(
...
+        var user = httpContext?.HttpContext?.User;
 
-        if (currentUser.ActorRole == ActorRole.SystemOperator || currentUser.ActorRole == ActorRole.EnergySupplier)
-        {
-            ownerGln = (await marketParticipantClient_V1.ActorGetAsync(ct))
-            .Where(x => x.MarketRole.EicFunction == DataHub.WebApi.Clients.MarketParticipant.v1.EicFunction.SystemOperator).SingleOrDefault()?.ActorNumber.Value;
-        }
-
-        if (string.IsNullOrWhiteSpace(ownerGln))
+        if (user == null)
```
ManBearTM: question: What is the difference between these two approaches? I'm using `CreateUserIdentity` in my PR, so just wondering if I should use `user.GetMarketParticipantNumber` instead

apps/dh/api-dh/source/DataHub.WebApi/Modules/Charges/Extensions/ChargeExtensions.cs
```diff
@@ -19,31 +19,30 @@ namespace Energinet.DataHub.WebApi.Modules.Charges.Extensions;
 
 public static class ChargeExtensions
 {
-    public static ChargeInformationPeriodDto? GetCurrentPeriod(this ChargeInformationDto charge)
-    {
-        return charge.Periods
+    public static ChargeInformationPeriodDto? GetCurrentPeriod(this ChargeInformationDto charge) => charge.Periods
```
ManBearTM: question: should we still use `currentPeriod` in the frontend (for things like description, name etc). Or should that use `latestPeriod` instead? How do we know when to use what? 😄

ManBearTM (approved):
Just a few questions, code looks good!

apps/dh/api-dh/source/DataHub.WebApi/Modules/Charges/Client/ChargesClient.cs
mimse: CreateUserIdentify uses something from ProcessManager, GetMarketParticipantNumber extracts it from the token. But I think it boils down to the same

apps/dh/api-dh/source/DataHub.WebApi/Modules/Charges/Extensions/ChargeExtensions.cs
mimse: Yes, we should change that. I will make a PR for it :-)

mimse (commented):
Nice work !

PR #5082, opened 2025-12-18
fix: update state accessor to use translation for process states

frontend-specialisten:
## Description

- Update `state` accessor to use translation for process states in `process-overview`

## References

- #0000


Sejruppen (approved):
praise: Nice :)

PR #5098, opened 2025-12-22
Adds BFF for Change Customer Characteristics

Sejruppen:


apps/dh/api-dh/source/DataHub.WebApi/Options/SubSystemBaseUrls.cs
```diff
@@ -43,4 +43,7 @@ public sealed record SubSystemBaseUrls
 
     [Required]
     public string ElectricityMarketBaseUrl { get; set; } = string.Empty;
+
+    [Required]
+    public string ChangeOfSupplierBaseUrl { get; set; } = string.Empty;
```
dzhavat: question: is this url needed?

ManBearTM: This again? 😄 If it is needed, make sure that it is configured in CD as well, otherwise deployment will fail due to the `Required` attribute

PR #5119, opened 2026-01-07
feat(dh): initial work for searching DH3-created metering points

dzhavat:
### DataHub
- add a new checkbox that can switch between searching DH2-migrated or DH3-created metering points. 
- put checkbox behind release toggle.

The checkbox DOES NOT have an effect on production since the param is not yet used in the BFF.

image

## References

- #0000


libs/dh/globalization/assets-localization/src/assets/i18n/da.json
```diff
@@ -599,7 +599,8 @@
         "containsLetters": "{{ meteringPoint.shared.error.containsLetters }}",
         "meteringPointIdLength": "{{ meteringPoint.shared.error.meteringPointIdLength }}"
       },
-      "noResultFound": "Intet resultat fundet"
+      "noResultFound": "Intet resultat fundet",
+      "searchDh2MeteringPoints": "Søg DataHub 2 målepunkter"
```
mimse: Suggestion: Could "Søg i migrerede målepunkter" Be better alternativ?

mimse (approved):
Looks good, just one comment

PR #5145, opened 2026-01-12
fix(dh): search dh3-created metering points

dzhavat:
### DataHub
- search dh3-created metering points by supporting navigation to `externalMeteringPointId`
- feature is NOT available in production
- only look DH3 metering points up in preprod regardless of checkbox value
- use checkbox value on search page to switch between migrated and dh3-created metering points

## References

- #0000

### DIFF
--- a/apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPoint/Mappers/MeteringPointMetadataMapper.cs
+++ b/apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPoint/Mappers/MeteringPointMetadataMapper.cs
@@ -370,5 +370,4 @@ public static class MeteringPointMetadataMapper
 
         return (long)(ulongRand % unsignedRange) + min;
     }
-
 }


apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPoint/Mappers/MeteringPointMetadataMapper.cs
```diff
@@ -0,0 +1,346 @@
...
+            InstallationAddress = meteringPoint.InstallationAddress?.MapToDto(),
+            ManuallyHandled = false,
+        };
+    }
+
+    public static CommercialRelationDto MapToDto(this DataHub.ElectricityMarket.Abstractions.Features.MeteringPoint.GetMeteringPoint.V1.MeteringPointDtoV1.CommercialRelationDto commercialRelation)
+    {
+        return new CommercialRelationDto
+        {
+            Id = 0,
```
mimse: Issue: If all ids are 0, I could get an issue with graphql cache. It uses the id to consolidate.
Two solutions I see
Disable consolidation of the main entity in the graphql provider file
Or
Make a custom id, as it has been done on charge, and charge links

mimse (changes requested):

apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPoint/Mappers/MeteringPointMetadataMapper.cs
dzhavat: Oh, good point. I think it's best to go with the custom id for now. Especially because the data here is not available in prod, so not fully utilizing the cache is ok.

apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPoint/MeteringPointNode.cs
```diff
@@ -202,17 +228,44 @@ public static async Task<bool> RequestConnectionStateChangeAsync(
...
+        var result = await electricityMarketClient
+        .SendAsync(new GetMeteringPointQueryV1(meteringPointId), ct)
+        .ConfigureAwait(false);
+
+        var meteringPoint = result.Data?.MeteringPoint ?? throw new InvalidOperationException("No MeteringPoint was returned");
+
+        // Map to MeteringPointDto
+        var meteringPointResult = new MeteringPointDto
+        {
+            Id = 0,
```
dzhavat: suggestion: generate a random id here

apps/dh/api-dh/source/DataHub.WebApi/Modules/ElectricityMarket/MeteringPoint/MeteringPointNode.cs
```diff
@@ -154,16 +162,34 @@ public static async Task<RelatedMeteringPointsDto> GetRelatedMeteringPointsAsync
     [Authorize(Roles = new[] { "metering-point:search" })]
     public static async Task<MeteringPointDto> GetMeteringPointAsync(
         string meteringPointId,
+        string? environment,
+        bool? searchMigratedMeteringPoints,
         CancellationToken ct,
         [Service] IHttpContextAccessor httpContextAccessor,
         [Service] IRequestAuthorization requestAuthorization,
-        [Service] AuthorizedHttpClientFactory authorizedHttpClientFactory)
+        [Service] AuthorizedHttpClientFactory authorizedHttpClientFactory,
```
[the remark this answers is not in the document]
mimse: as a general thing :-)

PR #5221, opened 2026-01-29
Use signals

mimse:


libs/watt/package/picker/__shared/watt-picker-base.ts
```diff
@@ -89,81 +79,56 @@ export abstract class WattPickerBase
     }
 
     return null;
-  }
+  });
```
dzhavat: issue: doesn't look like the `computed` is going to be rerun since it doesn't contain a Signal. Could that be a problem?

PR #5232, opened 2026-02-03
Actor Conversation: Cases list

Sejruppen:
Sets up the list of cases (UI only) with mocked data for now - API will be implemented later

### DIFF
--- a/libs/dh/metering-point/feature-actor-conversation/src/components/actor-conversation-list.ts
+++ b/libs/dh/metering-point/feature-actor-conversation/src/components/actor-conversation-list.ts
@@ -54,6 +54,12 @@ import { Conversation } from '../types';
     .no-margin {
       margin: 0;
     }
+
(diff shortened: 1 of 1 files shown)


libs/dh/globalization/assets-localization/src/assets/i18n/en.json
```diff
@@ -786,7 +786,8 @@
         "customerMasterData": "{{ shared.terminology.customerMasterData }}",
         "measurements": "{{ shared.terminology.measurements }}",
         "meteringPointMasterData": "{{ shared.terminology.meteringPointMasterData }}",
-        "priceAssociation": "{{ shared.terminology.chargeLink }}"
+        "priceAssociation": "{{ shared.terminology.chargeLink }}",
+        "newCase": "New case"
```
sahma19: Are these the official translations? :P

PR #5240, opened 2026-02-03
fix: remove show monthly sum checkbox from settlement reports

dzhavat:
## References

- #0000


libs/dh/reports/feature-settlement-reports/src/components/request-report/request-report-modal.component.ts
```diff
@@ -307,7 +292,7 @@ export class DhRequestReportModal extends WattTypedModal<SettlementReportRequest
               start: period.start,
               end: period.end ? period.end : null,
             },
-            includeMonthlySums: includeMonthlySum,
+            includeMonthlySums: false,
```
mimse: Suggestion: Should it be removed from the mutation as well?

mimse (approved):

PR #5242, opened 2026-02-04
Feature/1257 add date validation end of supply

frontend-specialisten:
## Description

- Add date validation end of supply

## References

- [#1257](https://github.com/Energinet-DataHub/team-mosaic/issues/1257)


ManBearTM (dismissed):
Looks good, perhaps the tests goes a little too deep and duplicates the implementation of DataHubCalendar, but let's just keep it - better than not having tests 🙂

PR #5291, opened 2026-02-11
feat: notifications - clear all (#2066)

michaelmajgaard:


libs/dh/core/feature-notifications/src/lib/dh-notifications-center.component.ts
```diff
@@ -125,7 +142,18 @@ import { DhSettlementReportNotificationComponent } from './dh-settlement-report-
         *transloco="let t; prefix: 'notificationsCenter'"
         class="notifications-panel watt-elevation"
       >
-        <h3>{{ t('headline') }}</h3>
+        <div class="notifications-panel__header">
+          <h3>{{ t('headline') }}</h3>
+
+          @if (notifications().length > 0) {
+            <watt-icon
```
mimse: Suggestion: Use button with an icon

PR #5314, opened 2026-02-16
Actor conversation: Message

Sejruppen:
Adds styling of the messages in the conversation
image

Video of the messages scrolling:
https://github.com/user-attachments/assets/a370452c-45af-470a-b21d-52c0690e178d


libs/dh/metering-point/feature-actor-conversation/src/components/actor-conversation-message.ts
```diff
@@ -0,0 +1,78 @@
...
+  styles: `
+    .message-container {
+      border-radius: var(--watt-radius-m);
+      border: 1px solid var(--watt-color-neutral-grey-300);
+    }
+
+    .no-margin {
+      margin: 0;
+    }
+  `,
```
dzhavat: suggestion: is there a reason for turning encapsulation off here? This will make the styles global i.e. `.message-container` will also apply to other elements with the same class. Consider removing adding the encapsulation back.

PR #5320, opened 2026-02-17
fix(dh): change checkbox text in pre-prod

dzhavat:
## References

- #0000


libs/dh/metering-point/feature-search/src/components/dh-search.component.ts
```diff
@@ -147,6 +152,9 @@ export class DhSearchComponent {
   private readonly router = inject(Router);
   private readonly modalService = inject(WattModalService);
 
+  readonly environment = inject(dhAppEnvironmentToken);
+  DhAppEnvironment = DhAppEnvironment;
```
mimse: Nitpick: I think its is just as typesafe, to use a string

mimse (dismissed):

libs/dh/metering-point/feature-search/src/components/dh-search.component.ts
dzhavat: Yeah, you're right :) 
Fixed

PR #5325, opened 2026-02-18
Actor conversation/read unread

Sejruppen:
Adds functionality to the menu item and adds read functionality when clicking on a conversation


libs/dh/metering-point/feature-actor-conversation/src/components/actor-conversation-shell.component.ts
```diff
@@ -150,8 +154,14 @@ export class DhActorConversationShellComponent {
     this.selectedConversationId.set(undefined);
   }
 
-  selectConversation(conversationId?: string): void {
+  async selectConversation(conversationId: string) {
     this.newConversationVisible.set(false);
     this.selectedConversationId.set(conversationId);
+    await this.readConversationMutation.mutate({
```
mimse: Suggestion: Why is this not done in the backend. When a conversation is read for the first time, the boolean is set? Then we dont need this logic :-)

mimse (changes requested):

mimse (approved):

PR #5329, opened 2026-02-18
Actor Conversation: Search (and other general updates)

agf2413:
Added search for actor conversations 
(edited by Sejruppen)

image

Also adds a new watt-component - "Separator"

image


libs/dh/metering-point/feature-actor-conversation/src/components/actor-conversation-list.ts
```diff
@@ -77,6 +86,26 @@ import { dayjs } from '@energinet/watt/core/date';
...
+        />
+        <watt-dropdown
+          [formControl]="formControl"
+          [options]="options()"
+          [multiple]="true"
+          [chipMode]="true"
+          [placeholder]="t('filters.placeholder')"
+        />
+      </vater-stack>
+      <hr class="watt-divider thick-divider" />
```
mimse: Suggestion: Maybe this should be watt component with variants. Or maybe add the thick to watt instead of a custom style here.

mimse (changes requested):
Looks good, have a few comments

PR #5360, opened 2026-02-24
feat(watt): watt-json-viewer

ManBearTM:
watt-json-viewer

Very simple JSON viewer, implemented using a recursive component strategy.

- Supports JSON as well as JS objects (including circular references)
- Shows "inline" values for primitives or when collapsed
  - KNOWN LIMITATION: Currently the entire value is inlined. If it is a large object/array, this will look strange.
- Objects and arrays can be expanded to show values on individual lines
- Includes "toggle all" support
- Has "good enough" syntax highlighting using regex


ManBearTM:
> Do you know how big the json files can be? I build logic in to watt code to handle very large json and xml files. Is that needed here as well?

AFAIK, this will be used for debugging a single metering point, so the JSON should be a reasonable size. It will not perform very well on large arrays or objects, so if we are to use it for messages in JSON format, then we need to handle this better. Initially though, it is just for metering point debugging (but I would love to extend this to other use cases as well).

> Can we somehow combine the two?

I did try to implement this in `watt-code`, but the problem was that `highlightjs` completely takes over the DOM, so I could not put any custom elements into the `<code>` tag, making expand/collapse impossible. Their suggested solution was a custom plugin to `highlightjs`, but that seemed very complicated and hard to maintain, so I opted for the "home brewed" version.

libs/watt/package/json-viewer/watt-json-viewer.component.ts
```diff
@@ -0,0 +1,63 @@
...
+//#endregion
+import { Component, computed, input, linkedSignal } from '@angular/core';
+import { TreeState, WattJson } from './watt-json.component';
+
+@Component({
+  // eslint-disable-next-line @angular-eslint/component-selector
+  selector: 'watt-json-viewer',
+  imports: [WattJson],
+  template: `<watt-json [json]="json()" [tree]="tree()" />`,
+})
```
dzhavat: suggestion: let's make this OnPush as well since that's the main component, and it only relies on Signals

PR #5473, opened 2026-03-11
fix(watt-button): remove padding when used with icon only and no text

Supelana:


mimse:
We have some side effects
before
image
after
image

mimse:
before:
image


after:
image

PR #5636, opened 2026-03-27
Make test faster

mimse:
Getting back to test speeds before implict libaries

Changes:

- Moved to happy dom in watt
- Suppressing warnings and errors related internals that does not affect tests
- Switched to waitForAsync to avoid observable, that does not work with happy-dom
- Scope tsconfig to pr lib, to avoid build to much of the app for every test
- Dont try to run tests in a feature that does not have any spec files
- Resolve vitest types globally
- Fixed baseUrl deprecation warning


dzhavat (approved):
praise: nice work! Anything that makes the CI faster is worth it 🏆

PR #5695, opened 2026-04-13
Move in: Add cutoff date to update customer data

Sejruppen:
We need to provide the backend with the correct ExpectedValidityDate from the BFF. This PR achieves that and also updates some packages due to a security risk


mimse (dismissed):
Not a fan, that this needs to be handled in the BFF. But at least we dont have the cutoff date as query param :-)

Other than that, just one comment

mimse (approved):

PR #5741, opened 2026-04-20
fix(dh): charges filter prevent illegal intervals

ManBearTM:


Sejruppen (approved):
praise: Nice addition! 🙂

PR #5753, opened 2026-04-22
Update customer data: CPR lookup idea

Sejruppen:


mimse (changes requested):
A few changes can be made :-)

PR #5757, opened 2026-04-22
perf(dh): reduce initial bundle by ~520 KB (22%)

frontend-specialisten:
## Summary

Reduces the `app-dh` initial (eager) bundle from **2.39 MB to ~1.85 MB** (-540 KB, -22.6%) by fixing four classes of leakage where heavy code was getting pulled into the first-paint graph despite existing lazy-loading infrastructure.

- Runtime behavior of each `Intl` service, `ErrorHandler`, phone validator, and modal is preserved. The only thing that changes is when/how the JS arrives on the client.
- Errors raised before App Insights finishes loading are buffered (capped at 50) and replayed through the real handler on `adopt()`. If the SDK import itself fails, the wrapper adopts a plain `ErrorHandler` so bootstrap errors still reach the console.
- `isValidPhoneNumber` behavior is identical for the 9 supported countries; dropping metadata for other countries only affects codes the app never accepts anyway.
- `WattNavList` keeps the same auto-expand-on-active-route behavior and ARIA relationship between header and body.

### What moved out of eager

| Package / module | Before | After |
|---|---|---|
| Profile modal (`DhProfileModalComponent` + transitive deps) | eager | lazy (dynamic `import()` on avatar click) |
| `libphonenumber-js` (via profile modal chain) | eager | lazy |
| `@angular/material/datepicker` (via watt intl providers) | eager | lazy |
| `@energinet/watt/datepicker`, `dropzone`, `code`, `phone-field` components | eager | lazy |
| `@microsoft/applicationinsights-web` + `angularplugin-js` | eager | lazy (fetched during `APP_INITIALIZER`) |
| `highlight.js` + `WattCodeComponent` | eager | lazy |
| `@angular/material/expansion` + `MatAccordion` | eager | gone (nav list rewritten) |
| `libphonenumber-js` metadata (in lazy chunk) | 83 KB (all countries, min) | 6 KB (DK, SE, NO, GB, DE, FI, PL, NL, CH) |


frontend-specialisten:
suggestion: This bundled PR is being closed in favor of the same work split into one PR per optimization, so each can be reviewed and measured on its own:

- #5800 perf(dh): narrow watt intl imports to bypass component barrels
- #5801 perf(dh): lazy-load profile modal from avatar menu
- #5802 perf(dh): defer Application Insights SDK out of eager bundle (~91 KB)
- #5803 perf(watt): ship libphonenumber metadata for 9 supported countries only (lazy chunk only, ~75 KB)
- #5804 perf(dh): import WATT_CODE_HIGHLIGHT_WORKER_FACTORY from narrow subpath
- #5805 perf(watt): rewrite WattNavList expand/collapse without MatExpansion (~20 KB)
- #5806 chore(dh): tighten app-dh initial bundle budget to 1.9 MB / 2.1 MB (merge LAST)

Combined effect on the initial bundle is the same as measured here: ~2.39 MB to ~1.85 MB (-540 KB / -22.6%). Each individual PR description outlines the eager-bundle items it removes and how to A/B verify the contribution.

PR #5766, opened 2026-04-27
fix(dh): adjustments to fees under charge links

dzhavat:
## References

- #0000


frontend-specialisten:
Consideration: this change introduces non-trivial conditional behaviour (action visibility branched on charge type, datepicker hidden for FEE, form value derived from query data) but there's no `.spec.ts` in `libs/dh/metering-point/feature-chargelink/` covering it. A test that mounts the edit modal with a FEE charge and asserts `startDate` is pre-populated from `periods[0].period.start` would have caught the chargeId/chargeLinkId mix-up Copilot flagged.

PR #5782, opened 2026-04-29
fix(dh): fix method call in pipe

dzhavat:
Co-authored-by: Copilot <copilot@github.com>

## References

- #0000


ManBearTM (approved):

ManBearTM:
Thank you! I'll look into fixing whatever is causing the DateRange scalar to not be converted, they really should not be strings (no one likes lying types)

